PR #3, opened 2018-10-01
fix(themes): Add 'seekAsia' theme

ashin:
Initial pass of base SeekAsia theme.

Added a version of the disabled checkbox to the playground for a better understanding of the design.


lib/themes/seekAsia/index.js
```diff
@@ -0,0 +1,8 @@
+import tokens from './tokens/tokens';
+import * as atoms from './atoms/atoms';
+
+export default {
+  name: 'jobStreet',
```
markdalgleish: Should be `seekAsia`.

lib/themes/seekAsia/atoms/private/palette.js
```diff
@@ -0,0 +1,11 @@
...
+export const black = '#000';
+export const blue2 = '#298EB9';
+export const blue3 = '#94C9E0';
+export const blue5 = '#EEF8FC';
+export const alert = '#ff4b4b';
+export const alertBackground = 'rgba(255,75,75, 0.15)'; // transparent seekAsia alert color
+export const completion = '#63b209';
+export const grey1 = '#333';
+export const grey4 = '#ccc';
+export const candidate = '#0c4b85'; // SEEK Asia name for this color
```
markdalgleish: Nice one 👏

I've been meaning to do this for the other themes too. Great to see the pattern introduced.

PR #9, opened 2018-10-02
build: Seperate out and document release configuration

jahredhope:
semantic-release [allows the configuration](https://github.com/semantic-release/semantic-release/blob/caribou/docs/usage/configuration.md#configuration) either via `package.json` values or `release.config.js` values.
`package.json` values can not be commented, and without context it can be unintuitive why certain configurations have been chosen.

This change helps document the release process, helping with any further changes that may be made.

Confirmed after this change no success plugins are loaded still:
screen shot 2018-10-03 at 8 25 32 am

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,9 +21,6 @@
       "path": "./node_modules/cz-conventional-changelog"
     }
   },
-  "release": {
-    "success": false
-  },
   "repository": {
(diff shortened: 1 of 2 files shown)


markdalgleish:
Nice one, much clearer now.

PR #18, opened 2018-10-17
ci: Deploy pull request preview site to Surge

markdalgleish:


markdalgleish:
🎉  First preview deployment: http://braid-design-system--501806cde209b0e28c229ba358cc59285e8a81e3.surge.sh/

sku.config.js
```diff
@@ -1,3 +1,5 @@
+const isPullRequest = !!process.env.TRAVIS_PULL_REQUEST_SHA;
```
michaeltaranto: This is true for dev as well though right? I would say its more doing something custom for gh-pages and the same for surge and dev.

markdalgleish: Note the double negative. Would this be clearer as `Boolean(process.env.TRAVIS_PULL_REQUEST_SHA)`? Is there a better way to write this?

markdalgleish: In dev, it's `undefined`, but in Travis it's `""`.

markdalgleish: Updated, let me know what you think 👍🏻

scripts/postCommitStatus.js
```diff
@@ -0,0 +1,36 @@
+/* eslint-disable no-console */
+(async () => {
+  try {
+    console.log('Posting commit status to GitHub...');
+
+    const { GH_TOKEN, TRAVIS_PULL_REQUEST_SHA } = process.env;
+
+    if (!GH_TOKEN && !TRAVIS_PULL_REQUEST_SHA) {
```
michaeltaranto: This should be a `||` right?

markdalgleish: Correct. You passed the test.

.travis.yml
```diff
@@ -6,7 +6,10 @@ before_script:
   - npm prune
 after_success:
   - if [ ${TRAVIS_SECURE_ENV_VARS} = "true" ]; then
-      if [ "$TRAVIS_BRANCH" == "master" ]; then
+      if [ -n "$TRAVIS_PULL_REQUEST_SHA" ]; then
+        npm run travis-deploy-once "npm run deploy-preview -- -d braid-design-system--${TRAVIS_PULL_REQUEST_SHA}.surge.sh";
+        npm run travis-deploy-once "npm run post-commit-status";
+      elif [ "$TRAVIS_BRANCH" == "master" ]; then
```
michaeltaranto: This reads well now 👏

PR #28, opened 2018-11-12
feat(webpack): Add initial webpack decorator, add Playroom to docs

markdalgleish:


markdalgleish:
Changing approach, which means we don't need the webpack decorator just yet. Closing for now.

PR #36, opened 2018-11-15
fix: Align white space token values, rename to T-shirt sizes

markdalgleish:
In order to ensure consistent scaling of white space across themes, we've adjusted the token values to match for every brand.

White space still scales relative to the underlying grid tokens, so the resulting white space values still differ for each theme, but they now scale more predictably when toggling themes.

## Migration Guide

The following spacing values need to be renamed:

```diff
-smallest
+xxsmall

-smaller
+xsmall

-larger
+xlarge

-largest
+xxlarge
```

For example:

```diff
-<Box paddingBottom="smaller">
+<Box paddingBottom="xsmall">

-theme.tokens.rowSpacing.largest
+theme.tokens.rowSpacing.xxlarge
```


markdalgleish:
Updated the PR to exclude `Box` from our Storybook test suite.

PR #41, opened 2018-11-18
docs(ComponentProps): Drive with react-docgen contract

michaeltaranto:
In preparation for migrating the component library to TypeScript, this PR refactors `ComponentProps` to build itself off a pre-constructed dataset. This dataset is automatically generated from the component tree on every `build` or `start`, and is built as a set of structures that follow the [react-docgen format](https://github.com/reactjs/react-docgen#result-data-structure), keyed by the component `displayName`.

As a result of not using runtime analysis, a component's `options` (eg. prop-types) will now also be visible on the docs site in production.


lib/componentDocs/build.js
```diff
@@ -0,0 +1,49 @@
...
+
+const normaliseData = ({ props = {}, ...rest }) =>
+  Object.keys(props).reduce(
+    (acc, prop) => ({
+      ...acc,
+      props: {
+        ...acc.props,
+        [prop]: {
+          ...props[prop],
+          defaultValue: normaliseDefaultValue(props[prop]),
```
mattsjones: Any chance of using `lodash.mapValues` here? Feel like this kinda stuff can be intimidating to some people and it would only need to be a `devDependency`.

lib/componentDocs/normaliseProp.js
```diff
@@ -0,0 +1,37 @@
+const normaliseType = ({ type }) => {
```
mattsjones: 👏

lib/componentDocs/build.js
michaeltaranto: It could be that my `lodash` skills aren't up to scratch, but looking at the docs I can't see how the code would clean up this code. Due to the depth of the update taking place there would still need to be a bunch of spreading no?

PR #52, opened 2018-11-21
fix: Add `toUnit` util

michaeltaranto:
Adding a `toUnit` until and implementing `px`. Consolidates all the unit concatenation template strings, allowing the code to focus on the calculation.


markdalgleish:
Side note, it's kinda cool to know that the screenshots didn't change when refactoring the CSS.

PR #53, opened 2018-11-23
fix(Text): Migrate to TypeScript

michaeltaranto:
Migrate `Text` to TypeScript and fix a couple of bugs:
- Missing `Roboto` as the primary font family for seekAnz theme
- Apply desktop font size rules to `interaction` classes

Also:
- Centralises knowledge of `basekick` and aligning the text to the grid, removing the need for `Text` to apply a `transform: none` override to support `interaction`.
- Removes redundant child `span` inside every `Text` instance.


lib/themes/utils/transformForTokens.js
```diff
@@ -0,0 +1,31 @@
+import merge from 'lodash/merge';
+import alignTextToGrid from './alignTextToGrid';
+import createResponsiveRules from './createResponsiveRules';
+
+export default ({ tokens }) => {
+  const rules = [];
+  Object.keys(tokens.text).forEach(typeName => {
```
mattsjones: Just a personal preference here but could we maybe reduce this into the object? That way we don't need merge.

michaeltaranto: I prefer reducing, but to clean it up would mean following this down into `createResponsiveRules` and everywhere that is used. I would like to come back and convert all the `utils` to TypeScript once the components are done.

PR #58, opened 2018-11-27
fix(ThemeConsumer): Deprecate in favour of more specific consumers

michaeltaranto:
This PR deprecates the public `ThemeConsumer` component and instead exposes two more explicit context consumers — `ThemeNameConsumer` and `ThemeTokensConsumer`. The fundamental change here is that `atoms` are now no longer part of the public API.

BREAKING CHANGE: `ThemeConsumer` is deprecated and the `atoms` of a theme are no longer exposed publicly.

### Migration Guide
If you were using `ThemeConsumer` to get access to the `Theme`, here are the migration options:

#### Theme Name
```diff
-<ThemeConsumer>
+<ThemeNameConsumer>
-  {theme => (
-    <span>The active theme is {theme.name}.</span>
+  {themeName => (
+    <span>The active theme is {themeName}.</span>
  )}
-</ThemeConsumer>
+</ThemeNameConsumer>
```

#### Theme Tokens
```diff
-<ThemeConsumer>
+<ThemeTokensConsumer>
-  {theme => (
-     <div style={{ minHeight: theme.tokens.rowHeight * 4}}>
+  {tokens => (
+     <div style={{ minHeight: tokens.rowHeight * 4}}>
         My special component
      </div>
  )}
-</ThemeConsumer>
+</ThemeTokensConsumer>
```

#### Theme Atoms
No longer exposed publicly, we believe exposing low level components and theme tokens is enough to build a component externally with the same look and feel as an official component. An example migration:
```diff
-<ThemeConsumer>
-  {theme => (
-    <span className={theme.atoms.backgroundColor.selection}>
+    <Box component="span" backgroundColor="selection">
        This element has been styled using the themed 'backgroundColor'.
-    </span>
+    </Box>
-  )}
-</ThemeConsumer>
```


lib/components/ThemeNameConsumer/ThemeNameConsumer.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { Component, ReactNode } from 'react';
+import ThemeContext from '../private/ThemeContext';
+
+interface Props {
+  children(name: string): ReactNode;
+}
+
+export default class ThemeNameConsumer extends Component<Props> {
```
markdalgleish: Needs a `displayName`?

lib/components/ThemeTokensConsumer/ThemeTokensConsumer.tsx
```diff
@@ -0,0 +1,23 @@
+import React, { Component, ReactChild } from 'react';
+import ThemeContext from '../private/ThemeContext';
+import { Tokens } from '../../themes/theme';
+
+interface Props {
+  children(tokens: Tokens): ReactChild;
+}
+
+export default class ThemeTokensConsumer extends Component<Props> {
```
markdalgleish: Also needs a `displayName`?

PR #80, opened 2019-01-14
fix: Migrate remaining components to TypeScript

markdalgleish:
This PR migrates the remaining JavaScript components to TypeScript:
- Checkbox
- ChecklistCard
- FieldMessage
- Radio

This also performs a number of refactors and clean-ups:
- Deleted the `withTheme` HOC in favour of `ThemeConsumer`, since render props are _dramatically_ easier to type.
- Exported all component prop interfaces to allow other components to easily extend them.
- Consistently named prop interfaces, e.g. a `Foo` component should have a `FooProps` interface, rather than just `Props`. This is partly to ensure that error messages from TypeScript are more readable when interfaces are not satisfied.
- Ensured all components extend prop interfaces correctly when spreading props, e.g. `export interface TextProps extends BoxProps { ... }`
- Removed props from interfaces that are duplicates of inherited props, e.g. `className`.
- Generally expanded the type annotations in existing TypeScript files (not merely to satisfy the compiler, but to better document the code).
- Removed `react-docgen` since we no longer use `prop-types` in Braid components.
- Moved `prop-types` to `devDependencies` (until the documentation site is also migrated to TypeScript).


lib/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,268 @@
...
+import TickIcon from '../icons/TickIcon/TickIcon';
+import styles from './Checkbox.css.js';
+import { px } from '../../themes/utils/toUnit';
+
+const textColorForState = ({
+  disabled,
+  hovered
+}: {
+  disabled: boolean | undefined;
+  hovered: boolean | undefined;
```
mattsjones: Think this should just be `boolean` as the state inits with `false`.

lib/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,268 @@
...
+  render() {
+    return (
+      <ThemeConsumer>
+        {theme => {
+          const {
+            variant,
+            id,
+            checked,
+            label,
+            disabled,
```
mattsjones: Thoughts on making `disabled` have a default value of `false`? I think that'd be cleaner.

lib/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,268 @@
...
+                      paddingBottom="small"
+                      className={styles.children}
+                      style={{ marginLeft: px(checkboxSize) }}
+                    >
+                      {children}
+                    </Box>
+                  </Box>
+                ) : null}
+              </div>
+              {message !== false ? (
```
mattsjones: Is this a tri-state? e.g. `true`, `false`, `undefined`?
If so can we represent that as an enum instead?

lib/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,246 @@
...
+import FieldMessage from '../FieldMessage/FieldMessage';
+import styles from './Radio.css.js';
+import { px } from '../../themes/utils/toUnit';
+
+const textColorForState = ({
+  disabled,
+  hovered
+}: {
+  disabled: boolean | undefined;
+  hovered: boolean | undefined;
```
mattsjones: Same as above

lib/components/Text/Text.tsx
```diff
@@ -26,48 +24,51 @@ const isTransformVariant = (
...
+          const {
+            component,
+            size,
+            color,
+            weight,
+            baseline = true,
+            className,
+            ...restProps
+          } = this.props;
+          const fontSize = size || 'standard';
```
mattsjones: Can we just make `size` has a default value of `standard`?

lib/components/icons/Icon/Icon.tsx
```diff
@@ -21,44 +18,47 @@ const isSizeVariant = (
...
-        />
-      );
-    }
+            size,
+            svgComponent,
+            inline = false,
+            fill,
+            ...restProps
+          } = this.props;
+          const sizeAtom = `${size || 'standard'}Text${inline ? 'Inline' : ''}`;
```
mattsjones: Can size have a default value of `standard`?

lib/components/Checkbox/Checkbox.tsx
markdalgleish: Nope, it's either a string, or `false` to disable the message entirely.

markdalgleish:
@mattsjones Responded to all feedback, just pushed an update.

PR #82, opened 2019-01-22
test: Migrate tests to TypeScript

markdalgleish:
Because our tests deliberately import components via the public API, I was forced to rename the main entry point from `components/index.js` to `components/index.ts`, but decided not to trigger a release yet since it doesn't impact consumers.

Since we're now using TypeScript, I've deleted the redundant test that ensured all theme objects have the same shape, especially because it proved to be quite difficult to type.


lib/components/Alert/Alert.test.tsx
```diff
@@ -1,15 +1,16 @@
...
 
 describe('Alert', () => {
   forEachTheme(({ ThemeProvider }) => {
     describe('Tones', () => {
-      const tones = ['critical', 'info'];
+      const tones: Array<AlertProps['tone']> = ['critical', 'info', undefined];
 
       tones.forEach(tone => {
-        test(tone, () => {
+        test(tone || 'default', () => {
```
michaeltaranto: Is it equivalent to push 'default' into the `tones` array above in place of `undefined`?

markdalgleish: `undefined` is one of the prop values, since it's optional. The `default` string here is just so that the test has a name.

TypeScript actually caught this one for me, telling me that `test` accepts a string as the first arg and I was unwittingly passing a value that could be undefined 😄

PR #84, opened 2019-01-24
chore: Migrate docs site, Storybook and Playroom to TypeScript

markdalgleish:


lib/components/Reset/Reset.docs.tsx
```diff
@@ -1,7 +1,6 @@
-import Reset from './Reset';
+import { ComponentDocs } from '../../../docs/src/types';
 
-export default {
-  component: Reset,
```
michaeltaranto: Was removing `Reset` deliberate here?

michaeltaranto: Ignore me

PR #90, opened 2019-02-08
fix(ChecklistCard): Show error border on entire row

markdalgleish:
This is to fix an issue where the error border wasn't very visible against the background colour when a checkbox is checked:

screen shot 2019-02-08 at 11 39 24 am

Instead, we now show the error border around the entire row:

screen shot 2019-02-08 at 11 40 25 am


stephaniemoss:
Agreed with Michael, there are a couple unideal visual things going on which should be reviewed in the near future, but happy for this to be implemented for now.

PR #92, opened 2019-02-13
fix(ChecklistCard): Ensure divider doesn’t cover error border

markdalgleish:
## Before:

screen shot 2019-02-13 at 11 50 16 am

### After:

screen shot 2019-02-13 at 11 50 30 am

Note: This introduces a double border between checkboxes when multiple error states are stacked, but this is being treated as the lesser of two evils for now.

### DIFF
--- a/lib/components/ChecklistCard/ChecklistCard.tsx
+++ b/lib/components/ChecklistCard/ChecklistCard.tsx
@@ -19,14 +19,24 @@ export default class ChecklistCard extends Component<ChecklistCardProps> {
     const { children, ...restProps } = this.props;
     return (
       <Card {...restProps}>
-        {React.Children.map(children, (child, i) => (
-          <Fragment>
-            {i > 0 && <Divider />}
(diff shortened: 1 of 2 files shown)


lib/components/ChecklistCard/ChecklistCard.tsx
```diff
@@ -19,14 +19,24 @@ export default class ChecklistCard extends Component<ChecklistCardProps> {
...
-        {React.Children.map(children, (child, i) => (
-          <Fragment>
-            {i > 0 && <Divider />}
-            {React.cloneElement(child as ReactElement<CheckboxProps>, {
-              variant: 'inChecklistCard'
-            })}
-          </Fragment>
-        ))}
+        {React.Children.map(children, (child, i) => {
+          if (typeof child !== 'object') {
```
mengtzu: What's this guarding against specifically?

markdalgleish: Children can be primitive types, like string or number, which don't have a `props` property.

PR #96, opened 2019-02-18
test(storybook): Prevent margin collapse in stories

markdalgleish:
This ensures that Chromatic includes component margins in screenshots.


markdalgleish:
Closing in favour of a fix via sku.

PR #99, opened 2019-02-20
fix: Add Heading component

markdalgleish:
Example usage:

```jsx
<Heading level="1">Heading Level 1</Heading>
<Heading level="1" weight="weak">Heading Level 1 (Weak)</Heading>

<Heading level="2">Heading Level 2</Heading>
<Heading level="2" weight="weak">Heading Level 2 (Weak)</Heading>

<Heading level="3">Heading Level 3</Heading>
<Heading level="3" weight="weak">Heading Level 3 (Weak)</Heading>
```


lib/themes/utils/fontSizeRulesForTokens.js
```diff
@@ -2,66 +2,71 @@ import merge from 'lodash/merge';
...
+          selector: `.${typeName}`,
+          mobileRules,
+          desktopRules
         })
       );
+
+      if (typeName === 'standard') {
+        const interactionHeight = tokens.interactionRows * tokens.rowHeight;
+        const mobileInteractionPadding = px(
+          (interactionHeight - type.mobile.rows * tokens.rowHeight) / 2
```
mengtzu: Dividing always makes me 😰 - how does this behave if it's an odd value on the left hand side?

mengtzu:
Looks good but I'm a little surprised to see `weak` as a thing - I guess that's the agreed upon language?

lib/themes/utils/fontSizeRulesForTokens.js
michaeltaranto: I hear you, in this case partial pixel values should not matter as by applying the difference between `interactionHeight` and the computer type height equally above and below the text, it will have the desired outcome of vertically centering within the touch target.

PR #101, opened 2019-02-21
fix(Columns): Add Columns and Column components

cm-seekasia:
Example usage:

```jsx
<Columns>
  <Column>...</Column>
  <Column>...</Column>
  <Column>...</Column>
</Columns>
```


markdalgleish:
Just pushed a big change to the API, to future proof things a bit. We now have a matching `<Column>` component, rather than letting `<Columns>` accept arbitrary children.

For example:

```jsx
<Columns>
  <Column>...</Column>
  <Column>...</Column>
  <Column>...</Column>
</Columns>
```

This allows us to add props to each `Column` in the future, e.g. customising the size of the column.

lib/components/Column/Column.css.js
```diff
@@ -0,0 +1,7 @@
+export default {
+  '.column': {
+    flexBasis: 0,
+    flexGrow: 1,
+    minWidth: 0
```
patrickliu-sa: It seems `flexBasis` work as the arbitrary `minWidth` under flexbox 😅

michaeltaranto: Yeah this was a lesson we learnt from the [`seek-style-guide`](https://github.com/seek-oss/seek-style-guide/pull/350) about how `min-width` is handled by flex elements.

PR #102, opened 2019-02-22
fix(TextLink): Add TextLink and TextLinkRenderer components

markdalgleish:
`<TextLink>` renders a native `<a>` element.

If you need to render something else (e.g. a React Router `<Link>` component), we provide the lower-level `<TextLinkRenderer>` component.

For example:

```jsx
import React from 'react';
import { Link } from 'react-router-dom';
import { TextLinkRenderer } from 'braid-design-system';

export default () => (
  <Text>
    <TextLinkRenderer>
      {textLinkProps => (
        <Link to="/" {...textLinkProps}>
          Link
        </Link>
      )}
    </TextLinkRenderer>
  </Text>
);
```


lib/components/TextLinkRenderer/TextLinkRenderer.tsx
```diff
@@ -0,0 +1,35 @@
...
+  static displayName = 'TextLinkRenderer';
+
+  render() {
+    const { children } = this.props;
+
+    return (
+      <ThemeConsumer>
+        {theme =>
+          children({
+            style: {},
```
michaeltaranto: What was the thinking behind this? Is this solely to prevent consumers adding their own `style` overrides?

markdalgleish: Yep, exactly.

PR #103, opened 2019-02-22
fix(Hidden): Add Hidden component

hscgavin:
## Example usage

Hidden on mobile:
```jsx
<Hidden mobile>
  <Text>Hidden on mobile</Text>
</Hidden>
```

Hidden on desktop:
```jsx
<Hidden desktop>
  <Text>Hidden on desktop</Text>
</Hidden>
```

Hidden inline:
```jsx
<Text>The end of this sentence is <Hidden inline mobile>hidden on mobile.</Hidden></Text>
```

Hidden on print:
```jsx
<Hidden print>
  <Text>Hidden on print</Text>
</Hidden>
```

Hidden on screen:
```jsx
<Hidden screen>
  <Text>Hidden on screen</Text>
</Hidden>
```


lib/components/Hidden/Hidden.docs.tsx
```diff
@@ -0,0 +1,25 @@
...
+import Text from '../Text/Text';
+
+export default {
+  component: Hidden,
+  examples: [
+    {
+      label: 'Hidden on mobile',
+      render: () => (
+        <Hidden component="span" hideFor="mobile">
+          <Text baseline={false}>Hidden on mobile</Text>
```
michaeltaranto: The `baseline={false}` is not really required, it's best to try and keep these docs code blocks as lean as possible.

lib/components/Hidden/Hidden.tsx
```diff
@@ -0,0 +1,28 @@
...
+export default class Hidden extends Component<HiddenProps> {
+  static displayName = 'Hidden';
+
+  render() {
+    const { children, component, hideFor } = this.props;
+
+    return (
+      <Box
+        display={[
+          hideFor === 'mobile' ? 'none' : 'inline',
```
michaeltaranto: Whats the reason for falling back to `inline` on mobile?

hscgavin: so children is the one decide the display style `display: xxx`, if falling back to `block`, children couldn't override it? thoughts?

mengtzu: We had to pass something there and `inline` is something.  If the child is block it will still behave as expected.

Better ideas gratefully accepted 😅

markdalgleish:
Just pushed an update that adds support for print/screen hiding, which necessitated a change in the API towards boolean props, since you can mix and match.

For example:

```jsx
<Hidden desktop print>...</Hidden>
```

docs/src/App/App.tsx
```diff
@@ -59,122 +59,126 @@ export default withRouter(
...
-                    )}
-                  />
-                  <div
-                    className={classnames(
-                      styles.menuButton__bar,
-                      styles.menuButton__bar3
-                    )}
-                  />
-                </button>
+                <Hidden print>
```
mengtzu: Dogfooding straight away 👍 🐶 🥘

markdalgleish: Yeah, trying to make this more of a habit.

lib/components/Hidden/Hidden.tsx
```diff
@@ -0,0 +1,45 @@
...
+import styles from './Hidden.css.js';
+
+export interface HiddenProps {
+  children: ReactNode;
+  mobile?: boolean;
+  desktop?: boolean;
+  screen?: boolean;
+  print?: boolean;
+  inline?: boolean;
+  component?: ReactType;
```
mattsjones: As discussed maybe remove the component option for now.

markdalgleish: Done 👍

PR #114, opened 2019-02-27
chore: Setup notion toolchain

mattsjones:


mattsjones:
Notion doesn't work that way. You specify an exact version and it just ensures you use it. You don't need to run any commands like with `nvm` though so this doesn't add any time to set up or anything. Also, Notion is purely for dev experience, it's not related to the engines field.

mattsjones:
To be clear, I created this PR by running:
```
notion pin node 10
notion pin yarn latest
```

Notion just picked the latest version of 10 as the pinned one.

markdalgleish:
Is this now out of sync with .nvmrc?

mattsjones:
nvm is pointing to `lts/*` so they're actually aligned. At least for now. I really do think notion is the way to go though.

PR #115, opened 2019-02-27
refactor(atoms): Make fonts type safe

mattsjones:
2nd phase of making atoms type-safe. This PR refactors the font rules. Interestingly there is actually less code and I'd argue it's more readable. Also adding the type declaration for `basekick`.


lib/atoms/font/makeFontSizes.ts
```diff
@@ -1,74 +1,60 @@
...
 
-      if (typeName === 'standard') {
-        const interactionHeight = tokens.interactionRows * tokens.rowHeight;
-        const mobileInteractionPadding = px(
-          (interactionHeight - type.mobile.rows * tokens.rowHeight) / 2
-        );
-        const desktopInteractionPadding = px(
-          (interactionHeight - type.desktop.rows * tokens.rowHeight) / 2
-        );
+const makeRules = (responsiveType: 'mobile' | 'desktop', tokens: Tokens) => {
```
michaeltaranto: We have a poorly named type called `Breakpoint` that is the type for this key on the reponsive text part of the tokens

lib/atoms/font/makeFontSizes.ts
```diff
@@ -1,74 +1,60 @@
...
-    (type, typeName) => {
-      const {
-        fontSize: mobileFontSize,
-        lineHeight: mobileLineHeight
-      } = alignTextToGrid(type.mobile, tokens);
-      const {
-        fontSize: desktopFontSize,
-        lineHeight: desktopLineHeight
-      } = alignTextToGrid(type.desktop, tokens);
+const getFontSizeRules = (textDefinition: TextDefinition, tokens: Tokens) => {
```
michaeltaranto: Should we be explicit with the return types on these functions, eg `CSSProperties`, given what we saw yesterday with the baseline stuff

mattsjones: In this case, the type is inferred correctly. Would rather use inference where possible. Once we start using the CSSProperty types we may need to review a bunch of things. But for now, I think this is fine.

PR #122, opened 2019-03-04
fix(TextLink): Make touchable, add ‘inline’, remove ‘interaction’ text

markdalgleish:
`TextLink` is now touchable by default and renders a `Text` component internally on your behalf.

For example:

```js
<Card>
  <TextLink href="...">This is a touchable link.</TextLink>
</Card>
```

Note, the `interaction` text size has now been deprecated.

## Migration Guide

To maintain existing inline link behaviour, you will need to use the `inline` prop:

```diff
<Text>
-  The last word of this sentence is an <TextLink>inline link</TextLink>
+  The last word of this sentence is an <TextLink inline>inline link</TextLink>
</Text>
```


lib/components/private/getCheckboxRadioSize.ts
```diff
@@ -4,7 +4,7 @@ export default (theme: Theme) => {
   // We currently don't support responsive checkboxes and
   // radio buttons, but nobody actually needs it (so far)
   const scale = theme.tokens.text.standard.mobile.size / 28;
```
michaeltaranto: 🤔

lib/themes/theme.d.ts
```diff
@@ -171,6 +171,7 @@ type SpacingVariants =
   | 'xlarge'
   | 'xxlarge';
 type HorizontalSpacingVariants = SpacingVariants | 'gutter';
+type VerticalPaddingVariants = SpacingVariants | 'standardTouchableText';
```
michaeltaranto: Didn't we decide on these types being singular?

lib/components/private/getCheckboxRadioSize.ts
markdalgleish: 🤔

lib/themes/theme.d.ts
markdalgleish: Fixed. Also updated the other two spacing types to be consistent.

PR #130, opened 2019-03-08
fix(Box): Add boxShadow, transition, transform and width support

michaeltaranto:
Beefing up `Box` to add support for a few more atoms in support for `Button` and `TextField`.

Changes/fixes include:
- Typo in `BackgroundColorParms`
- Background color atoms now explicitly set only `backgroundColor` rather than `background`
- Add `width` to `Box`, only option for now is `full`
- Use wireframe atoms file to infer types for normalise mappings
- Refactor `Checkbox`, `Radio` and `Divider` to adopt `Box` props in favour of using atoms directly


lib/atoms/types.ts
```diff
@@ -1,208 +1,3 @@
...
-  transform_level1Heading: string;
-  transform_level2Heading: string;
-  transform_level3Heading: string;
-  transform_standardText: string;
-  transition_fast: string;
-  width_largeText: string;
-  width_largeTextInline: string;
-  width_standardText: string;
-  width_standardTextInline: string;
-}
```
markdalgleish: 👋 😄

lib/components/Checkbox/Checkbox.css.js
```diff
@@ -60,7 +60,6 @@ export default {
     position: 'relative',
     zIndex: 5,
     // IE11:
-    width: '100%',
```
markdalgleish: Love that you took the effort to clean this up (and others like it) 👏👏👏

lib/components/Checkbox/Checkbox.tsx
```diff
@@ -92,9 +92,7 @@ export default class Checkbox extends Component<CheckboxProps, State> {
               <div className={styles.content}>
                 <Box
                   component="label"
-                  className={classnames({
-                    [styles.label]: true,
-                  })}
```
markdalgleish: lol, what was I thinking...

PR #133, opened 2019-03-08
feat(TextField): Make initial component

michaeltaranto:
First pass at the `TextField`, drawing on both style guides again to find what features are required. Added support for the following:
- Hover, focus and error states
- Standard, positive & critical `FieldMessage`s below the field
- Improved `FieldLabel` to handle secondary and tertiary labels

See the docs site and playroom for more examples:
- [Playroom example](https://braid-design-system--fbe13fcdd487f70749ff26498253d9dc4996a87a.surge.sh/playroom/#?code=PGRpdiBzdHlsZT17eyBiYWNrZ3JvdW5kQ29sb3I6ICIjZjJmMmYyIiwgbWluSGVpZ2h0OiAiMTAwdmgiIH19PgogIDxkaXYgc3R5bGU9e3sgbWF4V2lkdGg6ICI2MDBweCIgfX0-CiAgICA8Q2FyZD4KICAgICAgPFRoZW1lTmFtZUNvbnN1bWVyPgogICAgICAgIHtuYW1lID0-IDxIZWFkaW5nIGxldmVsPSIyIj5TaWduIGluIHRvIHtuYW1lfTwvSGVhZGluZz59CiAgICAgIDwvVGhlbWVOYW1lQ29uc3VtZXI-CiAgICAgIDxCb3ggY29tcG9uZW50PSJmb3JtIiBwYWRkaW5nVG9wPSJtZWRpdW0iPgogICAgICAgIDxUZXh0RmllbGQgaWQ9Im5hbWUiIGxhYmVsPSJVc2VybmFtZSIgdHlwZT0iZW1haWwiIHNlY29uZGFyeUxhYmVsPSJPciBlbWFpbCIgLz4KICAgICAgICA8VGV4dEZpZWxkCiAgICAgICAgICBpZD0icGFzc3dvcmQiCiAgICAgICAgICB0eXBlPSJwYXNzd29yZCIKICAgICAgICAgIGxhYmVsPSJQYXNzd29yZCIKICAgICAgICAgIHRlcnRpYXJ5TGFiZWw9ezxUZXh0TGluayBpbmxpbmU-Rm9yZ290IFBhc3N3b3JkPzwvVGV4dExpbms-fQogICAgICAgIC8-CiAgICAgIDwvQm94PgogICAgPC9DYXJkPgogIDwvZGl2Pgo8L2Rpdj4K)
- [Braid docs site](https://braid-design-system--fbe13fcdd487f70749ff26498253d9dc4996a87a.surge.sh/components/TextField)


lib/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,109 @@
...
+type RequiredInputProps = 'id' | 'value' | 'onChange';
+interface TextFieldProps
+  extends Required<Pick<InputProps, RequiredInputProps>> {
+  label?: string;
+  secondaryLabel?: ReactNode;
+  tertiaryLabel?: ReactNode;
+  placeholder?: string;
+  message?: ReactNode | false;
+  tone?: 'neutral' | 'critical' | 'positive';
+  type?:
```
mattsjones: Personally I think we should lock this down to text and maybe search? All the others will have some form of custom UI on top and are probably separate components. Thoughts?

michaeltaranto: Yeah I flipped a bit on this, at a minimum I think we should have: text, password, email, search, number, tell and url. Thoughts?

maraisr:
Out of curiosity, when following what semantic-release says, isn't this supposed to be `feat(TextField): Make initial component`? (fix => feat)

Are you guys more looking at it as a "fixing the bug, because textfield isn't there, and it's mean to be"?

patrickliu-sa:
Does this component also serve type as `email`,  `number`, or other types as well?

markdalgleish:
@maraisr Normally, yes—but for now we're trying to keep the version number in the v0.0.x range to signify that we're still in alpha.

michaeltaranto:
@patrickliu-sa yes at this stage we will support a subset of input types, including `password`, see [this list](https://github.com/seek-oss/braid-design-system/pull/133/files/768792c152df5c8b304ee9f13c58ed7ae85672fc#diff-d3aca9a721245d362d46e1457783622fR24), and as discussed on here and in slack we will likely remove the date and range ones from the list as a starting point

lib/components/FieldLabel/FieldLabel.css.js
```diff
@@ -0,0 +1,5 @@
+export default {
+  '.root': {
+    justifyContent: 'space-between',
```
markdalgleish: Hopefully we can delete this line one day 😄

lib/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,98 @@
...
+type RequiredInputProps = 'id' | 'value' | 'onChange';
+interface TextFieldProps
+  extends Required<Pick<InputProps, RequiredInputProps>> {
+  label?: string;
+  secondaryLabel?: ReactNode;
+  tertiaryLabel?: ReactNode;
+  placeholder?: string;
+  message?: ReactNode | false;
+  tone?: 'neutral' | 'critical' | 'positive';
+  type?: 'text' | 'password' | 'email' | 'search' | 'number' | 'tel' | 'url';
```
markdalgleish: As discussed, since this is a subset of allowed values, we should enforce this at runtime.

lib/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,108 @@
...
+              />
+              <Box
+                className={styles.focusOverlay}
+                boxShadow="outlineFocus"
+                borderRadius="standard"
+                paddingTop="standardTouchableText"
+                paddingBottom="standardTouchableText"
+              />
+            </Box>
+            {message === false ? null : (
```
markdalgleish: `FieldMessage` accepts `false` internally, you shouldn't have to perform this check.

michaeltaranto: Hmm interesting, does that feel right? Should this be a parent component concern? or happy to have it in `FieldMessage`? `FieldLabel` for example does not have that check internally

michaeltaranto: Agreed to internalise and also revisited logic for `FieldLabel` as part of this

PR #139, opened 2019-03-14
feat(Actions): Add Actions component

markdalgleish:
This PR adds the initial `Actions` component (formerly known as `ButtonGroup`), for rendering a responsive group of buttons/links at the bottom of a form/card/etc.

This PR also makes the following changes:
- Introduces a responsive `flexDirection` prop to `Box`, e.g. `<Box flexDirection={['column', 'row']}>`
- Fixes a bug with `none` values for margin and padding. The corresponding values in the CSS were incorrectly set to `none` rather than `0`.

### Example usage

```jsx
<Actions>
  <Button weight="strong">Submit</Button>
  <TextLink href="#">Cancel</TextLink>
<Actions>
```

### Mobile

Screen Shot 2019-03-15 at 10 07 55 am

### Desktop

Screen Shot 2019-03-15 at 10 07 28 am

## Development Notes

Internally, we've made the following changes.

- Moved `box-sizing: border-box` from `Box` into our CSS reset, so that link elements rendered by `TextLinkRenderer` use the correct box model.
- Introduced a new private component called `Overlay` to centralise rules for absolutely positioned cover elements, as well as a `FieldOverlay` component to centralise form overlay rules. This was done to ensure that the focus styles are consistent between `TextLinkRenderer` and `Button`, but the idea is that should expand this pattern to other fields going forward.
- Limited all `transition` atoms to transform and opacity.


lib/components/TextLinkRenderer/TextLinkRenderer.tsx
```diff
@@ -22,34 +24,66 @@ export default class TextLinkRenderer extends Component<TextLinkRendererProps> {
...
+        {({ atoms }) => (
+          <ActionsConsumer>
+            {inActions => {
+              const defaultStyles = [
+                styles.root,
+                atoms.reset.a,
+                atoms.color.link,
+              ];
+
+              if (inline) {
```
michaeltaranto: Thoughts on hoisting this to before the `ActionsConsumer`? If it's `inline` it has no contextual behaviour so can return fast.

lib/components/TextLinkRenderer/TextLinkRenderer.tsx
```diff
@@ -22,34 +24,66 @@ export default class TextLinkRenderer extends Component<TextLinkRendererProps> {
...
+                atoms.paddingBottom.standardTouchableText,
+              ];
+
+              if (inActions) {
+                return (
+                  <Text baseline={false}>
+                    <span className={styles.overlayContainer}>
+                      {children({
+                        style: {},
+                        className: classnames(defaultStyles, touchableStyles, [
```
michaeltaranto: Any reason for having an array as the third parameter here, rather than flat parameters?

lib/atoms/makeTransitions.ts
```diff
@@ -1,5 +1,7 @@
 export default () => ({
-  '.transition_fast': { transition: 'all .125s ease' },
+  '.transition_fast': {
+    transition: 'transform .125s ease, opacity .125s ease',
```
michaeltaranto: 💪

lib/components/TextLinkRenderer/TextLinkRenderer.tsx
markdalgleish: Ah yeah, nice idea 👍

lib/components/TextLinkRenderer/TextLinkRenderer.tsx
markdalgleish: More good ideas 👍

PR #157, opened 2019-03-31
feat(Alert): Add positive tone and weight modifier

mengtzu:
BREAKING CHANGE: `Alert` with `tone="info"` (the default tone)  has a lighter treatment by default.

### Description
The `Alert` component was previously too visually heavy for a lot of use cases. As a result we have implemented a `weight` scale of `regular` (default) and `strong`.

Also addressed in this PR is the accessibility of all the variants, ensuring that they meet the desired AA contrast ratio.

#### Info Alert weight
The `Alert` component was previously visually heavy when using `tone="info"`. The default is behaviour is a much lighter treatment, bringing it into line with the `critical` implementation. You can retain the previous behaviour by specifying the `weight` as `strong` via the prop:
```diff
- <Alert tone="info">
+ <Alert tone="info" weight="strong>
    My message
  </Alert>
```

#### Icon fills
The fill of all icons now defaults to `currentColor` ensuring that it takes on the colour of its parent. Worth validating all icons are visually as expected.

#### Theme changes
Aside from the addition of `info` to all themes, there were slight changes to certain theme values to ensure that they meet the desired AA contrast ratio.
- **JobStreet:** `critical` has been updated
- **SeekAnz:** `positive` has been updated
- **SeekAsia:** `positive` has been updated


mengtzu:
I've removed the weak tone variant, and added the Dark variant colours to `fill`.

(I possibly could have avoided the latter by using `currentColor`, let me know if that's better form)

PR #173, opened 2019-04-24
fix: Move min-height styles from Box to FieldMessage

markdalgleish:
## Breaking changes

- `Box` no longer has a `minHeight` prop.

## Development notes

I've introduced a `utils` collection that's passed into each themed block of CSS which contains `desktopStyles`, `responsiveStyles` and `rows` functions.

I've also refactored Icon styles to use theme utils where possible.


lib/atoms/makeSizes.ts
```diff
@@ -1,47 +1,7 @@
...
-  return merge({}, ...rules);
-};
 
 const width = {
   '.width_full': { width: '100%' },
 };
 
-export default (tokens: Tokens) =>
-  merge({}, makeSizeRules(tokens, 'minHeight'), width);
+export default () => merge({}, width);
```
mattsjones: Happy to leave the merge? lol

michaeltaranto: The merge is probably overkill here now.

lib/components/FieldMessage/FieldMessage.treat.ts
```diff
@@ -0,0 +1,33 @@
...
+    justifyContent: 'flex-end',
+  },
+  'root',
+);
+
+export const grow = style(
+  {
+    flexGrow: 1,
+  },
+  'grow',
```
mattsjones: If you used `css` instead we wouldn't need the debug name here. Personally, think that would be cleaner.

lib/components/FieldMessage/FieldMessage.tsx
```diff
@@ -25,7 +26,7 @@ const renderIcon = (tone: FieldTone = 'neutral') => {
   };
 
   return (
-    <Box paddingRight="xsmall" className={styles.fixedSize}>
+    <Box paddingRight="xsmall" className={useClassNames(styles.fixedSize)}>
```
mattsjones: I think `useStyles` would be cleaner here.

lib/components/FieldMessage/FieldMessage.treat.ts
markdalgleish: As discussed, gonna leave this it as-is until we optimise the output of themed `css` blocks.

lib/components/FieldMessage/FieldMessage.tsx
markdalgleish: Incorrect.

PR #175, opened 2019-05-07
fix(SeekAsia): Decrease touchable height and large text size

mengtzu:
Adjustments to SEEK Asia styles based on feedback from Joanne from reviewing the JobAlert work.

- Touchable height reduced from 11 rows to 10
- Large text size for desktop reverted to mobile size

### DIFF
--- a/lib/themes/seekAsia/tokens.ts
+++ b/lib/themes/seekAsia/tokens.ts
@@ -3,7 +3,7 @@ import { Tokens } from '../../themes/theme';
 const tokens: Tokens = {
   rowHeight: 4,
   columnWidth: 4,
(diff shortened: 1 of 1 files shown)


michaeltaranto:
Are we really ok with the touchable height being set below in the minimum recommended `44px`?

mengtzu:
I don't have a personal opinion on it, maybe start a conversation with Joanne in the design system slack channel?

michaeltaranto:
Covered off the concerns around the reduced touchable height as being acceptable for now. In the future we may introduce responsive touchable heights, at which time we can increase this on mobile.

PR #177, opened 2019-05-10
fix: Move fill styles from Box to Icon

markdalgleish:
BREAKING CHANGE: Box no longer has a ‘fill’ prop.

Also added a `name` property to all theme objects to enable better class minification in sku.


lib/components/icons/Icon/Icon.treat.ts
```diff
@@ -9,18 +10,18 @@ export const inline = style(
   'inline',
 );
 
-export const fill = style(
+export const fullHeight = style(
   {
     height: '100%',
   },
-  'fill',
+  'fullHeight',
```
mattsjones: Do we want to dump these now?

lib/components/icons/Icon/Icon.treat.ts
```diff
@@ -49,3 +50,11 @@ export const blockSizes = css(theme => {
     ),
   };
 }, 'blockSizes');
+
+export const fill = css(
+  ({ color }) => ({
+    ...mapTokensToStyleProperty(color.foreground, 'fill'),
+    currentColor: { fill: 'currentColor' },
+  }),
+  'fill',
```
mattsjones: Same here

PR #186, opened 2019-05-20
feat: Add support for disabling TextField, TextArea & Dropdown

michaeltaranto:
Bringing the suite of fields up to the same standard as inline fields, adding support for disabling and making a few impossible states impossible, eg. a field can no longer be disabled & critical.

This migrates `Field` to treat and centralises a bunch of field formatting to be shared between all consumers, eg. `TextField`, `TextArea` & `Dropdown`.

Also fixes a bug with `Checkbox` & `Radio`, when `disabled` they would collapse their `message` whitespace rather than blank out the message.


lib/components/private/Field/Field.treat.ts
```diff
@@ -0,0 +1,25 @@
...
+  position: 'relative',
+});
+
+export const field = style({
+  outline: 'none',
+});
+
+export const focusOverlay = style({
+  selectors: {
+    [`${field}:focus ~ &`]: {
```
markdalgleish: Still get excited seeing this pay off 😄

PR #194, opened 2019-05-23
fix(Box): Extract styles to hook

michaeltaranto:


lib/hooks/useBox/index.ts
```diff
@@ -0,0 +1,138 @@
...
+  transform,
+  width,
+}: BoxProps) => {
+  const { atoms } = useTheme();
+
+  return useClassNames(
+    atoms.backgroundColor[backgroundColor!],
+    atoms.boxShadow[boxShadow!],
+    atoms.borderRadius[borderRadius!],
+    atoms.boxShadow[boxShadow!],
```
mattsjones: Dump the repeated `boxShadow`?

michaeltaranto: Totally! 😄 Done

PR #196, opened 2019-05-23
refactor: Co-locate all selectors with their target element

markdalgleish:


mattsjones:
This is nice. It makes me wonder if we should have utilities to create the selectors though. For example `childOf(container)`?

PR #198, opened 2019-05-24
fix(Box): Migrate backgroundColor to treat

michaeltaranto:


lib/hooks/useBox/box.treat.ts
```diff
@@ -119,3 +119,7 @@ export const flexDirectionDesktop = css(({ utils: { desktopStyles } }) =>
     }),
   ),
 );
+
+export const backgroundColor = css(({ color: { background } }) =>
+  mapToCssRule(background, 'backgroundColor'),
+);
```
markdalgleish: 👌

lib/components/Box/Box.tsx
```diff
@@ -1,7 +1,11 @@
 import React from 'react';
-import useBox, { BoxProps } from '../../hooks/useBox';
+import useBox, { UseBoxProps } from '../../hooks/useBox';
 import { Reset } from '../Reset/Reset';
 import { ContrastProvider } from './ContrastContext';
+import { ResetProps } from '../../components/Reset/Reset';
```
mattsjones: Imported this file twice

docs/src/App/Home/Home.tsx
```diff
@@ -1,19 +1,22 @@
...
-  display: 'flex',
-  width: 'full',
-  marginLeft: ['none', 'xxsmall'],
-  marginRight: ['none', 'xxsmall'],
-  paddingBottom: ['xsmall', 'none'],
-};
+const Action = ({ children }: { children: ReactNode }) => (
+  <Box
+    display={'flex'}
+    width={'full'}
```
mattsjones: `width="full"` ?

lib/themes/makeTreatTheme.ts
```diff
@@ -45,6 +65,16 @@ const decorateTheme = (vars: ThemeVars) => ({
...
     },
+    background: {
+      ...vars.color.background,
+      formAccentActive: getActiveColor(vars.color.background.formAccent),
+      formAccentHover: getHoverColor(vars.color.background.formAccent),
+      brandAccentActive: getActiveColor(vars.color.background.brandAccent),
+      brandAccentHover: getHoverColor(vars.color.background.brandAccent),
+      infoLight: getLightVariant(vars.color.background.info),
+      criticalLight: getLightVariant(vars.color.background.critical),
+      positiveLight: getLightVariant(vars.color.background.positive),
```
mattsjones: I wonder if we should put these values on the theme or just the util itself and calculate where it's needed?

docs/src/App/Home/Home.tsx
michaeltaranto: Sorry, i was relying on prettier fixing my multi-cursor refactor laziness haha

mattsjones: I assumed that'd be the case

lib/themes/makeTreatTheme.ts
michaeltaranto: Hmm, i like the thinking, let me see what the implications are.

PR #200, opened 2019-05-24
fix: Migrate color and fill to useTextColor hook

michaeltaranto:


lib/hooks/typography/typography.treat.ts
```diff
@@ -85,3 +88,34 @@ export const heading = {
...
+
+  return {
+    ...mapTokensToStyleProperty(foreground, 'color'),
+    link: {
+      color: link,
+      ...(link !== linkHover
+        ? {
+            selectors: {
+              '&:hover': { color: linkHover },
+              '&:focus': { color: linkHover },
```
mattsjones: Use simple pseudos here dummies! 😉

PR #201, opened 2019-05-27
refactor: Centralise mapToStyleProperty and utils

michaeltaranto:
Upgrades `@types/lodash` to pull in fix for `mapValues` typings. This means we can clean up some work arounds in the code base. Taking the opportunity to centralise on a common `mapToStyleProperty` helper, while also moving utils that are hanging around beyond the atoms work into the top level utils folder.


package.json
```diff
@@ -43,8 +46,14 @@
...
-    "**/*.{js}": ["yarn lint", "git add"],
-    "**/*.{js,md,less}": ["yarn format", "git add"]
+    "**/*.{js}": [
+      "yarn lint",
+      "git add"
+    ],
+    "**/*.{js,md,less}": [
+      "yarn format",
+      "git add"
+    ]
```
mattsjones: The `package.json` format wars continue lol

yarn.lock
```diff
@@ -2076,9 +2076,9 @@
     "@types/react" "*"
 
 "@types/lodash@^4.14.123":
```
mattsjones: Should we update the required dependency as well now? Otherwise, prop types might not get generated for consumers with stale lockfiles.

lib/hooks/useBox/box.treat.ts
```diff
@@ -1,22 +1,10 @@
...
-  type MappedStyles = Record<Map, Styles>;
-
-  return mapValues(map, (value: Value) =>
-    mapper ? mapper(value, propertyName) : { [propertyName]: value },
-  ) as MappedStyles;
-};
+import { getLightVariant } from '../../utils/a11y';
+import isLight from '../../utils/isLight';
+import { px } from '../../utils/toUnit';
+import mapToStyleProperty from '../../utils/mapToStyleProperty';
```
mattsjones: What are your thoughts on having a central `util` export that gives you all of our utils? I reckon `import { px, getLightVariant, isLight, px, mapToStyleProperty } from '../../utils';` would be easier to handle/parse.

michaeltaranto: I have a few stacked pieces of work, are you happy if we do this as a refactor after the migration work?

PR #202, opened 2019-05-27
feat(Box): Deprecate standardTouchableText paddings

michaeltaranto:
BREAKING CHANGE: Deprecate the `standardTouchableText` padding from `Box`.

Removing this from the public api and have provided this internally for use through a hook. If we require this in some way for consumers we can think through the best api to expose then.


lib/hooks/typography/typography.treat.ts
```diff
@@ -117,3 +119,27 @@ export const color = css(theme => {
     },
   };
 });
+
+const touchableSpace = (
+  theme: Theme,
+  breakpoint: Breakpoint,
+  isDesktop?: boolean,
```
markdalgleish: Shall we remove this in favour of checking the `breakpoint` arg?

PR #212, opened 2019-05-31
fix(Icon): Add 'small' Icon size, tidy up typography

michaeltaranto:
Missed adding `small` support to inline icons because the icons were hard coded. This refactors the icons to be generated from the theme definition for text which should prevent this in the future. Also made the font weights generated from the theme and cleaned up the touchableSpace rules.

### DIFF
--- a/lib/hooks/typography/typography.treat.ts
+++ b/lib/hooks/typography/typography.treat.ts
@@ -9,17 +9,9 @@ export const fontFamily = style(({ typography }) => ({
   fontFamily: typography.fontFamily,
 }));
 
-export const fontWeight = css(({ typography }) => ({
-  regular: {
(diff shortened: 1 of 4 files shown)


lib/hooks/typography/typography.treat.ts
```diff
@@ -9,17 +9,9 @@ export const fontFamily = style(({ typography }) => ({
...
-  },
-  medium: {
-    fontWeight: typography.fontWeight.medium,
-  },
-  strong: {
-    fontWeight: typography.fontWeight.strong,
-  },
-}));
+export const fontWeight = css(({ typography }) =>
+  mapToStyleProperty(typography.fontWeight, 'fontWeight'),
```
markdalgleish: 👏 😍

PR #220, opened 2019-06-05
test: Add public API contract test, improve docs site type formatting

mattsjones:


package.json
```diff
@@ -9,11 +9,12 @@
   "scripts": {
     "check-integrity": "yarn check --integrity && echo 'Ignore paths from lint-staged'",
     "pretest": "sku configure",
-    "test": "yarn check --integrity && yarn lint && yarn build",
+    "test": "yarn check --integrity && yarn lint && yarn build && yarn contract-test",
+    "contract-test": "sku test",
```
michaeltaranto: Given the name of this command i would recommend explicitly passing the path to the contract test here.

lib/componentDocs/generateComponentDocs.js
```diff
@@ -1,45 +0,0 @@
...
-});
-
-const omitDocWithNoProps = ({ props = {} }) => Object.keys(props).length > 0;
-const omitNonComponentDocs = ({ displayName }) => /^[A-Z]/.test(displayName);
-
-(async () => {
-  const componentDocs = await globsToReactDocgen({
-    include: ['lib/themes/**/*.{ts,tsx}', 'lib/components/**/*.{ts,tsx}'],
-    exclude: [
-      'lib/components/private/**/*',
```
michaeltaranto: Love that we dont need to manage exclusions now, its solely testing the public api 🙌

generate-component-docs/__snapshots__/generate.test.ts.snap
```diff
@@ -0,0 +1,6618 @@
...
+    },
+  },
+}
+`;
+
+exports[`Public API Box 1`] = `
+Object {
+  "props": Object {
+    "about": Object {
+      "propName": "about",
```
michaeltaranto: Any thoughts on being able to apply the breaks on components like this and not traverse every html prop?

mattsjones: I thought about it. It's probably doable. But it's still part of our contract. Was thinking we could see how it goes.

generate-component-docs/__snapshots__/generate.test.ts.snap
```diff
@@ -0,0 +1,6618 @@
...
+      "required": false,
+      "type": "string",
+    },
+    "rowSpan": Object {
+      "propName": "rowSpan",
+      "required": false,
+      "type": "number",
+    },
+    "rows": Object {
+      "propName
```
michaeltaranto: `ReactNode` might be something else to not unpack, from a docs site perspective things can get noisy here.

PR #221, opened 2019-06-06
fix(Box): Correct component prop type

mattsjones:


lib/hooks/useBox/index.ts
```diff
@@ -7,7 +6,7 @@ import * as styleRefs from './box.treat';
 type ResponsiveProp<AtomName> = AtomName | [AtomName, AtomName];
 
 export interface UseBoxProps {
-  component: ReactType;
+  component: keyof JSX.IntrinsicElements;
```
markdalgleish: Should this also be a runtime check? Otherwise the types are lying.

markdalgleish:
Closing for now.

PR #226, opened 2019-06-11
feat(Text): Default Text to ‘span’, support id prop, add Paragraph

markdalgleish:


lib/components/Text/Text.tsx
```diff
@@ -14,7 +15,8 @@ export interface TextProps extends Pick<BoxProps, 'component'> {
 }
 
 export const Text = ({
-  component,
+  id,
+  component = 'span',
```
michaeltaranto: Worth combing the codebase for where were explicitly setting this as we can now remove those instances.

PR #236, opened 2019-06-20
feat(text): Simplify text/icon color API

markdalgleish:
BREAKING CHANGE:

Removes the following text/icon `color` values:
- `brandAccent`
- `formAccentDisabled`
- `criticalContrast`
- `positiveContrast`
- `infoContrast`
- `brandAccentForeground`

`brandAccent` and `formAccentDisabled` are currently unused.

The remaining values are all contrast/foreground variants, which are now provided automatically based on the nearest `Box` parent that sets a `backgroundColor`.

For example `<Box backgroundColor="formAccent"><Text>Hello world</Text></Box>` will automatically render white text if `formAccent` is dark. If needed, this can be overridden by explicitly setting a `color` prop on `Text`.

## Development Notes

I've also refactored `Alert` and `Button` to take advantage of this new API.


lib/components/Button/Button.tsx
```diff
@@ -41,12 +41,6 @@ const backgroundColor: Record<
   },
 };
 
-const foregroundColor: Record<ButtonWeight, TextProps['color']> = {
-  weak: 'formAccent',
-  regular: 'white',
-  strong: 'brandAccentForeground',
-};
```
michaeltaranto: :wave:

lib/components/icons/Icon/Icon.tsx
```diff
@@ -42,7 +42,7 @@ export const Icon = ({
       className={classnames(
         resolveSizeClasses(size, inline),
         styles.currentColor,
-        fill ? useTextColor(fill) : null,
+        useTextColor(fill),
```
markdalgleish: I didn't even realise that this fixes a bug where icons didn't have a `fill` set by default, so they ended up rendering with a black fill rather than `neutral`.

You can thank Chromatic for pointing this out 😄

PR #237, opened 2019-06-25
feat(MonthPicker): Add component

michaeltaranto:
### Description
Introduce `MonthPicker` component, providing a custom `Dropdown` experience on desktop, while rendering the native month picker input on tablet and mobile devices for better UX. The component handles the toggling internally based on the user-agent but would like to create a context that can resolve the UA from the `request` object to support server rendering.

#### Other changes
- Allowing `aria-describedby` overrides on existing fields, to support custom relationships between `FieldMessage` and custom braid components.
- Introduces [`is-mobile`](https://github.com/juliangruber/is-mobile) for UA sniffing. Weighs in at 0.4kb gzipped and supports the server side checking which is consistent with future plans.

### Preview Links
- [Braid Site](https://braid-design-system--d10851f2ee25fd0fffe9287f330c6eb1b7c18c4d.surge.sh/components/MonthPicker)
- [Playroom](https://braid-design-system--d10851f2ee25fd0fffe9287f330c6eb1b7c18c4d.surge.sh/playroom/#?code=PENhcmQ-CiAgPE1vbnRoUGlja2VyIGxhYmVsPSJEZWZhdWx0IHBpY2tlciIgaWQ9InN0YXJ0ZWQiIC8-CiAgPE1vbnRoUGlja2VyCiAgICBpZD0iZmluaXNoIgogICAgbGFiZWw9IkludmFsaWQgcGlja2VyIgogICAgdG9uZT0iY3JpdGljYWwiCiAgICBtZXNzYWdlPSJSZXF1aXJlZCBmaWVsZCIKICAvPgogIDxNb250aFBpY2tlciBpZD0iZGlzYWJsZWQiIGxhYmVsPSJEaXNhYmxlZCBwaWNrZXIiIGRpc2FibGVkPXt0cnVlfSAvPgogIAogIDxQYXJhZ3JhcGg-CiAgICA8VGV4dCBjb2xvcj0ic2Vjb25kYXJ5IiBzaXplPSJzbWFsbCI-Tm90ZTogU2VsZWN0aW5nIHZhbHVlcyB3b24ndCB3b3JrIG9uIG5vbi1tb2JpbGUvdGFibGV0IGRldmljZXMgYXMgdGhlIGRyb3Bkb3duIHZlcnNpb24gaXMgYSBjb250cm9sbGVkIGZpZWxkIG9ubHkgZHVlIHRvIGl0J3MgY3VzdG9tIGB2YWx1ZWAgaW50ZXJmYWNlLjwvVGV4dD4KICA8L1BhcmFncmFwaD4KPC9DYXJkPgoK)


lib/components/MonthPicker/MonthPicker.migration.md
```diff
@@ -0,0 +1,27 @@
+# MonthPicker Migration Guide
```
markdalgleish: 👏

lib/components/MonthPicker/MonthPicker.migration.md
```diff
@@ -0,0 +1,27 @@
...
+
+## API Changes
+
+- `native={boolean}` has been deprecated in favour of making this decision internally, so no need to explicitly set this.
```
mattsjones: Prefer `removed` to `deprecated` as it's actually gone.

lib/components/MonthPicker/MonthPicker.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+interface MonthPickerValue {
+  month: number | undefined;
+  year: number | undefined;
+}
+
+interface MonthPickerProps
+  extends Omit<FieldProps, 'autoComplete' | 'secondaryMessage'> {
+  value: NonNullable<MonthPickerValue>;
+  onChange: NonNullable<(value: MonthPickerValue, event: ChangeEvent) => void>;
```
mattsjones: What is `NonNullable` doing here?

lib/components/MonthPicker/MonthPicker.tsx
```diff
@@ -0,0 +1,265 @@
...
+const renderNativeInput = isMobile({ tablet: true });
+const monthRef = createRef<HTMLSelectElement>();
+const yearRef = createRef<HTMLSelectElement>();
+
+type FieldType = keyof MonthPickerValue | 'native';
+
+/* ---------------------------- */
+/* Custom value parsing helpers */
+/* ---------------------------- */
+const customValueToString = ({ month, year }: MonthPickerProps['value']) =>
```
mattsjones: Personally, I'd rather use `MonthPickerValue` rather than `MonthPickerProps['value']`.

lib/components/MonthPicker/MonthPicker.tsx
```diff
@@ -0,0 +1,253 @@
...
+    month: parseInt(month, 10),
+    year: parseInt(year, 10),
+  };
+};
+
+type FieldType = keyof MonthPickerValue | 'native';
+
+const nativeToCustomValue = (
+  nativeValue: string,
+  value: MonthPickerProps['value'],
```
mattsjones: Same here

lib/components/MonthPicker/MonthPicker.tsx
```diff
@@ -0,0 +1,253 @@
...
+  { value: '5', label: 'May' },
+  { value: '6', label: 'Jun' },
+  { value: '7', label: 'Jul' },
+  { value: '8', label: 'Aug' },
+  { value: '9', label: 'Sep' },
+  { value: '10', label: 'Oct' },
+  { value: '11', label: 'Nov' },
+  { value: '12', label: 'Dec' },
+];
+const getMonths = () =>
```
mattsjones: Can we make this a component? `Months`?

lib/components/MonthPicker/MonthPicker.tsx
```diff
@@ -0,0 +1,253 @@
...
+  months.map(month => (
+    <option value={month.value} key={month.value}>
+      {month.label}
+    </option>
+  ));
+
+const getYears = (min: number, max: number, ascending: boolean) => {
```
mattsjones: `Years`?

lib/components/MonthPicker/MonthPicker.tsx
```diff
@@ -0,0 +1,265 @@
...
+  { value: '6', label: 'Jun' },
+  { value: '7', label: 'Jul' },
+  { value: '8', label: 'Aug' },
+  { value: '9', label: 'Sep' },
+  { value: '10', label: 'Oct' },
+  { value: '11', label: 'Nov' },
+  { value: '12', label: 'Dec' },
+];
+const getMonths = () =>
+  months.map(m => (
```
markdalgleish: Could we name this argument `month`?

lib/components/MonthPicker/MonthPicker.tsx
```diff
@@ -0,0 +1,265 @@
...
+      month: value && value.month ? value.month : undefined,
+      year: parseInt(nativeValue, 10) || undefined,
+    },
+    native: stringToCustomValue(nativeValue),
+  }[fieldType]);
+
+/* ---------------------------- */
+/* Event handler management     */
+/* ---------------------------- */
+const handleChange = <E extends HTMLSelectElement | HTMLInputElement>(
```
markdalgleish: Could we name these types `Element` rather than `E`?

lib/components/MonthPicker/MonthPicker.tsx
michaeltaranto: Relic from refactor, i'll bin it

lib/components/MonthPicker/MonthPicker.tsx
```diff
@@ -0,0 +1,252 @@
...
+        borderRadius,
+        className,
+        ...nonVisualFieldProps
+      }) =>
+        renderNativeInput ? (
+          <Box
+            component="input"
+            type="month"
+            value={customValueToString(currentValue)}
+            onChange={onChange && handleChange({ onChange, value }, 'native')}
```
markdalgleish: I thought this was a bug because it looked like it was firing the change handler immediately during render. I feel like these would be better named as something like `makeChangeHandler`. What do you think?

lib/components/MonthPicker/MonthPicker.tsx
```diff
@@ -0,0 +1,252 @@
...
+    >
+      {({
+        backgroundColor,
+        boxShadow,
+        width,
+        paddingLeft,
+        paddingRight,
+        borderRadius,
+        className,
+        ...nonVisualFieldProps
```
markdalgleish: I wonder if this is a sign we've got the abstraction slightly wrong. Right now, `MonthPicker` needs to keep a running list of all visual props and keep them in sync with `Field`. In the future we might add another visual prop within `Field` and it would accidentally flow through into this `nonVisualFieldProps` object. Not sure what the answer is, though—might need to discuss this one.

markdalgleish: I noticed you added comments to `Field` to separate visual from non-visual props. I wonder if we should expose these as different objects, so consumers can safely ignore visual props if needed, e.g. `{ visualProps: { ... }, nonVisualProps: { ... } }`. Could obviously use a bikeshed.

lib/components/MonthPicker/MonthPicker.tsx
```diff
@@ -0,0 +1,252 @@
...
+const handleBlur = <Element extends HTMLSelectElement | HTMLInputElement>(
+  { onBlur, value }: Pick<MonthPickerProps, 'onBlur' | 'value'>,
+  fieldType: FieldType,
+) => (event: FocusEvent<Element>) => {
+  if (typeof onBlur === 'function') {
+    const isRelatedTargetExternal =
+      event.relatedTarget !== monthRef.current &&
+      event.relatedTarget !== yearRef.current;
+
+    if (event.relatedTarget === null || isRelatedTargetExternal) {
```
markdalgleish: This is neat 😍

lib/components/MonthPicker/MonthPicker.tsx
```diff
@@ -0,0 +1,252 @@
...
+      onBlur(nativeToCustomValue(event.target.value, value, fieldType), event);
```
markdalgleish: Why do we pass the value to the blur handler? Is it deliberate that we're also *not* passing it to the focus handler?

michaeltaranto: As discussed, it was for backwards compatibility with previous style guides, but we have decided to not pass any args to the `blur` and `focus` handlers, while also only passing the custom value to the `change` handler.

lib/components/MonthPicker/MonthPicker.tsx
michaeltaranto: As discussed we are going to look at addressing this by introducing a `Fieldset` component, which suits this use case having multiple children components — whereas `Field` is really for one-to-one mappings

PR #239, opened 2019-06-26
test(chromatic): Merge snapshots into a single story per component

markdalgleish:
This should greatly reduce our Chromatic snapshot usage, while also streamlining the review process.

Each component example has a minimum height of `300px`, which should solve the biggest issue with combined screenshots where updates to components higher in the page can cause cascading diffs to the rest of the components.

We're also rendering a divider line below each components so that we can catch differences in surrounding white space, which is otherwise invisible.

I've also removed the `wireframe` theme, since it's not a production theme and any accidental breakages wouldn't really be a big deal.

### DIFF
--- a/lib/stories/all.stories.tsx
+++ b/lib/stories/all.stories.tsx
@@ -21,23 +21,54 @@ req.keys().forEach(filename => {
   const stories = storiesOf(componentName, module);
   const docs = req(filename).default as ComponentDocs;
 
-  if (docs.storybook === false) {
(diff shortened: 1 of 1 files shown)


markdalgleish:
Edit: Also removed the `wireframe` theme, since it's not a production theme and any accidental breakages wouldn't really be a big deal.

markdalgleish:
Now a 78% reduction! 😄🎉

PR #246, opened 2019-06-27
test: Improve API snapshots

mattsjones:


generate-component-docs/generate.ts
```diff
@@ -9,6 +9,14 @@ const propBlacklist = ['key'];
 const tsconfigPath = path.join(__dirname, '../tsconfig.json');
 const componentsFile = path.join(__dirname, '../lib/components/index.ts');
 
+const reactNodeType =
+  'string | number | boolean | {} | ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<any, any, any>)> | ReactNodeArray | ReactPortal';
```
mattsjones: This is a bit of a hack and won't work if the type changes in any way. I think just keep an eye on it as it's much cleaner with the shorthand.

generate-component-docs/generate.ts
```diff
@@ -9,6 +9,14 @@ const propBlacklist = ['key'];
...
+
+const stringAliases: Record<string, string> = {
+  [reactNodeType]: 'ReactNode',
+  boolean: 'boolean',
```
markdalgleish: Why is this needed?

mattsjones: This was basically already in the code. Technically it seems a Boolean is a union of true and false. This just makes it show as the word boolean.

PR #259, opened 2019-07-08
refactor: Use YIQ color space to determine if color is light

michaeltaranto:
Previously we implemented custom logic for determining if a colour was light that was simply a subjective adjustment of a magic number luminance threshold. As it turns out there is an alternative research-based approach that leverages the [YIQ colour space](https://en.wikipedia.org/wiki/YIQ).

For further reading checkout this post on [Calculating Color Contrast](https://24ways.org/2010/calculating-color-contrast/).

Thanks to @patrickliu-sa for pointing us in this direction.

### DIFF
--- a/lib/utils/isLight/index.ts
+++ b/lib/utils/isLight/index.ts
@@ -1,4 +1,4 @@
-import { getLuminance } from 'polished';
+import { parseToRgb } from 'polished';
 // @ts-ignore
 import { parse as parseGradient } from 'gradient-parser';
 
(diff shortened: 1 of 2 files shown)


michaeltaranto:
Temporarily closing while we discuss some branding impacts.

michaeltaranto:
Re-opening this PR as we have added the ability to handle case specific overrides, while having all the benefits of handling contrast ratios correctly by default.

lib/utils/isLight/index.ts
```diff
@@ -28,5 +28,22 @@ export const isLight = (inputColor: string) => {
     ? getLinearGradientColors(inputColor)
     : [inputColor];
 
-  return colors.some(color => getLuminance(color) > 0.6);
+  return colors.some(color => {
+    const { red, green, blue } = parseToRgb(color);
+
+    // Convert RGB to YIQ to better take into account the
+    // luminence of the separate color channels:
```
markdalgleish:
```suggestion
    // luminance of the separate color channels:
```

PR #266, opened 2019-07-12
docs: Improve code snippet experience

mattsjones:


site/src/App/Code/Code.treat.ts
```diff
@@ -0,0 +1,48 @@
+import { style } from 'sku/treat';
+
+export const button = style({
+  cursor: 'pointer',
+  outline: 'none',
+});
+
+const overlay = style({
+  transition: 'opacity 0.2s',
```
michaeltaranto: `FieldOverlay` already sets `transition="fast"` so this should be redundant.

site/src/App/Code/Code.tsx
```diff
@@ -0,0 +1,101 @@
...
+        background="formAccentHover"
+        className={classnames(styles.hoverOverlay)}
+      />
+      <FieldOverlay
+        background="formAccentActive"
+        className={classnames(styles.activeOverlay)}
+      />
+      <Box
+        component="span"
+        // display="block"
```
michaeltaranto: 🗑

site/src/App/Code/Code.treat.ts
markdalgleish: I copied this from our button code, so this redundancy is duplicated there too.

site/src/App/Code/Code.treat.ts
```diff
@@ -0,0 +1,53 @@
+import { style } from 'sku/treat';
+
+export const code = style({
+  overflowX: 'auto',
+  borderBottomLeftRadius: '0 !important',
```
markdalgleish: Braid isn't currently good at this 😞

site/src/App/Code/Code.treat.ts
markdalgleish: Cleaned this up in a separate PR: https://github.com/seek-oss/braid-design-system/pull/267

site/src/App/Code/Code.treat.ts
michaeltaranto: Yeah, when i have faced this though rather than use `important` i have opted out of the box prop usage, eg. only apply border radius in the local treat file to prevent need for overriding. But not too fussed about the docs site.

markdalgleish: Might be a good idea to consider atomic border width classes.

markdalgleish: Or maybe a way of opting out of border radius at the `Box` level, since you're not really meant to mix and match different radius values.

PR #278, opened 2019-07-29
fix: Make field text visible within dark background context

markdalgleish:
This fixes an issue where text within a `TextField`, `Dropdown` or `Textarea` was white when nested within a `Box` with a dark background (e.g. `<Box background="brand">`).

This PR also ensures that text provides adequate contrast against all `Box` backgrounds.

Before: 
Screen Shot 2019-07-29 at 4 26 04 pm

After:
Screen Shot 2019-07-29 at 4 26 14 pm

## Development Notes

This issue was caused by the way `Field` works, where we pass classes down to the target field via render props. As a result, the white background of the field wasn't available on context within the `useTextTone` hook, so it was using the background context defined higher in the tree.

To fix this issue, the `useText`, `useHeading` and `useTextTone` hooks now support an optional `backgroundContext` override for times like this where it's not possible to provide a missing context value when using hooks.

I've also added some examples and test cases to make sure that this issue is caught by our test suite, and to make sure that text contrast in general is covered by screenshot tests. In fact, adding this test highlighted some dormant issues with certain combinations, which I've also fixed in this PR.

Because some of the test cases are much more exhaustive than typical documentation, I've added an optional `docsSite` boolean to the examples type which allows you to hide certain examples from the website. These examples will still show up in Storybook, which means that you'll still see them as part of our Chromatic suite.

### DIFF
--- a/site/src/App/Components/ComponentRoute.tsx
+++ b/site/src/App/Components/ComponentRoute.tsx
@@ -61,65 +61,57 @@ export const ComponentRoute = ({
           </Text>
         </Box>
       ) : null}
-      {examples.map((example, index) => {
-        const {
-          label,
-          docsSite = true,
-          render,
-          code,
-          Container = DefaultContainer,
-        } = example;
+      {examples
+        .filter(example => example.docsSite !== false)
+        .map((example, index) => {
+          const { label, render, code, Container = DefaultContainer } = example;
 
-        if (!docsSite) {
-          return null;
-        }
+          const codeAsString =
+            render && !code
+              ? cleanCodeSnippet(
+                  reactElementToJSXString(render({ id: 'id', handler }), {
+                    useBooleanShorthandSyntax: false,
+                    showDefaultProps: false,
+                    showFunctions: false,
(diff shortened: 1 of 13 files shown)


markdalgleish:
Before: 
Screen Shot 2019-07-29 at 4 26 04 pm

After:
Screen Shot 2019-07-29 at 4 26 14 pm

site/src/App/Components/ComponentRoute.tsx
```diff
@@ -61,55 +61,65 @@ export const ComponentRoute = ({
...
-                      <Box paddingBottom="small">
-                        <Text tone="secondary">Theme: {theme.name}</Text>
-                      </Box>
-                      <BraidProvider theme={theme}>
-                        <Container>
-                          {render({ id: `${index}_${theme.name}`, handler })}
-                        </Container>
-                      </BraidProvider>
+        if (!docsSite) {
+          return null;
```
michaeltaranto: Thoughts on doing this upfront as a `filter`?

PR #289, opened 2019-08-19
fix(Typography): Update header and text tokens

patrickliu-sa:
Align heading and text size and rows with the following setting agreed with Joanne
image
Note that the tokens are specific to jobsDB and JobStreet for now.


patrickliu-sa:
Close it for now as there might be a more comprehensive token changes will occur in one go due to rebranding

PR #291, opened 2019-08-20
fix(ChevronIcon): Icon retouch

patrickliu-sa:
Update chevron icon with better stroke thickness and better centering for icon rotation


mattsjones:
Hey @patrickliu-sa, 👋 

We're currently adding a full suite of cohesive icons to braid that come from a set template. Chevron will be updated as part of that. 

If you're having icon alignment issues please hit up #braid-support. If you want to propose changes to the design system, please post in #braid-contributors to discuss first.

PR #293, opened 2019-08-21
feat(Box): Add shorthand props for margin and padding

markdalgleish:
This PR adds the following props to `Box`:

- `padding`
- `paddingX`
- `paddingY`
- `margin`
- `marginX`
- `marginY`

## Migration Guide

Please note that this is entirely optional. The old longhand form will still work

```diff
-<Box paddingLeft="small" paddingRight="small">
+<Box paddingX="small">

-<Box paddingTop="small" paddingBottom="small">
+<Box paddingY="small">

-<Box
-  paddingTop="small"
-  paddingBottom="small"
-  paddingLeft="small"
-  paddingRight="small">
+<Box padding="small">
```


lib/components/private/Field/Field.tsx
```diff
@@ -37,8 +37,7 @@ interface FieldRenderProps extends Pick<FieldProps, PassthroughProps> {
   boxShadow: BoxProps['boxShadow'];
   borderRadius: BoxProps['borderRadius'];
   width: BoxProps['width'];
-  paddingLeft: BoxProps['paddingLeft'];
-  paddingRight: BoxProps['paddingRight'];
+  paddingX: BoxProps['paddingLeft'];
```
markdalgleish:
```suggestion
  paddingX: BoxProps['paddingX'];
```

lib/components/Dropdown/Dropdown.tsx
```diff
@@ -62,14 +62,11 @@ const NamedDropdown = forwardRef<HTMLSelectElement, DropdownProps>(
...
-        {(
-          { className, paddingLeft, paddingRight, ...fieldProps },
-          fieldRef,
-        ) => (
+        {({ className, paddingX, ...fieldProps }, fieldRef) => (
           <Fragment>
             <Box
               component="select"
-              paddingLeft={paddingLeft}
+              paddingLeft={paddingX}
```
mattsjones:
```suggestion
              paddingX={paddingX}
```

markdalgleish: No, this was actually deliberate. It's applying the horizontal padding, but only on the left side.

mattsjones: Weird...

markdalgleish: It's because it's applying the right padding to a different element.

PR #294, opened 2019-08-23
build: Automatically generate icons from SVGs

markdalgleish:
All SVG components are now gitignored. Instead of being checked in, they're generated from the raw SVG files automatically during the usual dev and build lifecycles. This process can be triggered manually via `yarn setup`.

This PR also introduces a snapshot test to validate that the generated icons match our expectations. Going forwards, any change to the generated SVG components need to be reflected in the snapshot.


scripts/generateIcons.js
```diff
@@ -0,0 +1,197 @@
...
+
+      // Bail out early if we're processing an icon variant (e.g. bookmark-active.svg)
+      // All subsequent steps should only happen once per icon component.
+      if (variantName) {
+        return;
+      }
+
+      const templateFileIfMissing = async (relativePath, contents) => {
+        const filePath = path.join(iconDir, relativePath);
+        if (!(await fs.exists(filePath))) {
```
michaeltaranto: Just looked into this and `fs-extra` provide the `fs.exists` api even though it's deprecated in node. They also export a first class api called [`pathExists`](https://github.com/jprichardson/node-fs-extra/blob/master/docs/pathExists.md), which is probably a safer option in that it's less likely to get deprecated.

scripts/generateIcons.js
```diff
@@ -0,0 +1,197 @@
...
+        }
+      };
+
+      // Create icon wrapper component, if it doesn't already exist
+      await templateFileIfMissing(
+        `${iconName}.tsx`,
+        dedent`
+          import React from 'react';
+          import { Box } from '../../Box/Box';
+          import useIcon, { UseIconProps } from '../../../hooks/useIcon';
```
michaeltaranto: I realise the build would catch incorrectly generated files, but could we be more certain about these paths by using `path.relative` and templating the result in. Thoughts?

scripts/generateIcons.js
```diff
@@ -0,0 +1,197 @@
...
+          Please refer to the [Icon Migration Guide.](../Icon.migration.md)
+        `,
+      );
+    }),
+  );
+
+  // Create icons/index.ts
+  const files = await fs.readdir(iconComponentsDir);
+  const iconComponents = files.filter(
+    file => file.includes('Icon') && !file.includes('.'),
```
michaeltaranto: `file.endsWith('Icon.tsx')`?

scripts/generateIcons.js
```diff
@@ -0,0 +1,197 @@
...
+        `${iconName}.tsx`,
+        dedent`
+          import React from 'react';
+          import { Box } from '../../Box/Box';
+          import useIcon, { UseIconProps } from '../../../hooks/useIcon';
+          import { ${svgComponentName} } from './${svgComponentName}';
+
+          export type ${iconName}Props = UseIconProps;
+
+          export const ${iconName} = (props: UseIconProps) => {
```
mattsjones: This should be `(props: ${iconName}Props)`

scripts/generateIcons.js
markdalgleish: Updated the code to make it clearer what's going on here.

scripts/generateIcons.js
```diff
@@ -0,0 +1,197 @@
...
+      inlineStyles: {
+        onlyMatchedOnce: false,
+      },
+    },
+    { convertStyleToAttrs: true },
+  ],
+});
+
+const svgrConfig = {
+  svgProps: { fill: 'currentColor', width: 16, height: 16 },
```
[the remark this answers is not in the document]
markdalgleish: Nice. Done 👍

PR #300, opened 2019-08-27
fix(Icon): Normalize icon SVGs

patrickliu-sa:
Normalizing existing SVG with shared format:

- Keep **xmlns** attribute as unified namespace declaration for backward compatibility
- Removed `width` and `height` attributes as size should be handled by treat
- Set `viewBox` as `"0 0 192 192"` as standard, equivalent to same ArtBoard size found in Adobe Illustrator
- Path data rounded up to 1 decimal

🤞 Prepare a role model for all new born SVGs


maraisr:
How did you get the namespace to retain with svgr, or was that a manual step? Can't see that as an option with the library? Also; the width and height props are set at https://github.com/seek-oss/braid-design-system/blob/master/scripts/generateIcons.js#L45 so should probably remove them also?

patrickliu-sa:
> How did you get the namespace to retain with svgr, or was that a manual step? Can't see that as an option with the library? Also; the width and height props are set at https://github.com/seek-oss/braid-design-system/blob/master/scripts/generateIcons.js#L45 so should probably remove them also?

Yes, I think we can remove that too since icon component is always fit into a fixed size container. We can do it in another PR

markdalgleish:
Closing this PR because we have new SVGs on the way.

PR #302, opened 2019-08-30
fix(Text): Ensure text context values are memoized

markdalgleish:
Because we're passing an object to the `TextContext` provider, it will fail a reference equality check against the last render, which will then trigger a re-render of all `TextContext` consumers. To fix this, we now ensure that the context value object is memoized across renders using [`useMemo`](https://reactjs.org/docs/hooks-reference.html#usememo).


maraisr:
Out of education - how did you guys notice this bug?

markdalgleish:
I was actually already aware that this is a potential issue with context values.

We recently added a new context to the system that used objects to provide multiple values and I made sure to memoize it there, but it made me realise that we'd missed it here when we migrated from a boolean context value to an object.

markdalgleish:
That said, the profiler in React's dev tools is really good at surfacing this sort of issue because it shows you which components are re-rendering and why.

PR #320, opened 2019-09-10
fix: Re-introduce BraidLoadableProvider

mattsjones:
BREAKING CHANGE: Introduced reset module that must be imported prior to components.

```diff
+import 'braid-design-system/reset';
import { BraidProvider } from 'braid-design-system'
```


README.md
```diff
@@ -20,21 +20,24 @@ In your sku project, first install this library:
...
-Then, import the desired theme at the _very top_ of your application (e.g. `src/client.js`), before importing anything else.
+At the _very top_ of your application, import the reset, required theme and the `BraidProvider` component.
 
-**WARNING: Since the theme contains styles that may be overridden by invididual components, the theme needs to be imported first to avoid CSS ordering issues.**
+**WARNING: The reset styles must be imported first to avoid CSS ordering issues.**
 
 For example:
 
 ```js
+import 'braid-design-system/lib/reset'; // <-- Must be first
```
markdalgleish: Similar to themes, could we make this `braid-design-system/reset`?

lib/reset/resetTracker.ts
```diff
@@ -0,0 +1,23 @@
+import dedent from 'dedent';
+
+let resetImported = false;
+
+export const markResetImported = () => {
+  resetImported = true;
+};
+
+export const ensureResetImported = () => {
```
markdalgleish: Should we put these function definitions behind a `process.env` check so the code isn't included in the production bundle?

package.json
```diff
@@ -4,7 +4,8 @@
   "description": "Themeable design system for the SEEK Group",
   "main": "lib/components/index.ts",
   "sideEffects": [
-    "lib/themes/**/*"
+    "lib/themes/**/*",
+    "lib/reset/**/*"
```
markdalgleish: Nice catch 👍

lib/playroom/components.ts
```diff
@@ -0,0 +1,2 @@
+import '../reset';
+export * from '../components';
```
markdalgleish: Nice catch here too 👍

lib/reset/resetTracker.ts
mattsjones: The other `process.env` checks allow for this entire file to be stripped in prod so no need.

markdalgleish: Awesome, even better 👍

lib/components/BraidLoadableProvider/BraidLoadableProvider.tsx
```diff
@@ -0,0 +1,29 @@
...
+  fallback?: ReactNode;
+}
+export const BraidLoadableProvider = ({
+  themeName,
+  children,
+  ...restProps
+}: BraidLoadableProviderProps) => (
+  <BraidLoadable themeName={themeName}>
+    {({ default: theme }: any) => (
+      <BraidProvider theme={theme} {...restProps}>
```
markdalgleish: If you put `{...restProps}` before the `theme` prop, it'll ensure that users can't override the theme.

PR #331, opened 2019-09-16
feat(BulletList): Add support for changing text size & bullet spacing

michaeltaranto:
BREAKING CHANGE: Setting `component="li"` on `Text` no longer renders a bullet. Please migrate to `Bullet`. Also no longer renders white space after the last `Bullet`, please validate your usage.

See usage: https://seek-oss.github.io/braid-design-system/components/BulletList.

### Migration Guide
Setting the `component` prop on `Text` to `"li"` no longer renders a bullet list item. Consumers depending on this behaviour should migrate to `Bullet` and leverage the new apis on `BulletList` to achieve the desired outcome.

#### Text size of bullets
Consumers using `Text` instead of `Bullet` to support customising the text size should now migrate to `Bullet` and set the `size` on `BulletList` as follows:
```diff
-<BulletList>
+<BulletList size="small">
-  <Text component="li" size="small">...</Text>
+  <Bullet>...</Bullet>
   ...
</BulletList>
```

#### Space between bullets
Consumers using `Text` to remove default spacing in between `Bullet`s should now migrate to `Bullet` and instead configure the `space` on `BulletList`. Eg:
```diff
-<BulletList>
+<BulletList space="none">
-  <Text component="li">...</Text>
+  <Bullet>...</Bullet>
   ...
</BulletList>
```

Alternatively you can also increase the spacing as required, using the space scale, eg:
```diff
-<BulletList>
+<BulletList space="large">
  <Bullet>...</Bullet>
   ...
</BulletList>
```

### DIFF
--- a/lib/components/BulletList/BulletList.tsx
+++ b/lib/components/BulletList/BulletList.tsx
@@ -1,10 +1,44 @@
-import React, { ReactNode } from 'react';
+import React, { ReactNode, createContext, Children } from 'react';
 import { Box } from '../Box/Box';
+import { UseTextProps } from '../../hooks/typography';
+import { SpaceY } from '../../hooks/useBox';
+
+const defaultSize = 'standard';
+const defaultSpace = 'xxsmall';
+interface BulletListContext {
+  size: UseTextProps['size'];
+  space?: SpaceY;
+}
+export const BulletListContext = createContext<BulletListContext>({
+  size: defaultSize,
+  space: defaultSpace,
+});
 
 export interface BulletListProps {
   children: ReactNode;
+  size?: UseTextProps['size'];
+  space?: SpaceY;
 }
 
-export const BulletList = ({ children }: BulletListProps) => (
-  <Box component="ul">{children}</Box>
-);
+export const BulletList = ({
+  children,
+  size = defaultSize,
+  space = defaultSpace,
+}: BulletListProps) => {
+  const bullets = Children.toArray(children);
+
+  return (
+    <Box component="ul">
+      {bullets.map((child, index) => (
+        <BulletListContext.Provider
+          value={{
+            size,
+            space: index !== bullets.length - 1 ? space : undefined,
+          }}
+        >
+          {child}
+        </BulletListContext.Provider>
+      ))}
+    </Box>
+  );
+};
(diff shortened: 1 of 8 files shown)


lib/components/Column/Column.docs.tsx
```diff
@@ -44,7 +44,7 @@ const docs: ComponentDocs = {
       Example: () => (
         <Fragment>
           {widths.map(width => (
-            <Columns>
+            <Columns key={width}>
```
mattsjones: 🙏 Thank you!

lib/components/Text/Text.tsx
```diff
@@ -53,9 +49,9 @@ export const Text = ({
     <TextContext.Provider value={textContextValue}>
       <Box
         id={id}
-        display={!isListItem ? 'block' : undefined}
+        display="block"
         component={component}
-        className={classnames(textStyles, isListItem && styles.listItem)}
+        className={classnames(textStyles)}
```
markdalgleish:
```suggestion
        className={textStyles}
```

lib/components/BulletList/BulletList.tsx
```diff
@@ -1,10 +1,44 @@
...
+import React, { ReactNode, createContext, Children } from 'react';
 import { Box } from '../Box/Box';
+import { UseTextProps } from '../../hooks/typography';
+import { SpaceY } from '../../hooks/useBox';
+
+const defaultSize = 'standard';
+const defaultSpace = 'xxsmall';
+interface BulletListContext {
+  size: UseTextProps['size'];
+  space?: SpaceY;
```
markdalgleish: Should this be a mandatory field? The context object should always have everything defined, right?

michaeltaranto: I could go either way, i refactored this to be optional so that the spacing logic lived in `BulletList` not `Bullet`. `Bullet` is then naive and renders `space` from context unconditionally. I could pass `none`, but micro-optimised to `undefined`.

lib/components/Text/Text.tsx
michaeltaranto: I'll do this manually and drop the import as well.

PR #341, opened 2019-09-24
chore(playroom): Fix Playroom code parsing

markdalgleish:


package.json
```diff
@@ -116,6 +116,9 @@
     "travis-deploy-once": "^5.0.9",
     "ts-node": "^8.2.0"
   },
+  "resolutions": {
+    "acorn": "6.3.0"
```
maraisr: Out of curiosity @markdalgleish - isn't the Babel parser like 100x better?

markdalgleish: This is for Playroom, primarily, which runs in the browser. We still use Babel.

PR #363, opened 2019-10-22
fix(Radio): Fix disabled styles for seekAsia & seekAsiaRebrand

patrickliu-sa:
Spotted ambiguous disabled radio states for several themes

### Before
image

### After
image

### DIFF
--- a/lib/themes/baseTokens/seekAsia.ts
+++ b/lib/themes/baseTokens/seekAsia.ts
@@ -19,11 +19,11 @@ export const makeTokens = ({
   const white = '#fff';
   const blue2 = '#298EB9';
   const blue3 = '#94C9E0';
-  const blue5 = '#EEF8FC';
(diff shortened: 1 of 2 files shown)


lib/themes/baseTokens/seekAsia.ts
```diff
@@ -201,11 +200,11 @@ export const makeTokens = ({
       background: {
         brand,
         input: white,
-        inputDisabled: grey4,
+        inputDisabled: grey5,
         brandAccent,
         formAccent,
         formAccentDisabled: grey4,
-        selection,
+        selection: grey5,
```
markdalgleish: This changes the highlight colour when selecting suggestions in `Autosuggest`. Can you double check that this is what you want?

patrickliu-sa: You're right. I'll revoke that colour change

PR #365, opened 2019-10-23
fix(seekAnz): Change selection colour from grey to blue

markdalgleish:
This is designed to enable much larger selection areas—not just small elements like suggestions in `Autosuggest`.

Before: 
Screen Shot 2019-10-23 at 11 40 00 am

After:
Screen Shot 2019-10-23 at 11 36 53 am


markdalgleish:
Updated PR description with screenshots.

PR #371, opened 2019-10-30
Add warning, notice & promote tone

patrickliu-sa:
Add `warning`, `notice` and `promote` tones to themes for brand refresh

image

Make `badge` as the first candidate of using these new tones.

Here's the mockup reference:
image


michaeltaranto:
Talking to the design team we are planning to adopt the colour usage matrix more iteratively, as they are not certain we need all permutations available. However the team is keen to add the `promote` tone so we will open a PR that adds this system wide in favour of just `Badge`.

PR #389, opened 2019-11-08
docs(site): Fix layout issues, fix typo

### LINKED ISSUE #385: Docs site windows/mac (with scroll bars) bug
Not sure if it's by design, but us Mac users that have scrollbars enabled, or Windows developers the lack of padding, makes it rather hard to read...

image

markdalgleish:
This PR adds space between the menu scrollbar and the page content. It also adds padding to the menu and page content areas to ensure that heading ascenders aren't visible when opening the menu on mobile.

Fixes #385.


site/src/App/Documentation/Documentation.tsx
```diff
@@ -49,6 +49,8 @@ const MenuSectionList = ({
   </Stack>
 );
 
+const responsiveGutter: ['gutter', 'large'] = ['gutter', 'large'];
```
maraisr: When a developer comes through and goes "I need to now add `small`", is this _type_ going to house any value? I mean, add `small` to the array and also the type, done. I guess what my question is, shouldn't this merely use type inference, and let the consumer "expect" a type. and that's where you get the value?

markdalgleish: From a type inference perspective, TypeScript only sees this as an array of strings, i.e. `string[]`. TypeScript has no idea that these strings are meaningful.

You can write `const responsiveGutter = ['gutter', 'large] as const` which tells TypeScript that the entire array is a constant. The only problem is that this also has the effect of marking the array as read only, which clashes with the types on Box: `The type 'readonly ["gutter", "large"]' is 'readonly' and cannot be assigned to the mutable type '[Space, Space]'.` This is something that we should probably fix on Box, but that's outside the scope of this work.

In terms of "value", ultimately the value here is that unsupported space values in this array will be raised at compile time rather than silently failing.

PR #391, opened 2019-11-08
feat(Icon): Add more components to suite

michaeltaranto:
Bulk-adding a wider selection of icons to the suite.

Screen Shot 2019-11-08 at 2 36 44 pm


icons/company.svg
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 23.0.2, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="0 0 24 24" style="enable-background:new 0 0 24 24;" xml:space="preserve">
+<title>SeekIconTemplate24px_03</title>
```
markdalgleish: We should probably filter out title tags.

icons/company.svg
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 23.0.2, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="0 0 24 24" style="enable-background:new 0 0 24 24;" xml:space="preserve">
```
markdalgleish: This style attribute seems redundant.

icons/company.svg
michaeltaranto: I dont think we should have an opinion about these files. They should be whatever the exported structure is from illustrator. I havent even looked at them (deliberately).

So long as our script handles it and has the desired output—that's what matters right?

PR #403, opened 2019-11-13
feat(Loader): Add initial component

michaeltaranto:


lib/components/Loader/Loader.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { Theme } from 'treat/theme';
+import { Box } from '../Box/Box';
+import { useBackground } from '../Box/BackgroundContext';
+import { UseBoxStylesProps } from '../Box/useBoxStyles';
+import * as styleRefs from './Loader.treat';
+
+const indicators = [...Array(3)];
+
+interface LoaderProps {
+  size?: keyof Theme['typography']['text'];
```
markdalgleish: I don't think we should be importing `Theme` directly from `treat`. Should we model this as `size?: keyof typeof styleRefs.size`?

lib/components/Loader/Loader.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+const indicatorColour = (
+  backgroundContext?: UseBoxStylesProps['background'],
+) => {
+  if (backgroundContext) {
+    const lookup: Partial<
```
markdalgleish: Should we hoist this out of the function so it's not re-created on every call?

lib/components/Loader/Loader.tsx
michaeltaranto: Yep now we can, that was a relic of a refactor. FWIW though i was followed this pattern from [`useBoxStyles`](https://github.com/seek-oss/braid-design-system/blob/25efb2c32cfae990b2775e7a1a1f7cc6b77f538a/lib/components/Box/useBoxStyles.ts#L12)

lib/components/Loader/Loader.tsx
michaeltaranto: Yep, it no longer depends on anything in the scope. Will do

lib/components/Loader/Loader.tsx
```diff
@@ -47,7 +35,11 @@ export const Loader = ({ size = 'standard' }: LoaderProps) => {
         <Box
           key={index}
           borderRadius="full"
-          background={indicatorColour(backgroundContext!)}
+          background={
+            backgroundContext
+              ? resolveBgForContext[backgroundContext]
```
markdalgleish: Should this fall back to `neutral` if no match is found?

michaeltaranto: Damn, i fixed that but it wasnt staged 🤦‍♂

PR #405, opened 2019-11-14
feat: Support for labelling Checkbox & Radio externally

michaeltaranto:
Adds support for consumers to use `Checkbox` or `Radio` in custom components/layouts while retaining screen reader accessibility by providing the `aria-labelledby` prop.


michaeltaranto:
Closing to explore more type-safe options and to explore ways of reducing the internal complexity of coupling these two components together.

PR #416, opened 2019-12-08
fix(Inline): Remove top margin when Inline is not first child

markdalgleish:
## Breaking Change

Even though this is a bug fix, it could cause unwanted side effects to your designs if you were relying on this incorrect behaviour. Check usages of `Inline` to ensure that the layout is still correct.

## Development Notes

I also used this as an opportunity to clean up the code a bit and give the `useNegativeOffsetX`/`useNegativeOffsetX` hooks more appropriate names (`useNegativeMarginLeft`/`useNegativeMarginTop`).


lib/hooks/useNegativeMargin/useNegativeMarginTop.treat.ts
```diff
@@ -1,40 +1,38 @@
...
-  selectors: {
-    '&:first-child:before': cssProperties,
-  },
-});
-
 export const preventCollapse = 1;
 export const base = style({
   paddingTop: preventCollapse,
-  ...beforeFirstChild({ content: '""', display: 'block' }),
+  ...{
```
mattsjones: ?

lib/components/Inline/Inline.tsx
```diff
@@ -19,15 +19,15 @@ export interface InlineProps {
...
 export const Inline = ({ space = 'none', align, children }: InlineProps) => {
   const styles = useStyles(styleRefs);
-  const negativeOffsetX = useNegativeOffsetX(space);
-  const negativeOffsetY = useNegativeOffsetY(space);
+  const negativeMarginLeft = useNegativeMarginLeft(space);
+  const negativeMarginTop = useNegativeMarginTop(space);
 
   return (
-    <Box className={classnames(negativeOffsetY)}>
+    <div className={classnames(negativeMarginTop)}>
```
mattsjones: Maybe keep as `Box`

PR #427, opened 2019-12-19
docs(playroom): Upgrade to latest version

michaeltaranto:
Upgrading to the latest playroom to take advantage of a heap of new features.

### New features
- Configurable themes & widths as user preference
- Configurable themes & widths in share url
- Add ability to dock editor to the right

### Performance
- Render only visible frames and lazy load others on scroll
- Improve response time when typing in the code editor

### Fixes
- Resolve code truncation issue when popping editor out

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ language: node_js
 cache: yarn
 notifications:
   email: false
-before_script:
-  - npm prune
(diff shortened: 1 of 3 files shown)


michaeltaranto:
Will upgrade via sku.

PR #437, opened 2019-12-30
fix(OverflowMenu): Allow conditional rendering in OverflowMenu children

georgespyropoulos:
The types for OverflowMenu don't currently allow for something like this:

```
<OverflowMenu>
  {something === true && <OverflowMenuItem>Item 1</OverflowMenuItem>}
  <OverflowMenuItem>Item 2</OverflowMenuItem>
</OverflowMenu>
```

I've extended the type to support returning booleans and nulls


markdalgleish:
Looks good. I've just updated `boolean` to `false` since we don't want `true` coming through. Also updated the API snapshot.

PR #440, opened 2019-12-31
feat(Stack): Allow Stacks to be semantic lists

markdalgleish:
This adds a `component` prop to `Stack` to customise the wrapper element, and if `ol` or `ul` are specified, each stack item is wrapped in an `li` element rather than a `div`.

For now, only `div`, `ol` and `ul` are supported. Any other values will throw an error in dev mode.

Because this is quite easy to break and can't be picked up in a visual test, I've added a few Jest tests to ensure the semantics are being managed correctly.

### DIFF
--- a/lib/components/Stack/Stack.test.tsx
+++ b/lib/components/Stack/Stack.test.tsx
@@ -1,0 +1,62 @@
+import '@testing-library/jest-dom/extend-expect';
+import React from 'react';
+import { render, cleanup } from '@testing-library/react';
+import { BraidProvider, Stack, Text } from '..';
+import { wireframe } from '../../themes';
+
+afterEach(cleanup);
+
+describe('Stack', () => {
+  it('should not render a list by default', () => {
+    const { queryAllByRole } = render(
+      <BraidProvider theme={wireframe}>
+        <Stack space="small">
+          <Text>1</Text>
+          <Text>2</Text>
+          <Text>3</Text>
+        </Stack>
+      </BraidProvider>,
+    );
+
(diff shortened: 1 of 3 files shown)


lib/components/Stack/Stack.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    const list = getByRole('list');
+    expect(list.nodeName).toBe('UL');
+
+    expect(
+      Array.from(list.childNodes).map(childNode => childNode.nodeName),
+    ).toEqual(['LI', 'LI', 'LI']);
+  });
+
+  it('should render a valid unordered list when "component" is "ol"', () => {
```
georgespyropoulos:
```suggestion
  it('should render a valid ordered list when "component" is "ol"', () => {
```

lib/components/Stack/Stack.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+      <BraidProvider theme={wireframe}>
+        <Stack space="small">
+          <Text>1</Text>
+          <Text>2</Text>
+          <Text>3</Text>
+        </Stack>
+      </BraidProvider>,
+    );
+
+    expect(queryAllByRole('list').length).toBe(0);
```
georgespyropoulos: Does `queryAllByRole` rely on there being an aria attribute on the element? Or will it return the element if it's simply a list?

markdalgleish: It supports both 😄

PR #442, opened 2020-01-06
chore: Only publish relevant files to npm

mattsjones:
Before:
image

After:
image


maraisr:
Out of curiosity; is there a reason why that couldn't have been articulated with the `files` property in the `package.json`? Ie, only send files you want, rather than specifically exclude files/folders? Just thinking for the chance that a new root folder gets added, that forgets to be added to the npmignore.

cc @mattsjones

mattsjones:
The `files` field would work as well. Just prefer a blacklist in this case though. I'd rather over-push to npm rather than forget to add a critical file/folder to the `files` whitelist.

.npmignore
```diff
@@ -1 +1,15 @@
-demo/
+.github/
+docs/
+generate-component-docs/
+report/
+scripts/
+site/
+dist-storybook/
+icons/
```
markdalgleish: Are you sure this won't exclude our icon components?

mattsjones: Yep, checked that.

PR #445, opened 2020-01-08
feat(OverflowMenuItem): Add data prop support

patrickliu-sa:
This add `data` prop support to `OverflowMenuItem` which is similar to `Button`. 

For instance `<OverflowMenuItem data={{ foo: "bar" }}>` 
will rendered as  `<OverflowMenuItem data-foo="bar">`


lib/components/OverflowMenuItem/OverflowMenuItem.docs.tsx
```diff
@@ -27,6 +27,28 @@ const docs: ComponentDocs = {
         </Box>
       ),
     },
+    {
```
mattsjones: I wouldn't add a new example for this as it doesn't test anything and isn't really adding anything to the docs.

PR #453, opened 2020-01-14
feat: Add useSpace and useColor hooks

mattsjones:
This PR also allows hooks to be documented on the website.


maraisr:
Curious; what is your `grid` token used for? I can see you times that with some space value quite alot?

mattsjones:
@maraisr `grid` is our base level unit for spacing things. All other (space-based) numbers in the system are multiples of it. It helps create a consistent feeling of layout spacing, a "vertical rhythm" so to speak.

maraisr:
@mattsjones that's cool! So is there a reason whys that's being calculated at will, rather than just have the 4px grid multiples just be tokens themselves? Or is that grid token being used in other scenarios also?

mattsjones:
To keep things "on-grid" sometimes you need the value directly. Perhaps you need to vertically centre something within a container. You might need something like: 
```
paddingBottom: grid / 2;
paddingTop: grid / 2;
```

PR #462, opened 2020-01-16
feat(Icons): Add crop size

mattsjones:
This PR also adds an extra check the the `generate-icons` script to increase performance. It will only update/create icons that have been updated since last time. You can force update all icons by running `yarn generate-icons --force`. 

Sorry about the PR size 😅


mattsjones:
Closing for now due to rethinking of problem.

PR #466, opened 2020-01-20
feat(Icons): Add cropX prop

mattsjones:


mattsjones:
Closing due to IE 11 constraints.

PR #472, opened 2020-01-29
fix(TextLink,TextLinkRenderer): Remove surrounding space, add hitArea prop

mattsjones:
## Breaking Change
- Block-level `TextLink`/`TestLinkRenderer` components no longer occupy full touchable height in the document flow. You will need to add a `hitArea="large"` prop and reintroduce white space via layout components.
- All `TextLink`/`TestLinkRenderer` components must now be nested within a `Text` component.

## Description

**When TextLink is mentioned below, it refers to both the `TextLink` and `TextLinkRenderer` components.**

Previously Braid had two different TextLink implementations, depending on whether it was nested within a `Text` component or not. The thinking was, TextLinks that are not part of a sentence should reserve surrounding white space to ensure that they satisfied a minimum touchable size. The documentation referred to these as "block links".

Unfortunately, this clashes with our layout philosophy of first-class white space via layout components. It was also a confusing API for consumers due to its implicit nature. 

This PR is removing the concept of block links entirely, in favour of an optional `hitArea` prop that can increase the hit area of the link without adding additional whitespace to the layout.

To help debug this in development mode, you can add a `data-braid-debug` attribute anywhere in the DOM tree to visualise the hit areas nested within it.

**Note: Inline links (i.e. links nested within a Text component) are completely unaffected by this change.**

## Migration Guide

```diff
-<TextLink href="...">
-  <Text>Block link</Text>
-</TextLink>

+<Text>
+  <TextLink hitArea="large" href="...">Block link</TextLink>
+</Text>
```

You may also need to reintroduce white space. This can be done via Braid's [layout components](https://seek-oss.github.io/braid-design-system/foundations/layout).

In simple cases where block links only contains 1–2 words, you can wrap them in a vertically centered touchable Box:

```tsx
<Box height="touchable" display="flex" alignItems="center">
  <Text>
    <TextLink hitArea="large" href="...">Block link</TextLink>
  </Text>
</Box>
```

As always, if you're finding this migration difficult, please reach out so we can give you a hand.

### DIFF
--- a/lib/hooks/typography/index.ts
+++ b/lib/hooks/typography/index.ts
@@ -29,21 +29,15 @@ export function useText({
   _LEGACY_SPACE_ = false,
 }: UseTextProps) {
   const styles = useStyles(styleRefs);
-  const inTextLinkRenderer = useContext(TextLinkRendererContext);
-  const space = useTouchableSpace(size);
   const textTone = useTextTone({ tone, backgroundContext });
 
   return classnames(
     styles.fontFamily,
     styles.text[size].base,
-    inTextLinkRenderer
-      ? space
-      : [
-          textTone,
-          styles.fontWeight[weight],
-          baseline ? styles.text[size].baseline : null,
-          baseline && !_LEGACY_SPACE_ ? styles.text[size].cropFirstLine : null,
-        ],
+    textTone,
+    styles.fontWeight[weight],
+    baseline ? styles.text[size].baseline : null,
+    baseline && !_LEGACY_SPACE_ ? styles.text[size].cropFirstLine : null,
   );
 }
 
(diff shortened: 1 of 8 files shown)


lib/hooks/typography/index.ts
```diff
@@ -29,21 +29,15 @@ export function useText({
...
-      : [
-          textTone,
-          styles.fontWeight[weight],
-          baseline ? styles.text[size].baseline : null,
-          baseline && !_LEGACY_SPACE_ ? styles.text[size].cropFirstLine : null,
-        ],
+    textTone,
+    styles.fontWeight[weight],
+    baseline ? styles.text[size].baseline : null,
+    baseline && !_LEGACY_SPACE_ ? styles.text[size].cropFirstLine : null,
```
michaeltaranto: 😍

PR #473, opened 2020-01-30
feat(Box): Add full classnames API to Box

mattsjones:


lib/components/Autosuggest/Autosuggest.tsx
```diff
@@ -572,7 +571,7 @@ export function Autosuggest<Value>({
                   width="full"
                   marginTop="xxsmall"
                   paddingY="xxsmall"
-                  className={classnames(styles.menu)}
+                  className={styles.menu}
```
markdalgleish: Love that we don't have to worry about this anymore 😍

lib/components/Box/useBoxStyles.ts
```diff
@@ -45,6 +45,7 @@ export interface UseBoxStylesProps {
   cursor?: keyof typeof styleRefs.cursor;
   pointerEvents?: keyof typeof styleRefs.pointerEvents;
   overflow?: keyof typeof styleRefs.overflow;
+  className?: Parameters<typeof classnames>[0];
```
markdalgleish: 👏

maraisr:
Quick one; sorry guys... Arent you worried about re-renders that could be caused by this - or are you no worse off than what it was?

PR #474, opened 2020-01-30
feat(Tag): Add Tag component

markdalgleish:
This PR adds the `Tag` component which supersedes [`Pill`](https://seek-oss.github.io/seek-style-guide/pill) from SEEK Style Guide.

One key difference from older implementations is that tags no longer expand to "touchable" height (i.e. at least 44px) when an `onClear` function has been provided. Instead, an absolutely positioned element is used to increase the hit area without impacting the layout. If you're looking to migrate from SEEK Style Guide, you may need to revisit your design to ensure that it still looks correct.

As a side bonus, in order to keep total CSS size to a minimum, this PR also adds `minWidth={0}` and `flexShrink={0}` to the Box API.

## Demo

```tsx
<Tag>Standard tag</Tag>
<Tag onClear={() => { ... }} clearLabel="Clear tag">Clearable tag</Tag>
```

Screen Shot 2020-01-30 at 3 00 17 pm

To see it in action for yourself, check out the [Playroom demo](https://braid-design-system--eb6a4298ca1b015433e9a750e2c828b8245edcc5.surge.sh/playroom/#?code=N4Igxg9gJgpiBcIA8BhAhgJygPgDoDsACQpAZQBc0wBrQgZwAcqYBeXEAcwFdzyYN2eIsRIAJGGigBLfB0IAbGADcY8tiAAsgimnxRMUQpQ50kAenGSZHISJIBJfPJkx6TMK3Z0AtmnnzBAjtiJAAVNBsAKTQlNFIwDCkGcnNwmyDgsIjsUVCAWQAZVOyMuyybFFJSYvThMrTsACUJMBSzBtKRcuwAOWgYADoAK1N2krqSM0dnfBhbLoARKSUpWAxCM3mQy2lZBWVVdS0QbBRFTDQAI0UjCNGd6y2HJxc3ZnUfPwCTzpC0wgg+DOEgwLGAAAoAJSEFjYQjAAC+COw0Vi8USyRqTz+EQBQPOoIh0Nh8KROXyRTGtWCJH+gOBmDBUJhcMRyMq1Sp2NpuPpBKZxNZZOaVDaHQmOLkfJBApZpORfVgw1G4rKUxesyeSCWKzWGy1Dz2ihUanYx2woQwXHwYDQfEMxnuEl21K601ejHeXl8-kCEp5rppoQAFq4AI5cKQ0QiXDAQADuRAAZhAAB6EIZcbwMOgAlTrcihhRoABeAE9CFAIBwBoQQ+HI9HYwnk2mM1mc78RBB80Yi-JSxWqzW60WI1HaM3E4QU+nM9ncz3+H3XAPy5XqwMu1jt3T8TKiXK2dziPXCOOm3Hp7P2wu88vC6vBxuR2eL5Or625x26F3iEuC37Z9h1rN9Gw-FsZzbecc3vQCn3XEDty5TpzHdTUMnMHQaCEcx0CwIQQARIA).

## Development Notes

In order to support clear icons with an absolute hit area, I had to refactor `IconButton` to leverage `useVirtualTouchable`, which forced me to revisit all usages throughout the app and reintroduce space where necessary.


lib/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { useStyles } from 'sku/treat';
+import { Box } from '../Box/Box';
+import { Text } from '../Text/Text';
+import { ClearButton } from '../iconButtons/ClearButton/ClearButton';
+import * as styleRefs from './Tag.treat';
+
+export interface TagProps {
+  onClear?: () => void;
+  clearLabel?: string;
+  children: ReactNode;
```
mattsjones: Should this just be `string`?

lib/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { ReactNode } from 'react';
+import { useStyles } from 'sku/treat';
+import { Box } from '../Box/Box';
+import { Text } from '../Text/Text';
+import { ClearButton } from '../iconButtons/ClearButton/ClearButton';
+import * as styleRefs from './Tag.treat';
+
+export interface TagProps {
+  onClear?: () => void;
+  clearLabel?: string;
```
mattsjones: Maybe we should enforce a clearLabel?

lib/components/Tag/Tag.docs.tsx
```diff
@@ -0,0 +1,43 @@
...
+        <Tag onClear={handler}>
+          Multiline tag
+          <br />
+          Multiline tag
+        </Tag>
+      ),
+    },
+    {
+      label: 'Test: Standard and clearable tags should be equal height',
+      docsSite: true,
```
mattsjones: `docsSite: false`?

markdalgleish: Oops, nice catch.

lib/components/iconButtons/IconButton.tsx
```diff
@@ -82,7 +83,7 @@ export const IconButton = forwardRef<HTMLButtonElement, IconButtonProps>(
         type="button"
         ref={forwardedRef}
         cursor="pointer"
-        className={styles.button}
+        className={classnames(styles.button, useVirtualTouchable())}
```
mattsjones: Don't need the `classnames` anymore

lib/components/private/touchable/useVirtualTouchable.treat.ts
```diff
@@ -12,8 +12,8 @@ export const virtualTouchable = style({
     minWidth: hitArea,
     height: '100%',
     top: '50%',
-    left: 0,
-    right: 0,
+    left: -10,
+    right: -10,
```
mattsjones: Any context here?

markdalgleish: This is so that the hit area is more-or-less centred over smaller elements (`IconButton` in this case), rather than hanging out over the right side of the element. The tradeoff is that it means the hit area extends horizontally outside the bounds of _all_ virtual touchable elements, but I actually think that's okay.

lib/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <Box
+        display="flex"
+        alignItems="center"
+        background="neutralLight"
+        paddingX="small"
+        paddingY={styles.constants.paddingY}
+        paddingRight={onClear ? 'xxsmall' : undefined}
+        className={styles.borderRadius}
+      >
+        <Text size={styles.constants.textSize} baseline={false}>
```
michaeltaranto: Should we `truncate` this content? Might need to have a play and see how it would handle long content

markdalgleish: IMO we shouldn't force that on consumers. They should be able to trust that we'll render what they give us.

lib/components/Tag/Tag.tsx
markdalgleish: Done. It's now required when `onClear` is provided.

lib/components/Tag/Tag.tsx
markdalgleish: Done.

lib/components/Tag/Tag.tsx
markdalgleish: Okay, you talked me into it. PR updated.

maraisr: What convinced you exactly @markdalgleish? Am curious.

markdalgleish: Given they're almost always going to be a set, having any of the tags span multiple lines will only serve to break the layout. To make this more palatable, I made sure to add a title tag with the full tag content.

PR #482, opened 2020-02-05
docs(playroom): Add snippets

michaeltaranto:


lib/components/FieldMessage/FieldMessage.tsx
```diff
@@ -35,7 +35,7 @@ export const FieldMessage = ({
     throw new Error(`Invalid tone: ${tone}`);
   }
 
-  if (!message && !reserveMessageSpace) {
+  if (!message && !secondaryMessage && !reserveMessageSpace) {
```
mattsjones: This will Ninja release next time. Maybe we should make this explicit as a fix?

lib/components/Stack/Stack.docs.tsx
```diff
@@ -77,6 +77,78 @@ const docs: ComponentDocs = {
       ),
     },
   ],
+  snippets: [
+    {
+      name: 'Xxsmall Space',
```
mattsjones: XXSmall?

lib/components/FieldMessage/FieldMessage.tsx
michaeltaranto: Yeah agree, i'll pull it out to another PR.

lib/components/Autosuggest/Autosuggest.docs.tsx
```diff
@@ -197,6 +195,127 @@ const docs: ComponentDocs = {
...
+          suggestions={[
+            { text: 'Apples' },
+            { text: 'Bananas' },
+            { text: 'Carrots' },
+          ]}
+        />
+      ),
+    },
+    {
+      name: 'Location search',
```
markdalgleish: Should this be something like "With icon" instead? Like what we've done with `TextField`?

lib/components/Tag/Tag.docs.tsx
```diff
@@ -46,6 +46,30 @@ const docs: ComponentDocs = {
       ),
     },
   ],
+  snippets: [
+    {
+      name: 'Standard',
+      code: <Tag>Tag</Tag>,
```
markdalgleish: Should these always be nested in an `Inline` like badges? Will they always be a set?

lib/components/Tiles/Tiles.docs.tsx
```diff
@@ -45,6 +45,58 @@ const docs: ComponentDocs = {
...
+          <Placeholder height={48} />
+          <Placeholder height={48} />
+          <Placeholder height={48} />
+          <Placeholder height={48} />
+          <Placeholder height={48} />
+        </Tiles>
+      ),
+    },
+    {
+      name: '3 columns, small space',
```
markdalgleish:
```suggestion
      name: '3 columns, medium space',
```

markdalgleish: Or, do we leave off the space description entirely?

PR #501, opened 2020-02-21
feat(icons): New icons plus some updated to existing

michaeltaranto:
Introducing some new icons to the suite as well as some edits. All the "updated icons" in the image below were increased in size within their viewport to look more balanced with the rest of the suite.

Screen Shot 2020-02-21 at 2 58 57 pm


lib/components/icons/IconVisibility/IconVisibility.docs.tsx
```diff
@@ -0,0 +1,18 @@
...
+const docs: ComponentDocs = {
+  category: 'Icon',
+  migrationGuide: true,
+  foundation: true,
+  screenshotWidths: [],
+  examples: [
+    {
+      label: 'Default',
+      Example: () => <IconVisibility />,
+    },
```
markdalgleish: Need an example for each state here.

PR #504, opened 2020-02-25
docs: Clean up docs

mattsjones:
- (Very basic) support for hooks in contract tests
- Create single point of truth for undocumented exports
- Handle multiple exports from single file
- Support descriptions in doc pages

### DIFF
--- a/sku.routes.js
+++ b/sku.routes.js
@@ -1,15 +1,18 @@
 const fs = require('fs');
 const path = require('path');
-
-const undocumentedExports = ['useThemeName', 'BoxRenderer'];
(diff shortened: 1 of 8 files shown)


sku.routes.js
```diff
@@ -1,15 +1,18 @@
 const fs = require('fs');
 const path = require('path');
-
-const undocumentedExports = ['useThemeName', 'BoxRenderer'];
```
markdalgleish: Oh nice—this made me realise that the `useThemeName` route is a 404 if you try to hit it directly. Great clean up.

PR #513, opened 2020-03-04
feat(Theme): Clone seekAnz theme to seekUnifiedBeta

mr-beerkiss:
BREAKING CHANGE:

- seekUnifiedBeta replaces the jobsDbReband theme.


lib/themes/seekUnifiedBeta/tokens.ts
```diff
@@ -0,0 +1,8 @@
+import { makeTokens } from '../baseTokens/seekUnifiedBeta';
```
michaeltaranto: Not sure if we need the base tokens abstraction for this theme. Do we foresee other themes extending/consuming this beta theme?

mr-beerkiss: I wasn't sure to be honest. Since this will eventually become the `seekUnified` theme I thought it made since to have it's own tokens, but maybe I'm misunderstanding how themes work

patrickliu-sa: It depends on how we're going to handle SEEK Biz theme in future: keep the current distinct theme, or extend from this new `seekUnifiedBeta` mimic current styles.

mr-beerkiss: `seekUnifiedBeta` is intended to be the theme for all APAC candidate teams

lib/themes/jobsDbRebrand/tokens.ts
```diff
@@ -1,5 +0,0 @@
-import { makeTokens } from '../baseTokens/seekAsiaRebrand';
```
michaeltaranto: The `seekAsiaRebrand` base tokens can probably be removed too

PR #520, opened 2020-03-10
fix(Reset): Hide search cancel button from input

ashin:
Before:

Screen Shot 2020-03-10 at 11 32 55 am

After:

Screen Shot 2020-03-10 at 11 33 11 am

### DIFF
--- a/.changeset/rotten-plants-search.md
+++ b/.changeset/rotten-plants-search.md
@@ -1,5 +1,5 @@
 ---
-"braid-design-system": patch
+'braid-design-system': patch
 ---
 
 Hide webkit native clear field on search type inputs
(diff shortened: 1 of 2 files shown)


lib/components/Autosuggest/Autosuggest.treat.ts
```diff
@@ -39,3 +39,11 @@ export const menu = [menuBase, menuHeight];
 export const groupHeading = style({
   textTransform: 'uppercase',
 });
+
+export const input = style({
+  selectors: {
+    [`&::-webkit-search-cancel-button`]: {
```
mattsjones: This should probably be part of the input reset. That way all fields will remove this behaviour.
https://github.com/seek-oss/braid-design-system/blob/a6ef88d129caa679ffce6719fbf944dfed2b46a5/lib/reset/reset.treat.ts#L67

ashin: Good suggestion. I've made that change.

PR #524, opened 2020-03-11
Improve changelog format

mattsjones:


.changeset/changelog.js
```diff
@@ -0,0 +1,58 @@
...
+      let { links } = await getInfo({
+        repo,
+        commit: changeset.commit,
+      });
+
+      const versionInfo = `${changeset.commit} ${
+        links.pull === null ? '' : links.pull
+      }`;
+
+      const summary = `- ${firstLine}: ${versionInfo}`;
```
michaeltaranto: I think now that were formatting this we can design it a bit more. I dont feel we need the commit if we have a link to the PR, so would it make more sense to do something like:
```ts
const summary = `- ${firstLine}${links.pull ? ` (${links.pull})` : ''}` 
```

PR #526, opened 2020-03-12
Only show focus rings on buttons for keyboard navigation

markdalgleish:
See changeset for details.


lib/components/private/hideFocusRings/HideFocusRingsRoot.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export const HideFocusRingsRoot = ({ children }: HideFocusRingsRootProps) => {
+  const [hideFocusRings, setHideFocusRings] = useState(false);
+
+  useEffect(() => {
+    const eventName = hideFocusRings ? 'keydown' : 'mousemove';
+    const toggleFocusRings = () => setHideFocusRings(x => !x);
+
+    window.addEventListener(eventName, toggleFocusRings);
```
mattsjones: 😍

markdalgleish:
Updated this so that we bail out of wiring up the focus ring logic when nested within a `BraidTestProvider`. Worth a re-review.

setupTests.ts
```diff
@@ -0,0 +1,13 @@
+import { format } from 'util';
+
+const error = global.console.error;
+
+global.console.error = (message: any, ...restArgs: any[]) => {
```
mattsjones: Nice

.changeset/stale-bags-eat.md
```diff
@@ -0,0 +1,15 @@
+---
+'braid-design-system': patch
```
mattsjones: Feel like `minor` would be more appropriate. But won't matter either way.

markdalgleish: Changed.

PR #528, opened 2020-03-12
chore(ci): Cache node_modules between runs

maraisr:
With this PR I aim to allow caching of node_modules to make builds run faster for you.

One addition could also be to cache the the yarn cache dir instead?


michaeltaranto:
Hey mate, thanks for sharing this technique. We might look at this in the future across all our defined workflows, but not wanting to jump the gun on optimisations.

PR #529, opened 2020-03-13
Add BraidTestProvider component

markdalgleish:
See changeset for details.

### DIFF
--- a/lib/components/BraidTestProvider/BraidTestProvider.docs.tsx
+++ b/lib/components/BraidTestProvider/BraidTestProvider.docs.tsx
@@ -1,13 +1,17 @@
 import React from 'react';
 import { seekAnz } from '../../themes';
 import { ComponentDocs } from '../../../site/src/types';
-import { Stack, Text } from '..';
+import { Stack, Text, Strong } from '..';
 
 const docs: ComponentDocs = {
   category: 'Logic',
   description: (
     <Stack space="large">
-      <Text>Alternative to `BraidProvider` for unit test environments.</Text>
+      <Text>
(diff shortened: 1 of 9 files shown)


lib/components/BraidTestProvider/BraidTestProvider.docs.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import { seekAnz } from '../../themes';
+import { ComponentDocs } from '../../../site/src/types';
+import { Stack, Text } from '..';
+
+const docs: ComponentDocs = {
+  category: 'Logic',
+  description: (
+    <Stack space="large">
+      <Text>Alternative to `BraidProvider` for unit test environments.</Text>
```
mattsjones: Could we add a strong warning to not import this in your actual App? Just to be sure.

markdalgleish: Maybe even throw an error?

mattsjones: How would we detect that?

markdalgleish: `NODE_ENV` of development or production?

mattsjones: Worried that some teams will have that set for some reason in there test setup.

markdalgleish: Yeah, fair point. I'll add it to the docs.

PR #530, opened 2020-03-13
Preparation for docs site uplift

michaeltaranto:


markdalgleish:
Just looked at the preview link on my phone, noticed a couple of issues:

- Scroll performance for the main content area is really bad for some reason—frame rate drops heavily.
- When the menu is open, I can scroll the page horizontally and the content goes over the top of the menu.

markdalgleish:
Also, we seem to have lost the scroll position management on route change.

PR #539, opened 2020-03-17
Docs: Fix components route not showing content on click

michaeltaranto:


michaeltaranto:
This was a test that has not seemed to resolve the issue. Tricky because can't produce with local build.

PR #540, opened 2020-03-18
Use seek-oss-ci user for changeset action commits

mattsjones:
See my [changesets fork](https://github.com/changesets/action/compare/master...mattsjones:configurable-user?expand=1) for details on the change.


maraisr:
Once again! Sorry guys... But curious; why did you want to use the seek oss accout, and not the github one? cc @mattsjones

PR #550, opened 2020-03-19
Themes: Introduce the Catho theme

michaeltaranto:
This is a preview deployment, we'll hold off merging this for now.

### Feature

Adds support to build product for the Catho market. This theme is an adaption of the [Quantum Design System](https://catho.github.io/quantum).


lib/themes/catho/tokens.ts
```diff
@@ -0,0 +1,217 @@
...
+      neutral,
+      neutralInverted: white,
+      formAccent,
+      critical,
+      info,
+      promote,
+      caution,
+      positive,
+      secondary,
+      secondaryInverted: 'hsla(0, 0%, 100%, 0.65)',
```
markdalgleish, line 194: Looks like IE11 doesn't support `hsla`: https://caniuse.com/#feat=mdn-css_types_color_hsl_function_accepts_alpha

michaeltaranto:
Yep, im updating this branch at the moment locally.

lib/themes/catho/tokens.ts
michaeltaranto: Looks like we use a similar technique across all themes. Might be something we should address holistically.

michaeltaranto: Hang on looks like we might be misreading, `hsl` does not accept an `alpha` value in IE11, but `hsla` does https://caniuse.com/#feat=mdn-css_types_color_alpha

markdalgleish: Ah yep, gotcha. Carry on.

PR #551, opened 2020-03-19
Migrate to import * as React from 'react'

maraisr:
Ref:
- facebook/react#18102
- https://twitter.com/sebmarkbage/status/1231673639080038400


maraisr:
After a conversation with @markdalgleish over twitter, this PR is deemed mute. And as such, will refrain from future PR's. Thank you again so much for the time, and guidance team.

PR #556, opened 2020-03-23
Box: Add userSelect="none"

patrickliu-sa:
Allow text within `<Box>` not being selected on _mouse:double-click_ or _mouse:hold-and-drag_


lib/components/Box/useBoxStyles.treat.ts
```diff
@@ -370,3 +370,9 @@ export const relativePosition = {
     mapToStyleProperty(relativePositionRules, 'right'),
   ) as PositionRulesType,
 };
+
+export const userSelect = styleMap({
+  auto: { userSelect: 'auto' },
+  none: { userSelect: 'none' },
+  text: { userSelect: 'text' },
```
markdalgleish: Why is `text` required?

lib/components/Box/useBoxStyles.treat.ts
```diff
@@ -370,3 +370,9 @@ export const relativePosition = {
     mapToStyleProperty(relativePositionRules, 'right'),
   ) as PositionRulesType,
 };
+
+export const userSelect = styleMap({
+  auto: { userSelect: 'auto' },
```
markdalgleish: Since this is the default value, this shouldn't be required, right?

lib/components/Box/useBoxStyles.treat.ts
patrickliu-sa: As `auto` can select sub-element if not constrained, it seems given more granular options would be better

markdalgleish: Can you give me an example of where you've used this?

markdalgleish: The hesitation here is that these rules are sent to _every_ user regardless of whether they are used or not, so we tend to only add commonly used values.

lib/components/Box/useBoxStyles.treat.ts
patrickliu-sa: Good catch, will remove it :)

lib/components/Box/useBoxStyles.treat.ts
patrickliu-sa: > Can you give me an example of where you've used this?

For instance, DB rebranded homepage has two accordions. Double clicked on its toggle might select all items in the target container occasionally.

markdalgleish:
We've opted to only add `userSelect="none"` for now since we typically to only provide values that Braid uses internally. I've also updated the PR so that Braid makes use of this new API rather than using custom styles.

PR #568, opened 2020-03-31
Docs: Update design and sticky header on mobile

michaeltaranto:


site/src/App/Navigation/Navigation.tsx
```diff
@@ -1,96 +1,162 @@
...
   children: ReactNode;
 }
+
 export const Navigation = ({ children }: NavigationProps) => {
   const styles = useStyles(styleRefs);
+  const { y: scrollTop } = useWindowScroll();
+  const [showStickyHeader, setShowStickyHeader] = useState(false);
   const [isMenuOpen, setMenuOpen] = useState(false);
 
+  const [lastScrollTop, setLastScrollTop] = useState(0);
```
mattsjones: It seems this value doesn't actually get used on render. Should it just be a ref instead?

site/src/App/Navigation/Navigation.tsx
```diff
@@ -1,96 +1,162 @@
...
+  useEffect(() => {
+    setDirection(null);
+  }, [location]);
+
+  useEffect(() => {
+    setShowStickyHeader(scrollTop > 300 && direction === 'up');
+  }, [direction, scrollTop]);
+
+  useEffect(() => {
+    document.body.classList[isMenuOpen ? 'add' : 'remove'](styles.scrollLock);
```
mattsjones: Could maybe refactor this to a hook e.g. `useScrollLock({ locked: isMenuOpen })`? 

Feel like it becomes more self documenting for future reference.

michaeltaranto: Love it!

PR #569, opened 2020-03-31
Divider: Add strong weight variant

markdalgleish:
See changeset for details.


lib/components/Divider/Divider.treat.ts
```diff
@@ -1,5 +1,8 @@
-import { style } from 'sku/treat';
+import { style, styleMap } from 'sku/treat';
 
-export const divider = style({
-  height: '1px',
-});
+export const base = style({ height: 1 });
```
michaeltaranto: I think we should base this height on the standard `borderWidth` from the theme.

PR #574, opened 2020-04-01
BraidProvider: Add 'linkComponent' prop to customise links

markdalgleish:
See changeset for details.


lib/components/BraidProvider/BraidProvider.tsx
```diff
@@ -1,5 +1,11 @@
-import React, { createContext, useContext, ReactNode } from 'react';
-import { TreatProvider } from 'sku/react-treat';
+import React, {
+  createContext,
+  useContext,
+  ReactNode,
+  ComponentType,
+  AnchorHTMLAttributes,
+} from 'react';
+import { TreatProvider } from 'sku/treat';
```
michaeltaranto, line 8: This looks like an accidental change, should be importing from `sku/react-treat`.

markdalgleish: 😫

lib/components/BraidProvider/BraidProvider.tsx
```diff
@@ -20,31 +26,47 @@ export const useBraidTheme = () => {
   return braidTheme;
 };
 
+export interface LinkComponentProps
+  extends AnchorHTMLAttributes<HTMLAnchorElement> {
+  href: string;
+}
+export type LinkComponent = ComponentType<LinkComponentProps>;
+const DefaultLinkComponent = (props: LinkComponentProps) => <a {...props} />;
```
michaeltaranto, line 34: Should the default be a reset anchor tag, eg. `<Box component="a" />`, thoughts? I cant decide.

lib/components/Link/Link.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export type LinkProps = LinkComponentProps;
+
+export const Link = ({ href, className, ...restProps }: LinkComponentProps) => {
+  const LinkComponent = useLinkComponent();
+
+  return (
+    <LinkComponent
+      href={href}
+      className={classnames(resetStyles.base, resetStyles.element.a, className)}
```
michaeltaranto, line 17: ah gotcha, just saw this.

any reason not to use `useBoxStyles({ component: 'a', className })`?

markdalgleish: Yeah, that probably makes more sense 👍

site/src/App/Code/Code.tsx
```diff
@@ -24,8 +24,8 @@ const formatSnippet = memoize(
   snippet =>
     prettier
       .format(snippet, {
-        parser: 'babel',
-        plugins: [babylonParser],
+        parser: 'typescript',
+        plugins: [typescriptParser],
```
michaeltaranto, line 28: 👏

lib/components/BraidProvider/BraidProvider.tsx
markdalgleish: This looks like a failed merge from master on my part.

PR #581, opened 2020-04-06
Add ButtonLink component

markdalgleish:
See changeset for details.

Also made the following changes:

- Refactored `Button` so it leverages `ButtonRenderer` to provide prop defaults.
- Improved link tests across the board.
- Ensured that all links in Playroom call `event.preventDefault()` in their default `onClick` implementations.
- Upgraded the `linkComponent` implementation on the docs site so that links with `href="#"` call `event.preventDefault()` to prevent navigation/scroll reset side effects.


lib/components/ButtonLink/ButtonLink.tsx
```diff
@@ -0,0 +1,34 @@
...
+export interface ButtonLinkProps
+  extends Omit<ButtonRendererProps, 'children'>,
+    Omit<LinkComponentProps, 'className' | 'style'> {
+  children?: ReactNode;
+}
+
+export const ButtonLink = ({
+  children,
+  weight = 'regular',
+  loading = false,
```
mattsjones, line 20: These defaults should be owned by the `ButtonRenderer`

markdalgleish: This was copied over from `Button`, so I'll fix it there too.

markdalgleish: Fixed.

PR #585, opened 2020-04-08
Remove lodash from MonthPicker and buildDataAttributes

mattsjones:
Other PR tweaks
- Add some tests for MonthPicker
- Remove Dropdown child validation

### DIFF
--- a/.changeset/pink-jokes-talk.md
+++ b/.changeset/pink-jokes-talk.md
@@ -2,10 +2,11 @@
 'braid-design-system': patch
 ---
 
-Drop lodash usage to decrease bundle size
+Drop lodash usage to decrease bundle size.
 
-Affected components:
+This directly affects `MonthPicker` and any components using the `data` prop:
 
 - All field components
-- All Menu components
+- OverflowMenu
+- MenuRenderer
 - Button
--- a/.changeset/pink-jokes-talk.md
+++ b/.changeset/pink-jokes-talk.md
@@ -1,0 +1,11 @@
+---
+'braid-design-system': patch
+---
+
+Drop lodash usage to decrease bundle size
+
+Affected components:
+
+- All field components
+- All Menu components
+- Button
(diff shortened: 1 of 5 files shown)


lib/components/private/buildDataAttributes.ts
```diff
@@ -1,6 +1,9 @@
-import mapKeys from 'lodash/mapKeys';
-
 export type DataAttributeMap = Record<string, string | number>;
 
-export default (data: DataAttributeMap = {}) =>
-  mapKeys(data, (_, attrName) => `data-${attrName}`);
+export default (
+  data: DataAttributeMap = {},
+): { [dataAttribute: string]: string } =>
+  Object.assign(
```
michaeltaranto, line 6: Any risk that consumers might not have polyfilled this? Should we add this in the release notes?

.changeset/pink-jokes-talk.md
```diff
@@ -0,0 +1,11 @@
+---
+'braid-design-system': patch
+---
+
+Drop lodash usage to decrease bundle size
+
+Affected components:
+
+- All field components
```
michaeltaranto, line 9: Maybe capture that this is only relating to use of the `data` prop.

lib/components/private/buildDataAttributes.ts
mattsjones: Good question. How about I switch to `object-assign` which react uses internally?

.changeset/pink-jokes-talk.md
```diff
@@ -0,0 +1,11 @@
...
+- All Menu components
```
michaeltaranto, line 10: For ease of scanning, i'd call out names here, `OverflowMenu` and `MenuRenderer`, and also maybe call out `MonthPicker` as a separate line item.

lib/components/private/buildDataAttributes.ts
michaeltaranto: Either that, or just spread into a new literal maybe? Given we are spreading below.

mattsjones: Changed to a `for of` implementation

PR #608, opened 2020-04-23
Box: Add additional styles to justifyContent prop

georgespyropoulos:


georgespyropoulos:
Not going ahead with this one, discussion here: https://seekchat.slack.com/archives/CD15GQHMW/p1587613561042400

maraisr:
I don't work at seek - mind enlightening me on that conversion here in the open? Keen to also see why this ideas was parked? :)

cc @georgespyropoulos @markdalgleish

markdalgleish:
The tradeoff with Box is that every style is shipped to every consumer regardless of whether it's used or not, so we try to keep the properties limited to the common use cases rather than trying to be a wholesale replacement for CSS. Within Braid, we typically only move things onto Box once it's shown up in a lot of treat files first.

While it's tempting to add things like this, we have to be a bit discerning so that we don't just keep adding new styles to Box indefinitely.

markdalgleish:
The other complication with this one is that [`space-evenly` is not supported in IE11.](https://caniuse.com/#search=space-evenly)

PR #621, opened 2020-05-06
Add zero opacity to Box

michaeltaranto:


lib/components/Toggle/Toggle.treat.ts
```diff
@@ -16,14 +16,10 @@ export const root = style({
 });
 
 const realFieldBase = style({
-  opacity: 0,
   height: hitArea,
   zIndex: 1,
   selectors: {
     ...debugTouchable(),
-    [`&:not(:disabled)`]: {
```
markdalgleish, line 24: Why was this removed?

lib/components/Toggle/Toggle.treat.ts
```diff
@@ -37,24 +33,15 @@ const realFieldPosition = style((theme) => {
 
 export const realField = [realFieldBase, realFieldPosition];
 
-export const label = [
-  style({
-    selectors: {
-      [`${realFieldBase}:not(:disabled) ~ &`]: {
-        cursor: 'pointer',
```
markdalgleish, line 44: Why was this removed? (Part 2)

lib/components/Toggle/Toggle.treat.ts
michaeltaranto: A toggle does not support being disabled. This was a copy and paste from `InlineField` originally i believe. So moved the `cursor: pointer` to box props (same for your next comment)

lib/components/Toggle/Toggle.tsx
```diff
@@ -48,6 +48,8 @@ export const Toggle = ({
         checked={on}
         onChange={handleChange(onChange)}
         position="absolute"
+        cursor="pointer"
```
markdalgleish, line 51: Ah, just saw this. Does this still apply when it's disabled?

michaeltaranto: You can't disable a toggle

markdalgleish: Didn't catch your earlier comment. Please ignore me.

lib/components/Toggle/Toggle.treat.ts
markdalgleish: Nice clean up 👍

.changeset/dry-forks-relate.md
```diff
@@ -0,0 +1,7 @@
+---
+'braid-design-system': minor
+---
+
+Add zero opacity to Box
+
+Provide zero opacity on `Box` as an optimisation.
```
markdalgleish, line 7: Probably worth adding an example usage here. Would also help make it clear to consumers that it's a number.

PR #624, opened 2020-05-07
Use 'assert' for runtime assertions during development

markdalgleish:
See changeset for details.

Note: We'll merge this once sku v10.3.0 has been out for a while, but I'm opening the PR now so the code doesn't rot in the meantime.


maraisr:
Curious; why didn't you use the `invariant`/`warning` methods typically used in the react space? The hyped `babel-plugin-dev-expression` already culls those things during builds, and optionally produce error codes - that map back to the English text.

markdalgleish:
Because it's decoupled from `NODE_ENV`. We want to be able to build Playroom in production mode but retain the assertions.

maraisr:
Sucks that second-grade is needed to be used; when tooling couple themselves to NODE_ENV. Imo that plugin should have a similar development mode that the react plugin has - so you can simply turn it of/on.

lib/components/MenuItem/MenuItem.tsx
```diff
@@ -39,14 +40,13 @@ export const MenuItem = ({ children, onClick, data }: MenuItemProps) => {
...
-    } else {
-      throw new Error('MenuItem rendered outside menu context');
-    }
+  assert(
+    menuContext !== null,
+    'A MenuItem must be rendered as an immediate child of a Menu. See the documentation for correct usage: https://seek-oss.github.io/braid-design-system/components/MenuRenderer',
+  );
+
+  if (menuContext === null) {
+    throw new Error();
```
michaeltaranto, line 49:
```suggestion
    throw new Error('MenuItem rendered outside menu context');
```

PR #632, opened 2020-05-12
Stack: Add support for Hidden stack items

markdalgleish:
See changeset for details.


lib/components/Stack/Stack.tsx
```diff
@@ -21,39 +26,49 @@ const alignToDisplay = {
...
 
-export const useStackItem = ({ align, component, space }: UseStackProps) => {
-  const styles = useStyles(styleRefs);
-
-  return useBoxStyles({
-    component,
-    className: styles.excludingLast,
-    paddingBottom: space,
+export const useStackItem = ({ align, space }: UseStackItemProps) =>
+  ({
```
markdalgleish, line 30: Modified this function to return Box props rather than a `className` string. This is so that I can dynamically change the responsive `display` value to `none` for hidden items. The alternative was to create a `StackItem` component, which I did begin to attempt, but that seemed a lot heavier than this approach.

lib/components/BulletList/BulletList.tsx
```diff
@@ -49,7 +50,9 @@ export const BulletList = ({
 
   return (
     <BulletListContext.Provider value={bulletListContextValue}>
-      <Box component="ul">{children}</Box>
+      <Box component="ul" className={useNegativeMarginTop(space)}>
```
michaeltaranto, line 53: Do you want to try replacing with a `Stack`?

lib/components/Bullet/Bullet.tsx
```diff
@@ -32,32 +32,32 @@ export const Bullet = ({ children }: BulletProps) => {
...
-        ]}
-      >
-        <Box display="flex">
-          <Box
-            display="flex"
-            alignItems="center"
-            className={useLineHeightContainer(size)}
-          >
+          })}
+        >
```
michaeltaranto, line 46: Now that we just have the text styles on it's own node, we should be able to use a `Text` component here

lib/components/Stack/Stack.tsx
```diff
@@ -21,39 +26,47 @@ const alignToDisplay = {
...
+    paddingTop: space,
     // If we're aligned left across all screen sizes,
     // there's actually no alignment work to do.
     ...(align === 'left'
-      ? {}
-      : {
-          display: mapResponsiveProp(align, alignToDisplay),
+      ? ({ display: 'block' } as const)
```
michaeltaranto, line 35: As discussed, let's default the display down in the call to `normaliseResponsiveProps`

lib/components/Stack/Stack.tsx
```diff
@@ -64,35 +77,96 @@ export const Stack = ({
...
+                  ] = normaliseResponsiveProp(stackItemProps.display);
+
+                  return {
+                    display: [
+                      hiddenOnMobile ? 'none' : displayMobile,
+                      hiddenOnTablet ? 'none' : displayTablet,
+                      hiddenOnDesktop ? 'none' : displayDesktop,
+                    ],
+                  } as const;
+                })()
```
michaeltaranto, line 146: Let's try and hoist this up out of the template.

PR #638, opened 2020-05-14
Improve field border contrast ratio

michaeltaranto:
Themes can now configure the colour of field borders separately to standard border colours. This facilitates using a higher contrast border colour on fields to further improve accessibility.

Updated themes include:

- seekAnz
- seekBusiness
- seekUnifiedBeta
- catho (based on referencing Quantum)

### DIFF
--- a/.changeset/brown-paws-brush.md
+++ b/.changeset/brown-paws-brush.md
@@ -1,0 +1,7 @@
+---
+'braid-design-system': patch
+---
+
+Toggle: Hide border on dark backgrounds
+
+Following the pattern of other bordered elements and hiding the border when on dark backgrounds. Reduces visual noise and cleans results in cleaner lines.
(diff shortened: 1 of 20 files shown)


.changeset/brown-paws-brush.md
```diff
@@ -0,0 +1,7 @@
+---
+'braid-design-system': patch
+---
+
+Toggle: Hide border on dark backgrounds
+
+Following the pattern of other bordered elements and hiding the border when on dark backgrounds. Reduces visual noise and cleans results in cleaner lines.
```
markdalgleish, line 7:
```suggestion
Following the pattern of other bordered elements, we now hide the border when on dark backgrounds. This reduces visual noise and results in cleaner lines.
```

markdalgleish:
Just thinking, it might be a good idea to make it *really* clear in the release notes that the UI is changing in a way that may require a design review.

PR #643, opened 2020-05-21
Add HiddenVisually, infer 'inline' prop for Hidden within text

markdalgleish:
See changesets for details.

This PR also starts the process of deprecating the `screen` boolean prop on `Hidden` since people *think* it's the same behaviour as `HiddenVisually`. In practice, `<Hidden screen>` hides the content for _everyone_, but makes it available on print, which is a total edge case that our system doesn't need to explicitly cater to.

### DIFF
--- a/lib/components/HiddenVisually/HiddenVisually.docs.tsx
+++ b/lib/components/HiddenVisually/HiddenVisually.docs.tsx
@@ -5,7 +5,6 @@ import { Text } from '../Text/Text';
 
 const docs: ComponentDocs = {
   category: 'Layout',
-  foundation: true,
   description: (
     <Text>
       Provides content to assistive technologies while hiding it from the
@@ -13,6 +12,7 @@ const docs: ComponentDocs = {
     </Text>
   ),
   screenshotWidths: [320],
+  screenshotOnlyInWireframe: true,
   examples: [
     {
       label: 'Inside Text',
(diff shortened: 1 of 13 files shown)


lib/components/HiddenVisually/HiddenVisually.docs.tsx
```diff
@@ -0,0 +1,28 @@
...
+  description: (
+    <Text>
+      Provides content to assistive technologies while hiding it from the
+      screen.
+    </Text>
+  ),
+  screenshotWidths: [320],
+  examples: [
+    {
+      label: 'Hidden below tablet',
```
mattcompiles, line 17: This label looks like a copy paste error?

lib/components/HiddenVisually/HiddenVisually.tsx
```diff
@@ -0,0 +1,29 @@
...
+interface HiddenVisuallyProps {
+  children: BoxProps['children'];
+}
+
+export const HiddenVisually = ({ children }: HiddenVisuallyProps) => {
+  const styles = useStyles(styleRefs);
+  const inText = Boolean(useContext(TextContext));
+  const inHeading = Boolean(useContext(HeadingContext));
+
+  const component = inText || inHeading ? 'span' : 'div';
```
mattcompiles, line 17: Does this matter seeing as it's `position: absolute`?

markdalgleish: This is more about maintaining a semantic document rather than the visual layout.

PR #656, opened 2020-05-28
Hide icons from screen readers that have no title

mattcompiles:


.changeset/silver-onions-decide.md
```diff
@@ -0,0 +1,7 @@
+---
+'braid-design-system': patch
+---
+
+Hide icons from screen readers that have no title
+
+Icons used purely as decoration should be hidden from screen readers.
```
michaeltaranto, line 7: I struggled on rewording this, but feel like we should describe both states rather than just the change.

```suggestion
Icons are mostly used for decorative purposes and as such are now hidden from screen readers unless providing a `title`. In this case, an Icon is identified as an image. 
```

PR #658, opened 2020-06-02
Upgrade @testiing-library dependencies

michaeltaranto:
Upgrading the @testing-library ecosystem to leverage new features in tests for an upcoming component, but also adding renovate schedule to keep them up to date.

Main code change is that clean up after each [should no be necessary unless using `pure`](https://github.com/testing-library/react-testing-library/releases/tag/v10.0.0).


lib/components/AccordionItem/AccordionItem.test.tsx
```diff
@@ -1,11 +1,9 @@
 import '@testing-library/jest-dom/extend-expect';
 import React, { useState } from 'react';
-import { render, cleanup } from '@testing-library/react';
+import { render } from '@testing-library/react';
 import { BraidTestProvider, AccordionItem } from '..';
 import { htmlToText } from '../../utils/htmlToText';
 
-afterEach(cleanup);
```
mattcompiles, line 7: 🎉

PR #662, opened 2020-06-23
useToast: Add key for toast deduplication

mattcompiles:


lib/components/useToast/useToast.test.tsx
```diff
@@ -155,4 +155,23 @@ describe('useToast', () => {
...
+  it('should deduplicate toasts with the same key', async () => {
+    const { showToast, queryAllToasts } = renderTestApp();
+
+    const key = 'deduped';
+
+    expect(queryAllToasts()).toHaveLength(0);
+
+    showToast({ tone: 'critical', message: 'First toast', key });
+    showToast({ tone: 'critical', message: 'Second toast', key });
+    showToast({ tone: 'critical', message: 'Third toast', key });
```
markdalgleish, line 168: Might be worth making this a bit more of a stress test. Maybe using multiple keys, and having an alternating mix of keyed and non-keyed toasts?

PR #670, opened 2020-07-01
Expose docs theme, improve site design

markdalgleish:


maraisr:
This is great Mark ✔ only 2 small-sih things for me though.

1. Would be great to see some sort of footer, or adding in some sort of gutter below. Its a bit hardpressed on the bottom there. It's more prominent on the examples, which has copy running on the bottom, rather than links.
2. Horizontal scroll bars that aren't hidden, look a bit funky.

image

PR #683, opened 2020-07-13
Add TextLinkButton component

markdalgleish:
See changeset for details.

### DIFF
--- a/generate-component-docs/__snapshots__/contract.test.ts.snap
+++ b/generate-component-docs/__snapshots__/contract.test.ts.snap
@@ -5991,6 +5991,22 @@ Object {
 }
 `;
 
+exports[`TextLinkButton 1`] = `
+Object {
+  exportType: component,
+  props: {
+    children: ReactNode
+    data?: Record<string, string | number>
+    hitArea?: 
+        | "large"
+        | "standard"
+    id?: string
+    onClick?: () => void
+    showVisited?: boolean
+},
+}
+`;
+
 exports[`TextLinkRenderer 1`] = `
 Object {
   exportType: component,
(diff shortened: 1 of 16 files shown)


generate-component-docs/__snapshots__/contract.test.ts.snap
```diff
@@ -5991,6 +5991,22 @@ Object {
 }
 `;
 
+exports[`TextLinkButton 1`] = `
+Object {
+  exportType: component,
+  props: {
+    children: ReactNode
+    data?: Record<string, string | number>
+    hitArea?:
```
michaeltaranto, line 6000: This prop is redundant on this "version" of a link right?

generate-component-docs/__snapshots__/contract.test.ts.snap
```diff
@@ -5991,6 +5991,22 @@ Object {
...
+  exportType: component,
+  props: {
+    children: ReactNode
+    data?: Record<string, string | number>
+    hitArea?: 
+        | "large"
+        | "standard"
+    id?: string
+    onClick?: () => void
+    showVisited?: boolean
```
michaeltaranto, line 6005: Does this still make sense?

markdalgleish: Good catch 👍

generate-component-docs/__snapshots__/contract.test.ts.snap
michaeltaranto: Ignore me, wrong mental model.

generate-component-docs/__snapshots__/contract.test.ts.snap
michaeltaranto: Again ignore this

markdalgleish: No, I think you're right on this. A button can't be marked as visited.

PR #696, opened 2020-07-27
Migrate to Capsize

michaeltaranto:
Improved the mechanism used to trim the white space around text. This will have subtle white space changes throughout the UI, but will largely just be improvements on balance of space around text.

As a result of adopting the new technique it is no longer feasible to maintain the `_LEGACY_SPACE_` fallback on `Heading` and `Text`.

#### Notes: Under the hood this uses our new open source library `capsize`. Check out [the site](https://seek-oss.github.io/capsize/) to have a play with its capabilities.


lib/hooks/typography/capsize.ts
```diff
@@ -0,0 +1,157 @@
...
+  | CapHeightWithLeading
+  | FontSizeWithLineGap
+  | FontSizeWithLeading;
+
+function capsize(options: CapHeightWithLeading): CapsizeStyles;
+function capsize(options: CapHeightWithLineGap): CapsizeStyles;
+function capsize(options: FontSizeWithLineGap): CapsizeStyles;
+function capsize(options: FontSizeWithLeading): CapsizeStyles;
+
+function capsize(options: CapsizeOptions) {
```
maraisr, line 64: point of curiosity @michaeltaranto—why hasn't braid consumed the npm package instead?

---
Please don't take this comment the wrong way, I love your work and this change 100%! Just genuinely curious.

michaeltaranto: Because it is not published yet. This PR will be updated when published.

maraisr: Ah yep! Thanks mate! Once again, brilliant effort! Love it.

.changeset/gorgeous-bears-glow.md
```diff
@@ -0,0 +1,63 @@
+---
+'braid-design-system': major
+---
+
+Improved trimming of white space around text
```
markdalgleish, line 5:
```suggestion
**`apac` and `seekBusiness` themes:** Improved trimming of white space around text
```

markdalgleish: Ignore me, I forgot that this also changes the height of text elements in the other themes.

.changeset/gorgeous-bears-glow.md
```diff
@@ -0,0 +1,57 @@
...
+
+
+## Breaking Changes
```
markdalgleish, line 8: I don't think we can use headings here since this ultimately gets nested inside a bullet list. This is why I usually go for the bold + all-caps style.

.changeset/gorgeous-bears-glow.md
```diff
@@ -0,0 +1,57 @@
...
+Improved the mechanism used to trim the white space around text. This will have subtle white space changes throughout the UI, but will largely just be improvements on balance of space around text.
```
markdalgleish, line 9: Should paragraph be before the "Breaking changes" heading?

.changeset/gorgeous-bears-glow.md
```diff
@@ -0,0 +1,57 @@
...
+
+As a result of adopting the new technique it is no longer feasible to maintain the `_LEGACY_SPACE_` fallback on `Heading` and `Text`.
```
markdalgleish, line 11: Should this section be a bullet list summary of the sections below?

lib/hooks/typography/basekick.ts
```diff
@@ -1,52 +0,0 @@
-interface BaseKickOptions {
```
markdalgleish, line 1: Let's pour one out for Basekick. Farewell, old friend. 🍻👋

michaeltaranto: 🌅 🥃 🥃 😢 👋

.changeset/gorgeous-bears-glow.md
```diff
@@ -0,0 +1,56 @@
...
+  }
+}
+```
+
+This is likely to not affect consumers, unless using these values off the theme explicitly in custom treat files.
+
+
+**Theme Tokens: Descender and Cap Height scales**
+
+Instead of the calculated values of `capHeightScale` and `decenderHeightScale`, a theme now accepts `fontMetrics`—a structure the represents the metadata from the font itself.
```
markdalgleish, line 45:
```suggestion
Instead of the calculated values of `capHeightScale` and `decenderHeightScale`, a theme now accepts `fontMetrics`—a structure that represents the metadata from the font itself.
```

.changeset/gorgeous-bears-glow.md
```diff
@@ -0,0 +1,57 @@
...
+-        size: 16,
++        capHeight: 11.43,
+        rows: 6
+      }
+    }
+  }
+}
+```
+
+This is not likely to affect consumers, unless using these theme values explicitly in custom treat files.
```
mattcompiles, line 42:
```suggestion
This is not likely to affect consumers, unless these theme values are used explicitly in custom treat files.
```

PR #700, opened 2020-07-29
Add useBreakpoint

mattcompiles:
See changeset for details


lib/components/useBreakpoint/BreakpointProvider.tsx
```diff
@@ -0,0 +1,68 @@
...
+  }
+  return 'mobile';
+};
+
+export const breakpointContext = createContext<Breakpoint | null>(null);
+
+interface BreakpointProviderProps {
+  children: ReactNode;
+}
+export function BreakpointProvider({ children }: BreakpointProviderProps) {
```
maraisr, line 29: Interesting that you've opted for a `Provider` approach. May I ask why this was done, instead of just letting the consumers of `useBreakpoint` not spin up its own listeners?

mattcompiles: It's a trade-off, the initial setup will always cause a re-render as we have to wait for the effect to know the viewport size (wouldn't be an issue if only client side rendering). It's quite easy to optimise the parent re-renders via memo but the benefit is the component is less likely to have to re-mount. You also won't have issues if you nest calls to `useBreakpoint`.

maraisr: Technically you don't have to wait in an effect though. You can `.matches` which gives you that information synchronously.

markdalgleish: Even if it's synchronous, we can't use a browser-only API to create the initial value because we have to support server rendering.

maraisr: We also have the SSR requirement, we solved this buy sending in a `fallbackCase` which gets used only in SS. This way you get optimal performance at runtime. Only 1 issue with this, hydration isnt guaranteed to be the same as server. But im working through it.

https://github.com/autoguru-au/overdrive/blob/main/packages/overdrive/lib/hooks/useMedia/useMedia.ts#L11

Cheers for the guidance here @markdalgleish ❤️  Always learn so much from you!

PR #705, opened 2020-07-31
Update changesets action

michaeltaranto:


.github/workflows/release.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         run: yarn --frozen-lockfile
 
       - name: Create Release Pull Request or Publish to npm
-        uses: changesets/action@65b9570
+        uses: changesets/action@bfeb9e0
```
maraisr, line 31: Curious; why not just `@master` and not worry about these PR's

PR #710, opened 2020-08-06
Add List component, mark BulletList/Bullet as deprecated

markdalgleish:
See changeset for details.

Also added "New" badges (for components added in the last 2 months) and "Deprecated" badges to the nav on the docs site.


lib/components/BulletList/BulletList.tsx
```diff
@@ -1,44 +1,90 @@
...
+      </List>
     </BulletListContext.Provider>
   );
 };
+
+export interface BulletProps {
+  children: ReactNode;
+}
+
+export const Bullet = ({ children }: BulletProps) => {
```
michaeltaranto, line 62: Seems to be a new file created `BulletList/Bullet.tsx` that is not needed since pulling `Bullet` into this file.

PR #721, opened 2020-08-12
Fix flaky Tabs screenshot test

mattcompiles:
Adds an optional `screenshotDelay` property to the docs. Starting with 300ms to see if that solves the Tabs animation issue.


mattcompiles:
Closing because I don't like storybook and it doesn't like me

PR #735, opened 2020-08-18
ContentBlock: Add support for xsmall & small widths

michaeltaranto:
**ContentBlock:** Add support for xsmall & small widths

**EXAMPLE USAGE**

```jsx
<ContentBlock width="small">
  ...
</ContentBlock>
```


lib/components/useToast/Toast.treat.ts
```diff
@@ -2,5 +2,4 @@ import { style } from 'sku/treat';
 
 export const toast = style({
   pointerEvents: 'all',
-  maxWidth: 400,
```
markdalgleish, line 5: 👏

PR #752, opened 2020-09-02
Add `aria-label` support to Loader

michaeltaranto:
**Loader:** Add support for `aria-label`

Provides a mechanism for consumers to better communicate to assistive technologies what is happening.

**EXAMPLE USAGE:**
```jsx
<Loader aria-label="Loading search results" />
```

**Developer Notes**
Also adding vscode snippet to support faster, more consistent release note authoring.

### DIFF
--- a/.changeset/sharp-pianos-share.md
+++ b/.changeset/sharp-pianos-share.md
@@ -1,0 +1,17 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated
+  - Loader
+---
+
+**Loader:** Add support for `aria-label`
+
+Provides a mechanism for consumers to better communicate to assistive technologies what is happening.
+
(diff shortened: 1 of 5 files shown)


.vscode/Changeset.code-snippets
```diff
@@ -0,0 +1,30 @@
+{
```
mattcompiles, line 1: 👏

.changeset/sharp-pianos-share.md
```diff
@@ -0,0 +1,17 @@
+---
+'braid-design-system': patch
```
mattcompiles, line 2: Should be `minor`

mattcompiles (approved):
Just one comment to address.

PR #753, opened 2020-09-10
Fix incorrect changeset frontmatter

mattcompiles:


.changeset/sharp-pianos-share.md
```diff
@@ -3,7 +3,7 @@
 ---
 
 ---
-updated
+updated:
```
michaeltaranto, line 6: Can you add this into the [vscode snippet too](https://github.com/seek-oss/braid-design-system/blob/master/.vscode/Changeset.code-snippets#L7). That's likely the cause

michaeltaranto (approved):

PR #754, opened 2020-09-10
Update Autosuggest to ARIA 1.2 combobox spec

mattcompiles:


lib/components/Autosuggest/createAccessbilityProps.ts
```diff
@@ -14,31 +14,28 @@ export const createAccessbilityProps = ({
...
-      autoComplete: 'off', // https://stackoverflow.com/questions/47775041/disable-autofill-in-chrome-63/47822599
-      spellCheck: false,
-      'aria-autocomplete': 'list',
+      role: 'combobox',
+      'aria-haspopup': 'listbox',
+      'aria-owns': isOpen ? menuId : undefined, // backwards compatibility for screenreaders implementing ARIA 1.0
       'aria-controls': menuId,
+      'aria-expanded': isOpen,
+      'aria-labelledby': labelId,
+      'aria-autocomplete': 'list',
```
michaeltaranto, line 28: Did we want to move to "both" here, as we support free text and suggestions.

michaeltaranto (commented):
I noticed we have no tests for the `label` association with the input field, should we add one to assert everything links together correctly?

michaeltaranto (approved):

PR #757, opened 2020-09-11
Autosuggest a11y uplift

mattcompiles:
Plus more, see changesets for details.


lib/components/Announcement/Announcement.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { createPortal } from 'react-dom';
+import { useStyles } from 'sku/react-treat';
+
+import { useBoxStyles } from '../Box/useBoxStyles';
+import * as styleRefs from '../HiddenVisually/HiddenVisually.treat';
+
+let announcementCounter = 0;
+
+interface AnnouncementProps {
+  children: string | undefined | null;
```
markdalgleish, line 11: Is it deliberate that this doesn't support multiple child nodes?

lib/components/Announcement/Announcement.docs.tsx
```diff
@@ -0,0 +1,103 @@
...
+        should be used for adding context to static markup. If you&apos;re
+        unsure which to use contact{' '}
+        <TextLink href="https://seekchat.slack.com/channels/braid-support">
+          #braid-support
+        </TextLink>
+        .
+      </Text>
+      <Heading level="3">Usage</Heading>
+      <Text>
+        Announcement must be rendered <Strong>before</Strong> sending it any
```
markdalgleish, line 47: Given this restriction, did you think about making it a hook?

mattcompiles: I did. I actually originally implemented it as a hook and then moved to a component. Happy to discuss further but I feel the component is slightly more elegant. I think it fits more naturally as something you render rather than having to worry about effects and cleanups.

lib/components/Announcement/Announcement.tsx
mattcompiles: That's correct. If you ever need multiple announcements in the same component (seems unlikely) you'd render two separate `Announcement` components.

markdalgleish: But I mean that it'll prevent interpolations like this: `<Announcement>{count} item{count === 1 ? '' : 's'}</Announcement>`

mattcompiles: Yeah good point. We can open it up.

.changeset/yellow-pets-clap.md
```diff
@@ -0,0 +1,15 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated:
+  - TextField
+  - PasswordField
+  - Textarea
+  - Dropdown
```
michaeltaranto, line 10: Have you looked into `MonthPicker` as well?

.changeset/curly-horses-build.md
```diff
@@ -0,0 +1,17 @@
+---
+'braid-design-system': minor
+---
+
+---
+updated:
+  - HiddenVisually 
+---
+
+**HiddenVisually:** Add support for passing ids
```
markdalgleish, line 10:
```suggestion
**HiddenVisually:** Add support for passing IDs
```

.changeset/rich-cameras-attack.md
```diff
@@ -0,0 +1,13 @@
...
+---
+
+---
+updated:
+  - Autosuggest 
+---
+
+**Autosuggest:** Improve screen reader experience
+
+Add description informing user that suggestions will appear below field. Also, notify user with how many suggestions are available and about automatic selections.
```
markdalgleish, line 12:
```suggestion
Add description informing user that suggestions will appear below field. Also, notify users about how many suggestions are available, and about automatic selections.
```

.changeset/yellow-pets-clap.md
```diff
@@ -0,0 +1,15 @@
...
+---
+
+**TextField,PasswordField,Textarea,Dropdown:** Add support for multiple field descriptions
+
+Previously, if a custom `aria-describedby` prop was passed it would take precedence over the `message` prop, which also uses `aria-describedby`. Both descriptions can now be applied at the same time.
```
markdalgleish, line 15:
```suggestion
Previously, if a custom `aria-describedby` prop was passed, it would take precedence over the `message` prop, which also uses `aria-describedby`. Both descriptions can now be applied at the same time. 
```

lib/components/Announcement/Announcement.tsx
```diff
@@ -0,0 +1,47 @@
+import { useEffect, useState } from 'react';
```
markdalgleish, line 1: Should this file be moved out of the top level components dir since it's not public anymore?

.changeset/yellow-pets-clap.md
```diff
@@ -0,0 +1,15 @@
...
+
+---
+updated:
+  - TextField
+  - PasswordField
+  - Textarea
+  - Dropdown
+---
+
+**TextField,PasswordField,Textarea,Dropdown:** Add support for multiple field descriptions
```
markdalgleish, line 13:
```suggestion
**TextField, PasswordField, Textarea, Dropdown:** Add support for multiple field descriptions
```

lib/components/Autosuggest/createAccessbilityProps.ts
```diff
@@ -12,6 +12,7 @@ export const createAccessbilityProps = ({
 }: AutosuggestProps) => {
   const menuId = `${id}-menu`;
   const labelId = `${id}-label`;
+  const assistiveDecriptionId = `${id}-aria-description`;
```
markdalgleish, line 15: Typo, should be `assistiveDescriptionId `

.changeset/yellow-pets-clap.md
```diff
@@ -0,0 +1,15 @@
+---
+'braid-design-system': patch
```
markdalgleish, line 2: Should this be `minor`?

markdalgleish: Ignore me, patch is right.

lib/components/Autosuggest/Autosuggest.tsx
```diff
@@ -23,6 +25,8 @@ import { smoothScroll } from '../private/smoothScroll';
 import { useScrollIntoView } from './useScrollIntoView';
 import { RemoveScroll } from 'react-remove-scroll';
 import { createAccessbilityProps, getItemId } from './createAccessbilityProps';
+import { autosuggest, AutosuggestTranslations } from '../../translations/en';
+
```
markdalgleish, line 29: Nice new line here. Very good.

markdalgleish (approved):
Nice one 👏

PR #763, opened 2020-09-15
Check for circular dependencies during build, fix existing issues

markdalgleish:
This introduces [circular-dependency-plugin](https://github.com/aackerman/circular-dependency-plugin) to catch circular dependencies in the webpack build.


mattcompiles:
https://bit.ly/2FkPvpV

mattcompiles (approved):

PR #784, opened 2020-10-02
Loader, Button: Improve performance of loading animations

michaeltaranto:
**Loader,Button:** Performance improvement for loading animations

Adjust animations properties and values to reduce CPU recalculation overheads.

### DIFF
--- a/.changeset/light-tables-repeat.md
+++ b/.changeset/light-tables-repeat.md
@@ -1,0 +1,13 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated:
+  - Loader
+  - Button
+---
+
+**Loader,Button:** Performance improvement for loading animations
+
+Adjust animations properties and values to reduce CPU recalculation overheads.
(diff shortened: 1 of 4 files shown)


.changeset/light-tables-repeat.md
```diff
@@ -0,0 +1,13 @@
...
+'braid-design-system': patch
+---
+
+---
+updated:
+  - Loader
+  - Button
+---
+
+**Loader,Button:** Performance improvement for loading animations
```
markdalgleish, line 11:
```suggestion
**Loader, Button, ButtonLink, ButtonRenderer:** Improve performance of loading animations
```

Need to add these to the frontmatter too.

maraisr:
@michaeltaranto out of curiosity how are you testing this? Like how are you knowing it was slow?

michaeltaranto:
Using the performance monitor in dev tools. The problem was made visible by a maxed out CPU on Gallery when sitting idle and we traced it back to this.

markdalgleish (approved):

PR #788, opened 2020-10-09
List: Add support for Roman numerals

markdalgleish:
See changeset for details.


michaeltaranto (approved):
Looks good to me, for bonus points you can throw in some more List snippets 😉 ✂️

PR #798, opened 2020-10-15
Form field description a11y & improved MonthPicker i18n support

michaeltaranto:
**TextField,Dropdown,PasswordField,MonthPicker,Textarea:** Apply aria-describedby to form elements only when needed

Only apply `aria-describedby` to form elements when needed, e.g. either a `message`, `description`, or an explicit `aria-describedby` is passed.

**MonthPicker:** Announce semantic grouping of fields and improved translation support.

When not on a native device, the MonthPicker uses a `fieldset` containing two dropdowns. This change ensures that the grouping is announced correctly. From a translations perspective the labels for the dropdowns are no longer a concatenation of the `label` and `monthLabel`/`yearLabel`, supporting translation of the entire phrase.

### DIFF
--- a/.changeset/fair-dragons-smoke.md
+++ b/.changeset/fair-dragons-smoke.md
@@ -1,0 +1,16 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated:
+  - TextField
+  - Dropdown
+  - PasswordField
+  - MonthPicker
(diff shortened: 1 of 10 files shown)


markdalgleish (approved):
Nice one 👍

PR #802, opened 2020-10-19
Apply aria-describedby only when needed

michaeltaranto:
**Radio,Checkbox:** Apply aria-describedby only when needed

Only apply aria-describedby when needed, e.g. either a message or description is passed.

### DIFF
--- a/lib/components/BulletList/BulletList.docs.tsx
+++ b/lib/components/BulletList/BulletList.docs.tsx
@@ -10,7 +10,7 @@ const docs: ComponentDocs = {
       <TextLink href="/components/List">List</TextLink> instead.
     </Text>
   ),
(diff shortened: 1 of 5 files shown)


lib/components/BulletList/BulletList.docs.tsx
```diff
@@ -10,7 +10,7 @@ const docs: ComponentDocs = {
       <TextLink href="/components/List">List</TextLink> instead.
     </Text>
   ),
-  migrationGuide: true,
+  migrationGuide: false,
```
michaeltaranto, line 13: Linked to a 404 page, so removing.

markdalgleish (approved):
Nice one 👍

PR #807, opened 2020-10-22
Add RadioGroup & RadioItem components

michaeltaranto:
**RadioGroup,RadioItem:** Add RadioGroup & RadioItem components

The RadioGroup provides an accessible way to group and control a set of **RadioItem** components. The RadioGroup is responsible for handling the value, tone, message, and disabled state—determining the presentation and selection of the items in the list.

**EXAMPLE USAGE:**
```jsx
<RadioGroup
  id="experience"
  label="Experience"
  value=""
  onChange={() => {}}
>
  <RadioItem label="Less than one year" value="0" />
  <RadioItem label="1 year" value="1" />
  <RadioItem label="2 years" value="2" />
  <RadioItem label="3+ years " value="3" />
</RadioGroup>
```


lib/components/Radio/Radio.tsx
```diff
@@ -1,25 +1,59 @@
...
-    reserveMessageSpace={false}
-    required={undefined}
-    ref={ref}
-  />
-));
+const NamedRadio = forwardRef<HTMLInputElement, RadioProps>((props, ref) => {
+  const radioListContext = useContext(RadioGroupContext);
+
+  if (process.env.NODE_ENV !== 'production') {
+    assert(
```
markdalgleish, line 20: This one should be outside of the env check so it runs in Playroom. Also, should be `radioGroupContext`.

lib/components/RadioGroup/RadioGroup.migration.md
```diff
@@ -0,0 +1,34 @@
+# RadioGroup Migration Guide
+
+A `RadioItem` must be wrapped inside of a `RadioGroup`. As part of the migration you should be hoisting the controlled state its handling to [RadioGroup.](https://seek-oss.github.io/braid-design-system/components/RadioGroup).
```
markdalgleish, line 3:
```suggestion
A `RadioItem` must be wrapped inside of a `RadioGroup`. As part of the migration, you should be hoisting the controlled state it's handling to [RadioGroup.](https://seek-oss.github.io/braid-design-system/components/RadioGroup).
```

.changeset/nasty-geckos-ring.md
```diff
@@ -0,0 +1,29 @@
...
+
+**RadioGroup,RadioItem:** Add RadioGroup & RadioItem components
+
+The RadioGroup provides an accessible way to group and control a set of **RadioItem** components. The RadioGroup is responsible for handling the value, tone, message, and disabled state—determining the presentation and selection of the items in the list.
+
+**EXAMPLE USAGE:**
+```jsx
+<RadioGroup
+  id="experience"
+  name="experience"
```
markdalgleish, line 19: Can drop this now.

lib/components/Radio/Radio.tsx
```diff
@@ -1,25 +1,59 @@
...
+    // eslint-disable-next-line no-console
+    console.warn(
+      dedent`
+        The "Radio" component has been deprecated and will be removed in a future version. Use "RadioGroup" with "RadioItem" instead.
+        %c
+        - <Radio name="count" label="One" value="1" onChange={...} />
+        - <Radio name="count" label="Two" value="2" onChange={...} checked />
+        - <Radio name="count" label="Three" value="3" onChange={...}  />
+        %c
+        + <RadioGroup name="count" value="2" onChange={...}>
```
markdalgleish, line 34: Should `name` be `id` here now?

lib/components/Radio/Radio.tsx
```diff
@@ -1,25 +1,59 @@
...
-    message={null}
-    reserveMessageSpace={false}
-    required={undefined}
-    ref={ref}
-  />
-));
+const NamedRadio = forwardRef<HTMLInputElement, RadioProps>((props, ref) => {
+  const radioListContext = useContext(RadioGroupContext);
+
+  assert(
```
markdalgleish, line 19: This should be a `console.warn` (when not in prod) rather than an `assert`, otherwise it's a breaking change.

michaeltaranto: It shouldnt be a breaking change, because existing `Radio` components cannot be inside a `RadioGroupContext` today.

markdalgleish: Ignore me, just realised this only throws when used inside `RadioGroup`.

markdalgleish: 😄

lib/components/Radio/Radio.tsx
```diff
@@ -1,25 +1,59 @@
...
-    {...props}
-    type="radio"
-    message={null}
-    reserveMessageSpace={false}
-    required={undefined}
-    ref={ref}
-  />
-));
+const NamedRadio = forwardRef<HTMLInputElement, RadioProps>((props, ref) => {
+  const radioListContext = useContext(RadioGroupContext);
```
markdalgleish, line 17: Could we rename this to `radioGroupContext`?

michaeltaranto: Yep, will update the reference in `RadioItem` too.

lib/components/Radio/Radio.docs.tsx
```diff
@@ -1,11 +1,18 @@
...
 
 const docs: ComponentDocs = {
   category: 'Content',
-  migrationGuide: true,
+  migrationGuide: false,
+  deprecationWarning: (
+    <Text weight="medium">
+      This component has been deprecated. Use{' '}
+      <TextLink href="/components/RadioGroup">RadioGroup</TextLink> and{' '}
+      <TextLink href="/components/RadioItem">RadioItem</TextLink> instead.
```
markdalgleish, line 13: This component doesn't have its own page.

markdalgleish (approved):
Love it 👏

PR #810, opened 2020-10-27
List: Add support for icons

michaeltaranto:
**List:** Add support for icons

Provides a way to use an icon for all the items in a list. When using `type="icon"` you must also provide the `icon` prop. See example below:

**EXAMPLE USAGE:**
```jsx
<List type="icon" icon={<IconTick tone="positive" />}>
  <Text>This is a list item.</Text>
  <Text>This is a list item.</Text>
  <Text>This is a list item.</Text>
</List>
```

### DIFF
--- a/.changeset/silly-lies-add.md
+++ b/.changeset/silly-lies-add.md
@@ -1,0 +1,22 @@
+---
+'braid-design-system': minor
+---
+
+---
+updated:
+  - List
+---
+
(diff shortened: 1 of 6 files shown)


markdalgleish (approved):
Awesome stuff 👏 

Happy for you to make a judgement call about the type snapshot / props docs.

PR #819, opened 2020-11-05
Autosuggest: Forward ref prop to input element

markdalgleish:


lib/components/Autosuggest/Autosuggest.tsx
```diff
@@ -700,4 +711,9 @@ export function Autosuggest<Value>({
       </Box>
     </Fragment>
   );
-}
+}) as <Value>(
```
mattcompiles, line 714: Just confirming this generic still works as intended from a consumer perspective?

mattcompiles (approved):
Nice one 👏

lib/components/Autosuggest/Autosuggest.tsx
markdalgleish: Yep, just double checked.

PR #824, opened 2020-11-09
Playroom: Add 'setDefaultState' function

markdalgleish:


markdalgleish:
Note, I also tweaked the code formatting on the docs site to unwrap fragments so that code snippets can contain multiple sibling nodes at the root, which Playroom supports.

markdalgleish:
Changed my mind on the API, went back from `defaultState()` to `setDefaultState()`. The code snippets in the prototyping guide seem clearer to me with the latter.

mattcompiles (approved):

PR #834, opened 2020-11-19
Split out Gallery examples from Docs

michaeltaranto:
In preparation for a documentation uplift we are changing how we handle component examples on the site. Instead of sharing one set of examples and opting in or out of gallery, docs site, storybook etc we are moving towards custom file extensions for each use case.

This PR breaks out the examples to be used in Gallery into their own custom file extensions `*.gallery.tsx`.


markdalgleish (approved):
Nice one 👏

PR #836, opened 2020-11-19
Add snippets tab to component docs

michaeltaranto:
Make snippets a first class part of the docs site by giving them their own tab. We will be reviewing and updating the snippets as a replacement to the examples currently on the Details tab.


lib/components/Dropdown/Dropdown.playroom.tsx
```diff
@@ -25,7 +25,6 @@ export const Dropdown = ({
   return (
     <BraidDropdown
       id={id ?? fallbackId}
-      name={name}
```
michaeltaranto, line 28: This was breaking the build, not sure how it hasn't been a problem earlier. Same as `PasswordField` change below

markdalgleish (approved):
Awesome 👏

PR #837, opened 2020-11-22
Split out screenshot examples from Docs

michaeltaranto:
The last PR In preparation for a documentation uplift we are changing how we handle component examples on the site. Instead of sharing one set of examples and opting in or out of gallery, docs site, storybook etc we are moving towards custom file extensions for each use case.

This PR breaks out the examples to be used in storybook/chromatic into their own custom file extensions *.screenshots.tsx.

Sorry about the size of this, but success for this PR is quite simple — an unchanged docs site and no diff in chromatic.


markdalgleish (approved):
😮

PR #843, opened 2020-11-25
Checkbox: Support inferring of tri-state checked value

michaeltaranto:
**Checkbox:** Support inferring of tri-state checked value

To simplify the use of tri-state checkboxes, the **checked** prop now supports resolving the tri-state value from a array of checked values.

**EXAMPLE USAGE:**
```jsx
 <Checkbox
  label="Select all"
  checked={[ true, false, false ]} // Will resolve to "mixed"
/>
```

### DIFF
--- a/.changeset/nervous-papayas-warn.md
+++ b/.changeset/nervous-papayas-warn.md
@@ -9,7 +9,7 @@ updated:
 
 **Checkbox:** Support inferring of tri-state checked value
 
-To simplify the use of tri-state checkboxes, the **checked** prop now supports resolving the tri-state value from a array of checked values.
+To simplify the use of tri-state checkboxes, the **checked** prop now supports resolving the tri-state value from an array of checked values.
 
 **EXAMPLE USAGE:**
 ```jsx
--- a/.changeset/nervous-papayas-warn.md
+++ b/.changeset/nervous-papayas-warn.md
@@ -13,9 +13,8 @@ To simplify the use of tri-state checkboxes, the **checked** prop now supports r
 
 **EXAMPLE USAGE:**
 ```jsx
-// Will resolve to "mixed" based on the array of values provided
  <Checkbox
   label="Select all"
-  checked={[ true, false, false ]}
+  checked={[ true, false, false ]} // Will resolve to "mixed"
 />
 ```
--- a/.changeset/nervous-papayas-warn.md
+++ b/.changeset/nervous-papayas-warn.md
@@ -1,0 +1,25 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated:
+  - Checkbox
+---
+
+**Checkbox:** Support inferring of tri-state checked value
+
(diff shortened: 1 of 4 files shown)


lib/components/Checkbox/Checkbox.test.tsx
```diff
@@ -195,3 +196,25 @@ describe('Checkbox', () => {
     expect(checkbox.checked).toBe(false);
   });
 });
+
+describe('resolveCheckedGroup', () => {
```
markdalgleish, line 200: Feel like we should be testing via the public API rather than an internal util. What do you think?

michaeltaranto: Yeah i think you're right. Legacy of previous iterations 😄

.changeset/nervous-papayas-warn.md
```diff
@@ -0,0 +1,21 @@
...
+  - Checkbox
+---
+
+**Checkbox:** Support inferring of tri-state checked value
+
+To simplify the use of tri-state checkboxes, the **checked** prop now supports resolving the tri-state value from a array of checked values.
+
+**EXAMPLE USAGE:**
+```jsx
+// Will resolve to "mixed" based on the array of values provided
```
markdalgleish, line 16: Nit: Move this comment down, on the same line as the checked prop?

lib/components/Checkbox/Checkbox.tsx
```diff
@@ -5,12 +5,42 @@ import {
...
+    if (acc === true) {
+      return val || 'mixed';
+    }
+
+    if (acc === false && index > 0) {
+      return val ? 'mixed' : false;
+    }
+
+    return val;
+  }, false);
```
BurntCaramel, line 23: Sorry to jump in, but it did make me think that this could be shortened to something like:

```suggestion
  values.some(val => val !== values[0]) ? 'mixed' : a[0] ?? false;
```

michaeltaranto: Sometimes you just stare at the same code for too long 😄 .

Thanks mate, i like the refactor 👏

.changeset/nervous-papayas-warn.md
```diff
@@ -0,0 +1,20 @@
...
+---
+
+---
+updated:
+  - Checkbox
+---
+
+**Checkbox:** Support inferring of tri-state checked value
+
+To simplify the use of tri-state checkboxes, the **checked** prop now supports resolving the tri-state value from a array of checked values.
```
markdalgleish, line 12:
```suggestion
To simplify the use of tri-state checkboxes, the **checked** prop now supports resolving the tri-state value from an array of checked values.
```

markdalgleish:
Might be worth adding a test for an empty array, just so we can't accidentally change the behaviour in the future?

markdalgleish (approved):
Just one minor typo fix and a question about adding another test, but happy to leave it with you from here. This is awesome 🙌

PR #845, opened 2020-11-27
PasswordField: Ensure disabled is handled correctly

michaeltaranto:
**PasswordField:** Ensure disabled is handled correctly

Fixes a bug where the **disabled** prop was not being honoured, it was hiding the visibility toggle but leaving the field enabled.

### DIFF
--- a/.changeset/weak-paws-help.md
+++ b/.changeset/weak-paws-help.md
@@ -1,0 +1,12 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated:
+  - PasswordField
+---
+
+**PasswordField:** Ensure disabled is handled correctly
+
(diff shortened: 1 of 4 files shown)


.changeset/weak-paws-help.md
```diff
@@ -0,0 +1,12 @@
...
+---
+
+---
+updated:
+  - PasswordField
+---
+
+**PasswordField:** Ensure disabled is handled correctly
+
+Fixes a bug where the **disabled** prop was not being honoured, it was hiding the visibility toggle but leaving the field enabled.
```
markdalgleish, line 12:
```suggestion
Fixes a bug where the **disabled** prop was hiding the visibility toggle but leaving the field enabled.
```

markdalgleish (approved):

jahredhope (approved):

PR #846, opened 2020-11-29
Dropdown: Only show blank option when value prop is blank

markdalgleish:


mattcompiles (approved):

.changeset/rotten-jars-rush.md
```diff
@@ -0,0 +1,5 @@
+---
+'braid-design-system': patch
+---
+
+**Dropdown:** Only show a blank option in the list when the `value` prop is blank and a placeholder isn't present
```
mattcompiles, line 5: Add `updated: Dropdown`?

markdalgleish: Ah, I always forget this.

PR #879, opened 2021-01-25
Button, Actions: Add 'size' prop, support 'small' size

markdalgleish:
Also reduced the horizontal padding of standard buttons from `gutter` to `medium`.

See changesets for details.


.changeset/dry-cameras-wave.md
```diff
@@ -0,0 +1,33 @@
...
+
+---
+updated:
+  - Button
+  - ButtonLink
+  - ButtonRenderer
+  - Actions
+---
+
+**Button, ButtonLink, ButtonRenderer, Actions:** Add `suze` prop, support `small` size
```
michaeltaranto, line 13:
```suggestion
**Button, ButtonLink, ButtonRenderer, Actions:** Add `size` prop, support `small` size
```

lib/components/Button/Button.docs.tsx
```diff
@@ -81,9 +175,10 @@ const docs: ComponentDocs = {
           </Text>
         </>
       ),
+      background: 'brand',
       Example: () =>
         source(
-          <Box background="brand" padding="large" borderRadius="standard">
+          <Box background="brand">
```
michaeltaranto, line 181: This `Box` can be removed now, think that was the plan behind the change anyway.

markdalgleish: I left this in for Playroom, but I used the `background` property on the example object so that the gutter matches the other examples. Might be a good idea to automatically add this Box to Playroom links when the example has a background?

michaeltaranto: Ah gotcha. Yeah that may be a nice update. But dont stress about coupling the two

michaeltaranto (approved):

PR #882, opened 2021-01-27
Update react-element-to-jsx-string

michaeltaranto:
Upgrading `react-element-to-jsx-string` to fix the bug where opening Playroom links from the docs site may not preserve white space as expected.

[This bug was fixed](https://github.com/algolia/react-element-to-jsx-string/pull/572) which means white space is preserved as expected when opening the Job Summary for example.

Clicking on the **Open in Playroom** button:
Before: [Link](http://braid-design-system--ae4ac7bafff1629a4c448d2627101a6dbbcb9c0f.surge.sh/examples/job-summary)
After: [Link](http://braid-design-system--496244d152fdc65a1320a587c71447b78d811963.surge.sh/examples/job-summary)

Previously collapses the white space between the classification icon and text.


mattcompiles (approved):
They added types as well?

michaeltaranto:
Added types in [2019](https://github.com/algolia/react-element-to-jsx-string/issues/475) and fixed some bugs in [Jan 2020](https://github.com/algolia/react-element-to-jsx-string/pull/516/files) we just never updated.

PR #893, opened 2021-02-10
Add variant to Button and deprecate weight

michaeltaranto:
Introduces a new `variant` prop to `Button`/`ButtonLink` giving consumers a single prop to use for selecting the visual style of the button. Choose from `solid` (default), `ghost`, `soft` or `transparent`. The colour of the button is now consistently controlled via the `tone` prop, with supported values being `brandAccent`, `critical` or `undefined`.

As a result the `weight` prop is now deprecated. See the migration guide below.

**EXAMPLE USAGE:**
```jsx
<Inline space="small" collapseBelow="desktop">
  <Button>Solid</Button>
  <Button variant="ghost">Ghost</Button>
  <Button variant="soft">Soft</Button>
  <Button variant="transparent">Transparent</Button>
</Inline>
```

**MIGRATION GUIDE:**
The `weight` prop is now deprecated. If you are not specifying a `weight` there is no change required.

If you are, each weight can be migrated as follows:

#### Regular
Can be replicated with a `variant` of `solid` (default).
```diff
-<Button weight="regular">...</Button>
+<Button variant="solid">...</Button>
```

Given it is the default `variant`, you could also choose to simply remove the `weight` prop.
```diff
-<Button weight="regular">...</Button>
+<Button>...</Button>
```

#### Strong
Can be replicated with a `variant` of `solid` (default), with a `tone` of `brandAccent`.
```diff
-<Button weight="strong">...</Button>
+<Button tone="brandAccent">...</Button>
```

#### Weak
Can be replicated with a `variant` of `ghost`.
```diff
-<Button weight="weak">...</Button>
+<Button variant="ghost">...</Button>
```

### DIFF
--- a/.changeset/light-oranges-call.md
+++ b/.changeset/light-oranges-call.md
@@ -1,0 +1,90 @@
+---
+'braid-design-system': minor
+---
+
+---
+updated:
+  - Text
+  - IconAdd
+  - IconBookmark
+  - IconCaution
+  - IconChevron
+  - IconClear
+  - IconCompany
+  - IconCompose
+  - IconCopy
+  - IconCreditCard
+  - IconCritical
+  - IconDate
+  - IconDelete
+  - IconDesktop
+  - IconDocument
+  - IconDocumentBroken
+  - IconDownload
+  - IconEdit
+  - IconEducation
+  - IconFilter
+  - IconGrid
+  - IconHeart
+  - IconHelp
+  - IconHistory
+  - IconHome
+  - IconImage
+  - IconInfo
+  - IconInvoice
+  - IconLanguage
+  - IconLink
+  - IconLinkBroken
+  - IconList
+  - IconLocation
+  - IconMail
+  - IconMinus
+  - IconMobile
+  - IconMoney
+  - IconNewWindow
+  - IconNote
+  - IconNotification
+  - IconOverflow
+  - IconPeople
+  - IconPersonAdd
+  - IconPersonVerified
+  - IconPhone
+  - IconPositive
+  - IconPrint
+  - IconProfile
+  - IconPromote
+  - IconRecommended
+  - IconRefresh
+  - IconResume
+  - IconSearch
+  - IconSecurity
+  - IconSend
+  - IconSent
+  - IconSettings
+  - IconShare
+  - IconSocialFacebook
+  - IconSocialGitHub
+  - IconSocialInstagram
+  - IconSocialLinkedIn
+  - IconSocialMedium
+  - IconSocialTwitter
+  - IconStar
+  - IconStatistics
+  - IconSubCategory
+  - IconTag
+  - IconTick
+  - IconTime
+  - IconUpload
+  - IconVideo
+  - IconVisibility
+  - IconWorkExperience
+---
+
+**Text:** Add brandAccent tone to Text and Icons
+
+**EXAMPLE USAGE:**
+```jsx
+<Text tone="brandAccent">...</Text>
+```
+
+
--- a/.changeset/many-keys-sort.md
+++ b/.changeset/many-keys-sort.md
@@ -15,10 +15,9 @@ Usage of `TextLink` or `TextLinkButton` inside of an `Actions` container should 
 Previously when a `TextLink` or `TextLinkButton` was placed inside of an `Actions` container, it would be given a custom layout to align with the `Button` elements. We are deprecating this behaviour.
 
 **MIGRATION GUIDE:**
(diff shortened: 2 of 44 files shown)


.changeset/many-keys-sort.md
```diff
@@ -0,0 +1,37 @@
...
+---
+
+**TextLink,TextLinkButton:** Deprecate inside of Actions in favour of transparent Button
+
+Usage of `TextLink` or `TextLinkButton` inside of an `Actions` container should now use a `Button` with a `variant` of `transparent`.
+
+Previously when a `TextLink` or `TextLinkButton` was placed inside of an `Actions` container, it would be given a custom layout to align with the `Button` elements. We are deprecating this behaviour.
+
+**MIGRATION GUIDE:**
+Going forward `Actions` should only contain `Button` elements. To migrate towards this, both `TextLink` and `TextLinkButton` should now use either a `ButtonLink` or `Button` respectively, with a `variant` or `transparent`.
```
markdalgleish, line 18:
```suggestion
Going forward `Actions` should only contain `Button` elements. To migrate towards this, both `TextLink` and `TextLinkButton` should now use either a `ButtonLink` or `Button` respectively, with a `variant` of `transparent`.
```

.changeset/many-keys-sort.md
```diff
@@ -0,0 +1,37 @@
...
+
+#### TextLink
+Can be replicated with a `variant` of `standard` (default).
```
markdalgleish, line 21: Should this line be removed?

.changeset/pretty-beers-reply.md
```diff
@@ -0,0 +1,59 @@
...
+
+---
+updated:
+  - Button
+  - ButtonLink
+---
+
+**Button,ButtonLink:** Add variant to Button and deprecate weight
+
+Introduces a new `variant` prop to `Button`/`ButtonLink` giving consumers a single prop to use for selecting the visual style of the button. Choose from `solid` (default), `ghost`, `soft` or `transparent`. The colour of the button is now consistently controlled via the `tone` prop, with supported values being `brandAccent`, `critical` or `undefined`.
```
markdalgleish, line 13:
```suggestion
Introduces a new `variant` prop to `Button`/`ButtonLink` giving consumers a single prop to use for selecting the visual style of the button. Choose from `solid` (default), `ghost`, `soft` or `transparent`. The colour of the button is now consistently controlled via the `tone` prop, with supported values being `"brandAccent"`, `"critical"` or `undefined`.
```

lib/components/Box/Box.docs.tsx
```diff
@@ -407,7 +407,7 @@ const docs: ComponentDocs = {
                 body:
                   'Used for elements that need to match the body background.',
                 brand: 'Used for branding larger areas of the screen.',
-                brandAccent: 'Used for branding smaller areas of the screen.',
+                brandAccent: 'Used for hero elements on the screen.',
```
markdalgleish, line 410: Ah, nice one 👏

lib/components/Button/Button.migration.md
```diff
@@ -2,41 +2,39 @@
...
 
 -<Button color="primary">Secondary</Button>
 +<Button>Secondary</Button>
 
 -<Button color="tertiary">Tertiary</Button>
-+<Button weight="weak">Tertiary</Button>
++<Button variant="ghost">Tertiary</Button>
+
+-<Button compact>Small</Button>
++<Button size="small">Small</Button>
```
markdalgleish, line 37: Nice one 👏

.changeset/light-oranges-call.md
```diff
@@ -0,0 +1,90 @@
...
+  - IconTag
+  - IconTick
+  - IconTime
+  - IconUpload
+  - IconVideo
+  - IconVisibility
+  - IconWorkExperience
+---
+
+**Text:** Add brandAccent tone to Text and Icons
```
michaeltaranto, line 83:
```suggestion
**Text, Icons:** Add brandAccent tone to Text and Icons
```

lib/components/ButtonRenderer/ButtonRenderer.tsx
```diff
@@ -206,10 +248,48 @@ export interface PrivateButtonRendererProps {
   ) => ReactNode;
 }
 
+type PropMapper = {
```
markdalgleish, line 251: This name threw me because this type isn't actually mapping anything. Thoughts on something like `VisualProps` or `ResolveToneAndVariantOptions`, or even just inlining it so we don't have to name it?

markdalgleish (approved):
Raised a minor bikeshed, but not enough to block an approval. This looks awesome 👏

lib/components/ButtonRenderer/ButtonRenderer.tsx
michaeltaranto: Yeah good find, i never went back to name it. I'll just inline it.

PR #897, opened 2021-02-12
Add TooltipRenderer component

markdalgleish:
See changeset for details.


lib/components/TooltipRenderer/TooltipRenderer.docs.tsx
```diff
@@ -0,0 +1,70 @@
...
+    </Stack>
+  ),
+  additional: [
+    {
+      label: 'Development considerations',
+      description: (
+        <Text>
+          The <Strong>triggerProps</Strong> object provides low level props that
+          are currently only supported on{' '}
+          <TextLink href="/components/Box">Box.</TextLink>
```
michaeltaranto, line 63: Might be worth calling out the props is requires here, so consumers have the option to handle them in their own component abstractions.

michaeltaranto (approved):
💡  📎   🚀

PR #901, opened 2021-02-15
Add 'placement' to TooltipRenderer, add 'cursor="default"' to Box

markdalgleish:
See changesets for details.

### DIFF
--- a/.changeset/metal-pans-breathe.md
+++ b/.changeset/metal-pans-breathe.md
@@ -1,0 +1,26 @@
+---
+'braid-design-system': minor
+---
+
+---
+updated:
+  - TooltipRenderer
+---
+
+**TooltipRenderer:** Add `placement` prop, support `top` and `bottom`. Set default placement to `top`.
+
+**EXAMPLE USAGE**
+
+```tsx
+<TooltipRenderer
+  id={id}
+  tooltip={<Text>This is a tooltip!</Text>}
+  placement="top"
+>
+  {({ triggerProps }) => (
+    <Box aria-label="Help" {...triggerProps}>
+      <IconHelp />
+    </Box>
+  )}
+</TooltipRenderer>
+```
--- a/.changeset/metal-pans-breathe.md
+++ b/.changeset/metal-pans-breathe.md
@@ -15,7 +15,7 @@ updated:
 <TooltipRenderer
   id={id}
   tooltip={<Text>This is a tooltip!</Text>}
-  placement="top"
+  placement="bottom"
 >
   {({ triggerProps }) => (
     <Box aria-label="Help" {...triggerProps}>
--- a/lib/components/TooltipRenderer/TooltipRenderer.docs.tsx
+++ b/lib/components/TooltipRenderer/TooltipRenderer.docs.tsx
@@ -12,21 +12,14 @@ import {
   Strong,
 } from '..';
 import { TooltipContent } from './TooltipRenderer';
+import { constants } from './TooltipRenderer.treat';
 
 const docs: ComponentDocs = {
   category: 'Content',
   Example: ({ id }) =>
     source(
       <Inline space="small">
-        <TooltipRenderer
-          id={id}
-          tooltip={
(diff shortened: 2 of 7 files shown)


lib/components/TooltipRenderer/TooltipRenderer.treat.ts
```diff
@@ -1,18 +1,22 @@
...
 });
 
 export const verticalOffsetBeforeEntrance = style({
-  transform: 'translateZ(0) translateY(-4px)',
+  transform: 'translateZ(0) translateY(4px)',
   selectors: {
-    '[data-popper-placement*=top] &': {
-      transform: 'translateZ(0) translateY(4px)',
+    '[data-popper-placement*=bottom] &': {
+      transform: 'translateZ(0) translateY(-4px)',
```
tgdev, line 19: should the -4px value be a theme variable (grid?) or will it work across all themes at -4px?

markdalgleish: This transition isn't following the grid, it's just a coincidence :)

tgdev (approved):
Nice work! X and Y positioning would be ideal (which might also open things up to responsive positioning array) but this is a great start. Thanks!

markdalgleish:
@tgdev I did actually work on `"left"` and `"right"` placement too, but found that it's really tough to do well responsively since it's highly likely the tooltip won't fit on the screen. Decided I'd wait for an actual use case rather than guessing how best to handle it.

.changeset/metal-pans-breathe.md
```diff
@@ -0,0 +1,26 @@
...
+
+**TooltipRenderer:** Add `placement` prop, support `top` and `bottom`. Set default placement to `top`.
+
+**EXAMPLE USAGE**
+
+```tsx
+<TooltipRenderer
+  id={id}
+  tooltip={<Text>This is a tooltip!</Text>}
+  placement="top"
```
michaeltaranto, line 18: Suggest we make the example different from the default value.
```suggestion
  placement="bottom"
```

markdalgleish: Nice catch. This wasn't the default when I wrote it.

lib/components/TooltipRenderer/TooltipRenderer.docs.tsx
```diff
@@ -56,13 +49,78 @@ const docs: ComponentDocs = {
...
+      ),
+    },
+    {
+      label: 'Placement',
       description: (
         <Text>
-          You can use multiple text elements and layout components to create
-          more custom layouts.
+          You can set the <Strong>placement</Strong> of the tooltip to{' '}
+          <Strong>top</Strong> or <Strong>bottom</Strong>. Whichever direction
```
michaeltaranto, line 65: Worth calling out that `top` is the default here in the docs?

lib/components/TooltipRenderer/TooltipRenderer.docs.tsx
```diff
@@ -56,13 +49,78 @@ const docs: ComponentDocs = {
...
+          you choose, the tooltip will be centred against the trigger element.
         </Text>
       ),
+      Example: ({ id }) =>
+        source(
+          <Inline space="small">
+            <TooltipRenderer
+              id={`${id}_1`}
+              placement="top"
```
michaeltaranto, line 74:
```suggestion
              placement="bottom"
```

lib/components/TooltipRenderer/TooltipRenderer.docs.tsx
```diff
@@ -56,13 +49,78 @@ const docs: ComponentDocs = {
...
+              tooltip={<Text>The placement is “top”</Text>}
```
michaeltaranto, line 75:
```suggestion
              placement="bottom"
              tooltip={<Text>The placement is “bottom”</Text>}
```

michaeltaranto (approved):

PR #909, opened 2021-02-19
Tabs, Tiles: Remove Card from screenshot tests

michaeltaranto:
These tests should not be bound to the visual implementation of a `Card`, so decoupling before upcoming `Card` changes.

### DIFF
--- a/lib/components/Tiles/Tiles.screenshots.tsx
+++ b/lib/components/Tiles/Tiles.screenshots.tsx
@@ -1,6 +1,6 @@
 import React, { Fragment } from 'react';
 import { ComponentScreenshot } from '../../../site/src/types';
-import { Tiles, Card, Text } from '../';
+import { Tiles, Box, Text } from '../';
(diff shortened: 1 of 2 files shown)


lib/components/Tiles/Tiles.screenshots.tsx
```diff
@@ -34,9 +34,9 @@ export const screenshots: ComponentScreenshot = {
       Example: () => (
         <Tiles space={['none', 'small']} columns={[1, 2]} dividers>
           {[...new Array(2 * exampleRows)].map((_, i) => (
-            <Card key={i}>
+            <Box background="card" padding="gutter" key={i}>
```
markdalgleish, line 37: Would Placeholder be better here then?

michaeltaranto: Because of the border on Placeholder it clashes with the divider. So felt the visual test was stronger like this

markdalgleish (approved):

PR #911, opened 2021-02-23
Docs: Add docs for icon props on List

michaeltaranto:


markdalgleish:
Looking at it with fresh eyes, feel like the nested lists section should be last since it's the least likely to be used—what do you think? Felt weird that it was above "types of lists", but feel like it should be even further down than that.

markdalgleish (approved):

PR #925, opened 2021-04-20
Accordion: Add size, tone, space and dividers props

markdalgleish:
See changeset for details.


michaeltaranto (approved):
Nice one. If you feel like putting the boolean guards on the playroom versions great, but either way this is heaps more flexible now!

PR #926, opened 2021-04-23
Pagination: Add keyline to improve active page indicator contrast

michaeltaranto:


lib/components/Pagination/Pagination.tsx
```diff
@@ -94,6 +96,17 @@ const Page = ({ number, current }: { number: number; current: boolean }) => {
         borderRadius="standard"
         className={[styles.background, current ? styles.current : undefined]}
       />
+      <Overlay
+        component="span"
+        transition={current ? undefined : 'fast'}
```
markdalgleish, line 101: Not sure this should have a transition? Looks like my screen is ghosting or something.

michaeltaranto:
```suggestion
```

markdalgleish (approved):

PR #959, opened 2021-07-05
Support nested Dialogs & Drawers

michaeltaranto:


lib/components/private/Modal/Modal.css.ts
```diff
@@ -4,6 +4,14 @@ import { externalGutter } from './ModalExternalGutter';
 import { TextBreakpoint } from '../../../themes/tokenType';
 import { responsiveStyle } from '../../../css/responsiveStyle';
 import { vars } from '../../../themes/vars.css';
+import { atoms } from '../../../css/atoms/atoms';
+
+export const fixedStackingContext = atoms({
```
markdalgleish, line 9: Ooh, this is a great use case for `atoms` 👏

PR #977, opened 2021-08-06
Add TooltipRenderer snippet

jackhurley23:
Since TooltipRenderer is hard to write by hand it would be useful to have a Snippet in playroom


markdalgleish (approved):

markdalgleish:
Thank you! 🙏

PR #984, opened 2021-08-17
 Update to sku 11

mattcompiles:


.eslintignore
```diff
@@ -1,8 +1,9 @@
+site/dist/playroom/
+
```
michaeltaranto, line 2: Unnecessary given `site/dist` below.

.gitignore
```diff
@@ -7,6 +7,7 @@ package-lock.json
 .idea/
 componentDocs.json
 lib/components/icons/*/*Svg.tsx
+site/dist/playroom/
```
michaeltaranto, line 10: As above

.prettierignore
```diff
@@ -4,8 +4,8 @@ coverage/
 dist-storybook/
 report/
 site/dist/
-site/dist/playroom/
 # end managed by sku
 CHANGELOG.md
 .changeset/*.md
 componentUpdates.json
+site/dist/playroom/
```
michaeltaranto, line 11: As above

lib/components/MenuRenderer/testHelper.tsx
```diff
@@ -1,5 +1,6 @@
 import '@testing-library/jest-dom/extend-expect';
-import React, { useState } from 'react';
+import { useState } from 'react';
+import * as React from 'react';
```
michaeltaranto, line 3: Maybe destructure `FunctionComponent` so we dont have to have this separate import

site/src/render.tsx
```diff
@@ -1,5 +1,5 @@
 import { Render } from 'sku';
-import React from 'react';
+import * as React from 'react';
```
michaeltaranto, line 2: Can just import `ReactElement` directly here

jahredhope (approved):

PR #1005, opened 2021-09-27
Use color palette for code blocks, fix release page warnings

michaeltaranto:
Update the site to use the internal palette for styling the code blocks. Also using custom styles on the release page to get around the nested Text component warnings on the Releases page in dev, due to markdown-based rendering.

### DIFF
--- a/site/src/App/Markdown/Markdown.css.ts
+++ b/site/src/App/Markdown/Markdown.css.ts
@@ -1,0 +1,31 @@
+import { style, composeStyles } from '@vanilla-extract/css';
+import { createTextStyle } from '@capsizecss/vanilla-extract';
+import { vars } from '../../../../lib/themes/vars.css';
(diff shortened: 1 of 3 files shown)


site/src/App/Markdown/Markdown.css.ts
```diff
@@ -0,0 +1,31 @@
+import { style, composeStyles } from '@vanilla-extract/css';
```
mattcompiles, line 1: Don't need to use `composeStyles` anymore.

mattcompiles (approved):

site/src/App/Markdown/Markdown.css.ts
michaeltaranto: Will update this separately with a PR that bumps vanilla and rolls it through.

PR #1011, opened 2021-09-28
Update theme colour tokens with improved semantics

michaeltaranto:
**Box, atoms, vars:** Update theme colour tokens with improved semantics.

Simplifies the semantics of the colour-based tokens to support upcoming colour mode work. Changes to the property values on `backgroundColor` and `borderColor` has flow on affects to the apis on `Box` and `atoms`.


mattcompiles (approved):
👌

PR #1012, opened 2021-09-29
Replace BackgroundProvider with Box background of customDark/customLight

michaeltaranto:
**BackgroundProvider** Removed in favour of setting a `background` of `customDark`/`customLight` directly on the `Box` that has the custom background color.

**MIGRATION GUIDE**

```diff
-<Box style={{ backgroundColor: 'purple' }}>
+<Box background="customDark" style={{ backgroundColor: 'purple' }}>
-   <BackgroundProvider value="UNKNOWN_DARK">
     <Text>Inverted text given dark context</Text>
-   </BackgroundProvider>
 </Box>
```


jahredhope (approved):

lib/components/Box/ColoredBox.tsx
```diff
@@ -0,0 +1,36 @@
...
+  component: NonNullable<BoxProps['component']>;
+  background: NonNullable<BoxProps['background']>;
+}
+
+export const useColoredBoxClasses = (
+  background: ColoredBoxProps['background'],
+) => [
+  atoms({
+    background:
+      background === 'customDark' || background === 'customLight'
```
mattcompiles, line 16: Feel this would be nicer if we hoisted the condition out of the `atoms` call?

lib/components/Box/ColoredBox.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { atoms } from '../../css/atoms/atoms';
+import { renderBackgroundProvider } from './BackgroundContext';
+import { BoxProps } from './Box';
+
+export interface ColoredBoxProps extends BoxProps {
+  component: NonNullable<BoxProps['component']>;
+  background: NonNullable<BoxProps['background']>;
+}
+
+export const useColoredBoxClasses = (
```
mattcompiles, line 11: Could we remove the "use" prefix as this isn't a hook so it's a bit misleading?

lib/components/Box/ColoredBox.tsx
```diff
@@ -0,0 +1,36 @@
...
+        : background,
+  }),
+];
+
+export const ColoredBox = forwardRef<HTMLElement, ColoredBoxProps>(
+  ({ component, background, className, ...props }, ref) => {
+    const colorClasses = useColoredBoxClasses(background);
+
+    const element = createElement(component, {
+      className: `${className} ${colorClasses}`,
```
mattcompiles, line 27: This could lead to extra whitespace as `className` can be empty yeah?

lib/components/Box/ColoredBox.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { BoxProps } from './Box';
+
+export interface ColoredBoxProps extends BoxProps {
+  component: NonNullable<BoxProps['component']>;
+  background: NonNullable<BoxProps['background']>;
+}
+
+export const useColoredBoxClasses = (
+  background: ColoredBoxProps['background'],
+) => [
```
mattcompiles, line 13: Why does this return an array?

mattcompiles (approved):

PR #1014, opened 2021-09-29
Move BraidTestProvider to separate entry

michaeltaranto:


generate-component-docs/__snapshots__/contract.test.ts.snap
```diff
@@ -1835,36 +1835,6 @@ Object {
...
-        | "jobsDb"
-        | "occ"
-        | "seekAnz"
-        | "seekBusiness"
-        | "seekUnifiedBeta"
-        | "wireframe"
-},
-}
-`;
-
```
michaeltaranto, line 1867: To be discussed.

mattcompiles (approved):

PR #1023, opened 2021-11-02
Move @types/react to devDependencies

michaeltaranto:
Braid requires consumers to provide React, therefore they should also provide the appropriate version of @types/react rather than rely on the version installed in Braid.


benjervis (approved):
🚀

PR #1063, opened 2022-02-15
Playroom: Better handling of partial primitive space values

michaeltaranto:


benjervis:
Is it possible to have this set up so that it does the error highlight thing in the editor? So much nicer that it doesn't blow up, but it would be good to know that I'm not using a valid value for this sort of prop

Screen Shot 2022-02-22 at 11 19 25 am

michaeltaranto:
Unfortunately there is no programatic access between our components and the Playroom editor, meaning we cant mark a specific line as an error. That kind of UX polish would have to come with a migration away from CodeMirror and to something that has more validation options.

benjervis (approved):

PR #1064, opened 2022-02-16
Columns: Optional Column children & component span support

michaeltaranto:
**Columns:** Make child Column optional and default internally

As an optimisation to the default use of `Columns`, it is now only necessary use `Column` as a child element when customising the `width` property. If the use case is equally distributed columns, wrapping in `Columns` will now be enough.

**EXAMPLE USAGE:**
The following will now result in two columns equally distributed:
```jsx
<Columns space="medium">
  <Card>...</Card>
  <Card>...</Card>
</Columns>
```

or wrap a single child in a `Column` to customise it's width:

```jsx
<Columns space="medium">
  <Card>...</Card>
  <Card>...</Card>
  <Column width="content">
    <OverflowMenu />
  </Column>
</Columns>
```
---
 **Columns:** Support using span elements via component prop

Setting the `component` prop to `span` will now ensure all elements controlled by `Columns` are `span`s. This is useful when using layout components inside dom elements that should not contain `div`s from a HTML validation perspective.

**EXAMPLE USAGE:**
```jsx
<Columns space="medium" component="span">
  ...
</Columns>
```

### DIFF
--- a/.changeset/brown-llamas-raise.md
+++ b/.changeset/brown-llamas-raise.md
@@ -1,0 +1,33 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated:
+  - Columns
+---
+
+**Columns:** Make child Column optional and default internally
+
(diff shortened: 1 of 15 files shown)


benjervis (approved):
👏

PR #1071, opened 2022-02-23
Remove redundant use of Column

michaeltaranto:
**AccordionItem, Alert, MonthPicker, Notice, Dialog, Drawer, Toast:** Remove default Column elements

Remove redundant `Column` elements when not customising the `width` prop.

---

**"Examples" on docs site**

Revisited some of the steps/flow around the `Job Summary` and `Marketing Banner` examples now that `Column` is only necessary when customising the `width` prop.

### DIFF
--- a/.changeset/clean-guests-lie.md
+++ b/.changeset/clean-guests-lie.md
@@ -1,0 +1,18 @@
+---
+'braid-design-system': patch
+---
+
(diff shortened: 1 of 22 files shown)


michaeltaranto:
Holding off until v32

PR #1088, opened 2022-04-06
MenuRenderer: Guard against open/close handlers firing incorrectly

michaeltaranto:


lib/components/MenuRenderer/MenuRenderer.tsx
```diff
@@ -166,16 +166,20 @@ export const MenuRenderer = ({
...
 
   useEffect(() => {
-    if (open) {
-      hasOpened.current = true;
+    if (lastOpen.current === open) {
+      return;
+    }
 
+    if (open) {
       if (typeof onOpen === 'function') {
```
askoufis, line 174: Any reason not to collapse these two `if`s into one?

askoufis (approved):
Looks good, just one non-blocking nitpick.

lib/components/MenuRenderer/MenuRenderer.tsx
michaeltaranto: Yep, because it `onOpen` is not a function we don't want to trigger the else clause.

askoufis: But wouldn't `open && typeof onOpen === 'function'` achieve that, or am I missing something?

PR #1099, opened 2022-04-21
upgrade: Make deprecations configurable by version

michaeltaranto:
Parameterise deprecations to support the `upgrade` command providing different deprecations for different versions.

### DIFF
--- a/codemod/src/plugin-deprecate/plugin-deprecate-props.ts
+++ b/codemod/src/plugin-deprecate/plugin-deprecate-props.ts
@@ -1,12 +1,12 @@
 import type { PluginObj, PluginPass } from '@babel/core';
 import { types as t } from '@babel/core';
-import { isComponentDeprecated, isPropDeprecated } from './deprecationMap';
 import { subVisitor } from './subVisitor';
 import { deArray, StringLiteralPath, updateStringLiteral } from './helpers';
 
 interface Context extends PluginPass {
   importNames: Map<string, string>;
   namespace: string | null;
+  deprecations: Record<string, Record<string, Record<string, string>>>;
 }
 
 export default function (): PluginObj<Context> {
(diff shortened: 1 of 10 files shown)


codemod/src/plugin-deprecate/plugin-deprecate-props.ts
```diff
@@ -33,7 +40,7 @@ export default function (): PluginObj<Context> {
                 if (
                   t.isImportSpecifier(specifier) &&
                   t.isIdentifier(specifier.imported) &&
-                  isComponentDeprecated(specifier.imported.name)
+                  Boolean(this.deprecations[specifier.imported.name])
```
askoufis, line 43: This is optionally chained in deprecate-atoms, should it be the same here too?
```suggestion
                  Boolean(this.deprecations?.[specifier.imported.name])
```

codemod/src/plugin-deprecate/plugin-deprecate-props.ts
```diff
@@ -1,12 +1,12 @@
 import type { PluginObj, PluginPass } from '@babel/core';
 import { types as t } from '@babel/core';
-import { isComponentDeprecated, isPropDeprecated } from './deprecationMap';
 import { subVisitor } from './subVisitor';
 import { deArray, StringLiteralPath, updateStringLiteral } from './helpers';
 
 interface Context extends PluginPass {
   importNames: Map<string, string>;
   namespace: string | null;
+  deprecations: Record<string, Record<string, Record<string, string>>>;
```
askoufis, line 9: I feel like this type is used often enough that it should get its own alias.

codemod/src/plugin-deprecate/plugin-deprecate-props.ts
michaeltaranto: I'll go the other way, deprecations are required, i'll remove the optional chain (and defaulting to blank object) from the other locations.

askoufis (approved):

PR #1103, opened 2022-05-09
Button, ButtonLink: Improve Button `bleed`

michaeltaranto:
Previously the `bleedY` prop allowed the background of `Button` to bleed vertically into the surrounding layout. This worked well for all variants except `transparent`, which needed to bleed horizontally as well.

To support this we have introduced the `bleed` prop which will apply the bleed based on the `variant`. We have also deprecated `bleedY` which will be removed in a future release.

**EXAMPLE USAGE:**
```diff
 <Button
-  bleedY
+  bleed
  {...}
 >
   Button
 </Button>
```

**MIGRATION GUIDE**

Migration from `bleedY` to `bleed` can be automated by running the Braid upgrade command, passing the `v31.11` version. You must provide a glob to target your project’s source files. For example:

```
yarn braid-upgrade v31.11 "**/*.{ts,tsx}"
```

It is recommended to visually review the any `Button` usage with the `transparent` variant, to ensure the layout is as expected.

### DIFF
--- a/codemod/src/plugin-deprecate/plugin-prop-rename.ts
+++ b/codemod/src/plugin-deprecate/plugin-prop-rename.ts
@@ -1,0 +1,173 @@
+import type { PluginObj, PluginPass, Visitor } from '@babel/core';
+import { types as t } from '@babel/core';
+import type { NodePath } from '@babel/traverse';
+import { renderRecursiveDepthWarning } from '../warning-renderer/warning';
+import { deArray } from './helpers';
+
+interface Context extends PluginPass {
+  importNames: Map<string, string>;
+  namespace: string | null;
+  renames: Record<string, Record<string, string>>;
+}
+
+interface SubVisitorContext extends Context {
+  componentName: string;
+  propName?: string;
+  propLocation?: t.SourceLocation | null;
+  recurses: number;
+}
+const subVisitor: Visitor<SubVisitorContext> = {
+  JSXIdentifier(p) {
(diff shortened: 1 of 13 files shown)


codemod/src/plugin-deprecate/plugin-prop-rename.ts
```diff
@@ -0,0 +1,173 @@
...
+
+interface SubVisitorContext extends Context {
+  componentName: string;
+  propName?: string;
+  propLocation?: t.SourceLocation | null;
+  recurses: number;
+}
+const subVisitor: Visitor<SubVisitorContext> = {
+  JSXIdentifier(p) {
+    const newValue = this.renames[this.componentName]?.[p.node.name];
```
askoufis, line 21: Would this also rename a component called `bleedY` to `bleed`?  Probably not worth guarding against that case since it's quite unlikely.

michaeltaranto: No i don't believe so. Its guarding against the prop look up on possible `undefined` (if there is no rename specified for the discovered component). We then check if there is a rename available just below it as well.

Happy to talk it through and add additional tests if you have an example in mind.

askoufis: Oh right, the prop name is scoped to a component. So you'd have to import `Button as bleedY` for it to do that.

askoufis (approved):

PR #1104, opened 2022-05-10
`apac` and `seekBusiness` themes: Update colour palette

michaeltaranto:
The colours used in these themes have been updated to the latest design standards as they were subtly off due to coming from an incorrect source.

### Developer notes:
- Removed unused internal swatches from earlier research version of SEEK palette.


askoufis (approved):
🌈

PR #1114, opened 2022-05-20
React 18 support

michaeltaranto:
### Changes
- Add React 18 to peer deps version list
- Bump as many react libraries to fix peer dep warnings
- No changes necessary to system from consumer side, with exception for `Dropdown` needing to wrap children in a `Fragment`.

### Site fixes
- Migrate site from `react-head` to `react-helmet-async` due to TS incompatibility, but it also comes with its own types.

### Testing fixes
- Migrate some tests to use `@testing-library/user-event` for their interactions.
- The `@testing-libary/react-hooks` library is now built into the `@testing-library/react` package.
- Exclude props starting with `__@` from the contract snapshot, avoiding the ever incrementing indexes in the `filterSuggestions` snapshot.

### Additional note
There are still a handful of deps that complain about React peer dep version, but are functionally ok. Will update as new versions become available.


site/src/client.tsx
```diff
@@ -38,16 +38,16 @@ const ScrollManager = ({ children }: { children: ReactNode }) => {
 export default (app: RenderContext) => {
   initUpdates(new Date(app.renderDate), app.versionMap, app.currentVersion);
 
-  hydrate(
-    <HeadProvider>
+  hydrateRoot(
+    document.getElementById('app')!,
```
benjervis, line 42: Is this `!` a real thing, or typo?

site/src/render.tsx
```diff
@@ -76,23 +76,22 @@ const skuRender: Render<RenderContext> = {
...
     ).join('');
+    const { helmet } = helmetContext;
 
     return dedent`
       <!doctype html>
       <html lang="en">
         <head>
           ${__experimentalDarkMode__}
-          ${
-            // @ts-expect-error
```
benjervis, line 92: 👏

lib/components/Stepper/Stepper.test.tsx
```diff
@@ -190,6 +190,8 @@ describe('Stepper', () => {
       await userEvent.keyboard('{arrowleft}');
       expect(secondStep).toHaveFocus();
       await userEvent.tab();
+      expect(thirdStep).toHaveFocus();
```
benjervis, line 193: Did we change something to give the Stepper an extra step?

benjervis (approved):
🎉

site/src/client.tsx
michaeltaranto: Haha, nah i got this from their [migration guide](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis). I originally added a null check on the `getElementById` call, but then spotted this in their docs.

lib/components/Stepper/Stepper.test.tsx
michaeltaranto: I havent been able to put my finger on this one, why this is passing in master has me confused as the test now replicates what happens in the browser (both production and this branch).

Will have a poke around and see if i can provide some more clarity.

site/src/client.tsx
benjervis: Oh the null thing, forgot about types 😅

PR #1125, opened 2022-06-22
Extend changeset DX to add braid release front matter

michaeltaranto:
Extend the changeset creation experience in the cli to include prompts for creating and inserting the additional front matter block used to Braid to add release information to the website.

Running `pnpm changeset` will now prompt for the following and enrich the changeset accordingly:
- What is the impact to the top level api of Braid? (updated / new / no change)
- Add scope list (inferred from original changeset if formatted as `<scope>: <message>`, otherwise prompt for scope list explicitly)
- Add release notes to be included on site
- Add `Example Usage` code block if required


askoufis (approved):
🚀

PR #1126, opened 2022-07-10
Stepper: Add align prop

michaeltaranto:
Provide the `align` prop which now includes support for `left` alignment.

**EXAMPLE USAGE:**
```jsx
<Stepper align="left">
  ...
</Stepper>
```

### DIFF
--- a/lib/components/Stepper/Stepper.docs.tsx
+++ b/lib/components/Stepper/Stepper.docs.tsx
@@ -288,7 +288,7 @@ const docs: ComponentDocs = {
         <>
           <Text>
             The Stepper is center-aligned by default, but this can be customised
-            via the align prop.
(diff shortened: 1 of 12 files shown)


lib/components/Stepper/Stepper.docs.tsx
```diff
@@ -282,6 +282,31 @@ const docs: ComponentDocs = {
           </Stepper>,
         ),
     },
+    {
+      label: 'Alignment',
+      description: (
+        <>
+          <Text>
+            The Stepper is center-aligned by default, but this can be customised
+            via the align prop.
```
askoufis, line 291: Should `align` be emphasised here since it's a prop name?
```suggestion
            via the <Strong>align</Strong> prop.
```

PR #1130, opened 2022-07-19
IconArrow: Add new component

michaeltaranto:
Add new `IconArrow` component. The orientation of the arrow can be controlled using the `direction` prop.

**EXAMPLE USAGE:**
```jsx
<IconArrow direction="left" />
```

#### Notes
Also fix `useToast` docs site hooks rendering issue


benjervis (approved):

lib/components/useToast/useToast.docs.tsx
```diff
@@ -15,13 +15,11 @@ import { useThemeSettings } from '../../../site/src/App/ThemeSetting';
...
 import Toast, { toastDuration } from './Toast';
 import source from '../../utils/source.macro';
-import { useToast } from './ToastContext';
 
 const docs: ComponentDocs = {
   category: 'Content',
-  Example: ({ id }) => {
+  Example: ({ id, showToast }) => {
     const { theme } = useThemeSettings();
-    const showToast = useToast();
```
mrm007, line 24: It seems like the example doesn't use `useToast` anymore.

michaeltaranto: A little confusing i admit. If you expand the code section on the docs it doesnt show the code that was changed here, rather it shows the usage, e.g. `showToast`. This is consistent with the other examples down the page too. The reason for this is so the code can be copied into Playroom (where there are limitations around Hook usage), but I agree it doesnt do the documentation any favours.

Further down under [Development Considerations](https://seek-oss.github.io/braid-design-system/components/useToast#development-considerations) we show how to use the Hook, and talk about `showToast` under [Playroom Prototyping](https://seek-oss.github.io/braid-design-system/components/useToast#playroom-prototyping).

I have a few ideas though that we can chat about offline.

PR #1135, opened 2022-08-11
Fix clashing ids in Storybook and Playroom

mrm007:
Found this behaviour when testing Storybook in the monorepo branch:

<details>
<summary>Screen recording</summary>

checkboxes
 

</details>

### DIFF
--- a/lib/components/Checkbox/Checkbox.docs.tsx
+++ b/lib/components/Checkbox/Checkbox.docs.tsx
@@ -147,21 +147,21 @@ const docs: ComponentDocs = {
           to the <Strong>disabled</Strong> prop.
         </Text>
       ),
-      Example: ({ id }) =>
+      Example: ({ id, toggleState }) =>
         source(
           <Stack space="medium">
             <Checkbox
               id={`${id}_unchecked`}
               disabled={true}
               checked={false}
-              onChange={() => {}}
+              onChange={() => toggleState('unchecked')}
               label="Unchecked"
             />
             <Checkbox
               id={`${id}_checked`}
               disabled={true}
               checked={true}
-              onChange={() => {}}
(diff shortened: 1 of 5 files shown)


lib/components/Checkbox/Checkbox.docs.tsx
```diff
@@ -147,21 +147,21 @@ const docs: ComponentDocs = {
...
+              onChange={() => {}}
               label="Unchecked"
             />
             <Checkbox
-              id={id}
+              id={`${id}_checked`}
               disabled={true}
               checked={true}
-              onChange={() => toggleState('checked')}
+              onChange={() => {}}
```
askoufis, line 164: Might be worth keeping the `toggleState` here. If someone is copying and pasting this example, they'll likely conditionally, rather than permanently, disable the checkbox, so IMO the state should still be able to change.

lib/components/Checkbox/Checkbox.docs.tsx
```diff
@@ -151,14 +151,14 @@ const docs: ComponentDocs = {
         source(
           <Stack space="medium">
             <Checkbox
-              id={id}
+              id={`${id}_unchecked`}
               disabled={true}
               checked={false}
-              onChange={() => toggleState('checked')}
+              onChange={() => toggleState('unchecked')}
```
askoufis, line 157: Nitpick: feel like this should stay as `checked`. Not that the state name matters too much, but I think it makes more sense to match the name of the prop so you don't have to negate it

mrm007: There are two Checkboxes here. Having both with `checked` means they are sharing state.

askoufis (approved):

PR #1140, opened 2022-08-24
Autosuggest: Add configurable message for no suggestions

michaeltaranto:
Provides consumers a way to give the user more context when no suggestions are available. The `noSuggestionsMessage` prop accepts either a simple message or a more structured prompt by providing an object containing **title** and **description**.

This message is only displayed when there are no available suggestions provided.

**EXAMPLE USAGE:**
For the simple case:

```jsx
<Autosuggest
  ...
  suggestions={[]}
  noSuggestionsMessage="No results found"
/>
```

Or, for more a structured prompt:

```jsx
<Autosuggest
  ...
  suggestions={[]}
  noSuggestionsMessage={{
    title: "No results found",
    description: "Try searching for something else",
  }}
/>
```

**MIGRATION GUIDE:**

In addition, the old mechanism allowing consumers to pass an object to `suggestions` containing a `message` has been deprecated. This will continue to work for now but will be removed in a future release.

It is recommended to migrate to the `noSuggestionsMessage` prop.

```diff
 <Autosuggest
   ...
-  suggestions={{ message: 'No results found' }}
+  noSuggestionsMessage="No results found"
 />
```

### DIFF
--- a/packages/braid-design-system/lib/components/Autosuggest/Autosuggest.docs.tsx
+++ b/packages/braid-design-system/lib/components/Autosuggest/Autosuggest.docs.tsx
@@ -601,9 +601,10 @@ const docs: ComponentDocs = {
             <Strong>noSuggestionsMessage</Strong> prop.
           </Text>
           <Text>
-            The message can either be a single piece of text, or a more
-            structured prompt by providing an object containing{' '}
-            <Strong>text</Strong> and <Strong>description</Strong>.
+            A simple message can be shown by providing a single piece of text.
+            Alternatively, a more structured prompt can be shown by providing an
+            object containing <Strong>title</Strong> and{' '}
+            <Strong>description</Strong>.
           </Text>
         </>
       ),
@@ -619,7 +620,7 @@ const docs: ComponentDocs = {
                 scrollToTopOnMobile
                 label="Label"
                 description="Focus the field to see a simple message"
-                id={id}
+                id={`${id}_noSuggestionsMessage1`}
                 value={getState('field1')}
                 onChange={setState('field1')}
                 onClear={() => resetState('field1')}
@@ -631,13 +632,13 @@ const docs: ComponentDocs = {
                 scrollToTopOnMobile
                 label="Label"
                 description="Focus the field to see more structured prompt"
-                id={id}
+                id={`${id}_noSuggestionsMessage2`}
                 value={getState('field2')}
                 onChange={setState('field2')}
                 onClear={() => resetState('field2')}
                 suggestions={[]}
                 noSuggestionsMessage={{
-                  text: 'No results found',
+                  title: 'No results found',
                   description: 'Try searching for something else',
                 }}
               />
--- a/packages/braid-design-system/lib/components/Autosuggest/Autosuggest.docs.tsx
+++ b/packages/braid-design-system/lib/components/Autosuggest/Autosuggest.docs.tsx
@@ -597,53 +597,51 @@ const docs: ComponentDocs = {
         <>
           <Text>
             If no suggestions are available and you’d like to provide messaging
-            to the user, you can provide an object with a{' '}
-            <Strong>message</Strong> property to the{' '}
-            <Strong>suggestions</Strong> prop.
+            to the user, you can provide it via the{' '}
+            <Strong>noSuggestionsMessage</Strong> prop.
+          </Text>
+          <Text>
+            The message can either be a single piece of text, or a more
+            structured prompt by providing an object containing{' '}
+            <Strong>text</Strong> and <Strong>description</Strong>.
           </Text>
         </>
       ),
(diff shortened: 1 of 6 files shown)


packages/braid-design-system/lib/components/Autosuggest/Autosuggest.tsx
```diff
@@ -292,11 +300,43 @@ export type AutosuggestLabelProps = FieldLabelVariant;
...
+      ? { text: noSuggestionsMessage }
+      : noSuggestionsMessage;
+  }
+
+  if ('message' in suggestionProp) {
+    const message = suggestionProp.message;
+    if (process.env.NODE_ENV !== 'production') {
+      // eslint-disable-next-line no-console
+      console.warn(
+        dedent`
```
mattcompiles, line 318: We definitely need something to make this easier haha. This is way too much effort.

mattcompiles (approved):
Looks good. 👏

packages/braid-design-system/lib/components/Autosuggest/Autosuggest.docs.tsx
```diff
@@ -597,53 +597,51 @@ const docs: ComponentDocs = {
...
+                onClear={() => resetState('field1')}
+                suggestions={[]}
+                noSuggestionsMessage="No results found"
+              />
+              <Autosuggest
+                showMobileBackdrop
+                scrollToTopOnMobile
+                label="Label"
+                description="Focus the field to see more structured prompt"
+                id={id}
```
mrm007, line 634: These two components have the same id, so `input`s, `button`s and other HTML elements created by `Autosuggest` will have the same id.

packages/braid-design-system/lib/components/Autosuggest/Autosuggest.docs.tsx
```diff
@@ -597,53 +597,51 @@ const docs: ComponentDocs = {
...
-            to the user, you can provide an object with a{' '}
-            <Strong>message</Strong> property to the{' '}
-            <Strong>suggestions</Strong> prop.
+            to the user, you can provide it via the{' '}
+            <Strong>noSuggestionsMessage</Strong> prop.
+          </Text>
+          <Text>
+            The message can either be a single piece of text, or a more
+            structured prompt by providing an object containing{' '}
+            <Strong>text</Strong> and <Strong>description</Strong>.
```
askoufis, line 606: Super-duper nitpicky, but this wording feels weird to me. I think it's because `by providing` is here without another verb that specifies "what you get by providing an object". I would've expected something like "xyz can be achieved by providing an object...".

```suggestion
            A simple message can be shown by providing a single piece of text.
            Alternatively, a more structured prompt can be shown by providing
            an object containing <Strong>text</Strong> and <Strong>description</Strong>.
```

Would need to update the changeset messaging too.

packages/braid-design-system/lib/components/Autosuggest/Autosuggest.tsx
```diff
@@ -648,7 +685,15 @@ export const Autosuggest = forwardRef(function <Value>(
...
-      announcements.push(translations.noSuggestionsAvailableAnnouncement);
+      announcements.push(
+        noSuggestionsMessage
+          ? `${noSuggestionsMessage.text}${
+              noSuggestionsMessage.description
+                ? ` ${noSuggestionsMessage.description}`
+                : ''
+            }`
+          : translations.noSuggestionsAvailableAnnouncement,
+      );
```
mrm007, line 696: Simpler if you just `.push` both each message separately, as they're [`.join`ed below](https://github.com/seek-oss/braid-design-system/blob/d3fb85d1d4776649443ba56a007adcba2cdd4f2d/packages/braid-design-system/lib/components/Autosuggest/Autosuggest.tsx#L842).

```suggestion
      announcements.push(
        noSuggestionsMessage
          ? [
              noSuggestionsMessage.text,
              noSuggestionsMessage.description,
            ].filter(Boolean)
          : translations.noSuggestionsAvailableAnnouncement,
      );
```

mrm007: Or another suggestion, a bit more verbose but I find it easier to read:

```js
    } else if (noSuggestionsMessage) {
      announcements.push(noSuggestionsMessage.text);

      if (noSuggestionsMessage.description) {
        announcements.push(noSuggestionsMessage.description);
      }
    } else {
      announcements.push(translations.noSuggestionsAvailableAnnouncement);
    }
```

michaeltaranto: This would result in an array inside an array, and not be "joined" as below. But i did miss the join happening further down so will simplify.

mrm007: Oops, rookie mistake. I was thinking of `.concat`'s behaviour which doesn't nest arrays.

PR #1152, opened 2022-09-27
TextLink: Improve screenshot tests

michaeltaranto:
Improving the visual testing of `TextLink` in advance of some clean ups. This will ensure no regressions slip through, but we really should have had these ages ago given their styling complexity.

The existing tests were a hangover from the shared docs and testing responsibilities, now the concerns are separate we can be more exhaustive.


askoufis (approved):
⛓️

PR #1153, opened 2022-09-28
Heading: Nested icons inherit text colour

michaeltaranto:
When using icons inside of a `Heading`, the default `tone` was always `neutral`, rather than inheriting the colour of the nearest component.

For example, when an icon was used inside of a `TextLink` within a `Heading`:
```jsx
<Heading level="1">
  Title with <TextLink>link <IconArrow /></TextLink>
</Heading>
// => Previously, IconArrow was the heading text colour
// `neutral`, now inherits the `link` colour.
```

or equally, when an icon was used inside of a `Secondary` component within a `Heading`:
```jsx
<Heading level="1">
  Title with <Secondary>secondary <IconArrow /></Secondary>
</Heading>
// => Previously, IconArrow was the heading text colour
// `neutral`, now inherits the `secondary` colour.
```


packages/braid-design-system/lib/components/Heading/Heading.tsx
```diff
@@ -18,36 +10,28 @@ const resolveDefaultComponent = {
...
+  weight?: keyof typeof typographyStyles.headingWeight;
   children: ReactNode;
-  level: HeadingLevel;
-  weight?: HeadingWeight;
-  align?: BoxProps['textAlign'];
-  component?: BoxProps['component'];
-  id?: string;
-  truncate?: boolean;
-  data?: DataAttributeMap;
 }
```
askoufis, line 17: 🤌

askoufis (approved):

packages/braid-design-system/lib/css/globalTypographyStyles.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export const globalTextStyle = ({
+  weight = 'regular',
+  size = 'standard',
+}: GlobalTextStyleProps = {}): StyleRule => ({
+  fontFamily: vars.fontFamily,
+  fontWeight: vars.textWeight[weight],
+  color: vars.foregroundColor.neutral,
+  ...responsiveStyle({
```
mrm007, line 18: Did you mean to use `typographyStyles.textSizeUntrimmed` here?

packages/braid-design-system/lib/css/globalTypographyStyles.ts
```diff
@@ -0,0 +1,52 @@
...
+  weight?: keyof typeof typographyStyles.fontWeight;
+  size?: keyof typeof typographyStyles.textSizeUntrimmed;
+}
+
+export const globalTextStyle = ({
+  weight = 'regular',
+  size = 'standard',
+}: GlobalTextStyleProps = {}): StyleRule => ({
+  fontFamily: vars.fontFamily,
+  fontWeight: vars.textWeight[weight],
```
mrm007, line 16: Were these supposed to be `typographyStyles.fontWeight` and `typographyStyles.fontWeight[weight]`?

packages/braid-design-system/lib/css/globalTypographyStyles.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export const globalHeadingStyle = ({
+  weight = 'regular',
+  level,
+}: GlobalHeadingProps): StyleRule => ({
+  fontFamily: vars.fontFamily,
+  fontWeight: vars.headingWeight[weight],
+  color: vars.foregroundColor.neutral,
+  ...responsiveStyle({
```
mrm007, line 42: Same questions as above.

packages/braid-design-system/lib/hooks/useIcon/index.ts
```diff
@@ -6,20 +6,22 @@ import { OptionalTitle } from '../../components/icons/SVGTypes';
...
 import buildDataAttributes, {
   DataAttributeMap,
 } from '../../components/private/buildDataAttributes';
+import * as typographyStyles from '../../css/typography.css';
 import * as styles from './icon.css';
 
-type IconSize = NonNullable<UseTextProps['size']> | 'fill';
+type IconSize =
+  | NonNullable<keyof typeof typographyStyles.textSizeUntrimmed>
+  | 'fill';
```
mrm007, line 18: I believe this still works as expected:

```suggestion
type IconSize = keyof typeof typographyStyles.textSizeUntrimmed | 'fill';
```

packages/braid-design-system/lib/components/Text/TextContext.ts
```diff
@@ -1,4 +1,4 @@
 import { createContext } from 'react';
-import { UseTextProps } from '../../hooks/typography';
+import type { TextStyleProps } from '../../css/typography';
 
-export const TextContext = createContext<UseTextProps | false>(false);
+export const TextContext = createContext<TextStyleProps | false>(false);
```
mrm007, line 4: My initial reaction was 🤨
After searching through the codebase I only found this instance where this pattern is used. Everywhere else it was

```ts
const SomethingContext = createContext<Something | null>(null);
```

packages/braid-design-system/lib/hooks/useIcon/index.ts
```diff
@@ -57,9 +59,11 @@ export default (
   const headingContext = useContext(HeadingContext);
   const inheritedTone =
```
mrm007, line 60: _(GitHub won't let me add a suggestion over deleted lines)_

With the `| null` change I suggested in `TextContext`, this can be simplified:

```ts
  const inheritedTone = textContext?.tone || 'neutral';
  const resolvedTone = typographyStyles.tone[tone || inheritedTone];
  const toneClass =
    tone || (!headingContext && !textContext?.tone) ? resolvedTone : undefined;
```

Unit tests and `tsc` passed locally.

mrm007 (commented):
Massive cleanup 👏

Just a few semi-nitpicky comments from me 👼

packages/braid-design-system/lib/components/Text/TextContext.ts
michaeltaranto: Interesting, I hear your point. We do use a boolean for context elsewhere (e.g. `HeadingContext`, `Modal`, etc.). But the way the code can be simplified with null coalescing probably justifies the change.

packages/braid-design-system/lib/css/globalTypographyStyles.ts
michaeltaranto: Deliberately not using sharing the internal typography classes here. Providing this to consumers as a data structure (StyleRule) allows consumers to control the selector, e.g.:

```ts
globalStyle(`${scope} p`, globalTextStyle(...));
```

mrm007: Gotcha, I saw the style rules looked like the ones in in `typographyStyles.fontWeight` and `typographyStyles.textSizeUntrimmed`.

`GlobalTextStyleProps` also references `keyof typeof typographyStyles.*` and the values are not used here, so to me it looked like there is an invisible coupling between the two (`globalTextStyle` and `typographyStyles`).

michaeltaranto: Ah now that I can update, they should be `keyof typeof vars.textWeight` etc. Updating...

mrm007 (approved):

PR #1155, opened 2022-09-29
Improve typography screenshots and gallery examples

michaeltaranto:
Improving screenshot testing and gallery examples for both `Heading` and `Text` components

### DIFF
--- a/packages/braid-design-system/lib/components/Text/Text.screenshots.tsx
+++ b/packages/braid-design-system/lib/components/Text/Text.screenshots.tsx
@@ -12,22 +12,6 @@ const Container = ({ children }: { children: ReactNode }) => (
   <div style={{ maxWidth: '300px' }}>{children}</div>
 );
 
-// type TextSize = keyof typeof textSizeUntrimmed;
-// const validateTextSizes = <T extends TextSize[]>(
-//   array: T & ([TextSize] extends [T[number]] ? unknown : 'Invalid'),
-// ) => array;
-// // Validate that all entries are included in this array.
-// // This array is being used to order the sort examples visually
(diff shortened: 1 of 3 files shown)


packages/braid-design-system/lib/components/Text/Text.screenshots.tsx
```diff
@@ -5,40 +5,70 @@ import { Text, Stack, Column, Columns, IconPositive } from '../';
...
 const Container = ({ children }: { children: ReactNode }) => (
   <div style={{ maxWidth: '300px' }}>{children}</div>
 );
 
+const textSizes = ['large', 'standard', 'small', 'xsmall'] as Array<
+  keyof typeof textSizeUntrimmed
+>;
+const textWeights = ['strong', 'medium', 'weak'] as Array<
+  keyof typeof fontWeight
+>;
```
askoufis, line 20: I'm not sure how much value these typecasts are providing. The `textSizeUntrimmed` object could only contain `large` as a key, and you'd end up with an array of type `"large"[]`, which is not the correct type, but would still pass type-checking.

You're already passing these text sizes into the `size` prop of `Text` anyway, which should catch any bad keys sneaking through I feel like `as const` would work just fine here, and won't result in a potentially incorrect type.
```suggestion
const textSizes = ['large', 'standard', 'small', 'xsmall'] as const;
const textWeights = ['strong', 'medium', 'weak'] as const;
```

michaeltaranto: The intent is that if any additional sizes/weights were added in the future this would break and require you to add the new key in the correct position within the array. Realising that this wouldnt work as intended though.

Only reason im creating this array is to control the visual order of the variants.

askoufis (approved):

PR #1166, opened 2022-10-19
Iconongraphy: Uplift documentation

michaeltaranto:
Add new top level documentation for the icon suite, better demoing their api and usage.


askoufis (approved):
👁️ 🌽 🦆 s

PR #1168, opened 2022-10-20
Data prop standardisation and validation

michaeltaranto:
- Add support for `data` prop to Box for consistency with all other components. This is in addition to the supporting the native syntax as Box is the only component that exposes the native HTML api.
- Add support for `data` prop to Toast
- Document custom data attribute syntax on the site.
- Provide dev time validation/warnings to consumers when trying to pass the native format to components that do not support it, as TypeScript does not validate kebab cased properties. This will prevent silent failures were attributes are seeming provided but not applied.

Screen Shot 2022-10-21 at 10 06 13 am


packages/braid-design-system/lib/components/Box/Box.docs.tsx
```diff
@@ -127,6 +128,39 @@ const docs: ComponentDocs = {
...
+          </Notice>
+          <Code playroom={false}>
+            {
+              source(
+                <Box
+                  data={{
+                    testid: 'customIdentifier',
+                  }}
+                >
+                  {/* Becomes: data-testid="customIdentifier" */}
```
askoufis, line 155: Feels a bit weird that this comment is inside the component. Would it make more sense one line higher so it's more obvious that it's tied to the `data` prop?

packages/braid-design-system/lib/components/private/buildDataAttributes.ts
```diff
@@ -1,6 +1,68 @@
...
+        }
+
+        return `="${v}"`;
+      };
+
+      if (dataAttrs.length > 0) {
+        // eslint-disable-next-line no-console
+        console.warn(
+          dedent`
+          Braid components do not support the native data attribute format. Use the “data” prop instead.
```
askoufis, line 33: Would it be worth linking to some docs here too? https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

packages/braid-design-system/lib/components/private/buildDataAttributes.ts
```diff
@@ -1,6 +1,68 @@
...
+        if (typeof v === 'boolean') {
+          return v ? '' : `={false}`;
+        }
+
+        if (typeof v === 'number') {
+          return `={${v}}`;
+        }
+
+        return `="${v}"`;
+      };
```
askoufis, line 27: Could be moved to the top-level scope since it's a pure function, but it's only run in dev so I'm not fussed. The dev-time perf tho...

askoufis (commented):
Great stuff, just some small comments

packages/braid-design-system/lib/components/private/buildDataAttributes.ts
michaeltaranto: Yep I typically would have, but left this inside so it gets stripped out of the production bundle along with the entire if block.

askoufis: Oh of course, that makes sense.

packages/braid-design-system/lib/components/Box/Box.docs.tsx
michaeltaranto: Yeah there was some formatting constraints in the code block, but will throw around some alternatives.

askoufis (approved):

PR #1174, opened 2022-10-26
TextField: Add `inputMode` and `step` support

michaeltaranto:
Provide support for the native `inputMode` and `step` attributes.

The `inputMode` will also be defaulted based on the specified `type`. For example: `<TextField type="number" />` will default the `inputMode` to `numeric`.

**EXAMPLE USAGE:**
```jsx
<TextField inputMode="numeric" step=".01" />
```


askoufis (approved):

packages/braid-design-system/lib/components/TextField/TextField.docs.tsx
```diff
@@ -10,9 +10,10 @@ import {
   IconHelp,
   Stack,
   Heading,
+  IconLanguage,
+  TextDropdown,
 } from '../';
 import source from '../../utils/source.macro';
-import { IconLanguage } from '../icons';
```
mrm007, line 15: 👌

PR #1180, opened 2022-11-03
Tab, Tabs: Updated visual design

michaeltaranto:
The appearance of a `Tab` has been updated. Changes include:
- Inactive tab uses `regular` text weight
- Hover state of inactive tab toggles `neutral` tone instead of underline
- Active tab indicator underlines content only, without the horizontal gutter
- Active tab indicator rounds top radii
- The `minimal` divider under `Tabs` underlines content only, without the horizontal gutter

Before: (first tab active, third tab hovered)
Screen Shot 2022-11-03 at 11 41 47 am

After: (first tab active, third tab hovered)
Screen Shot 2022-11-03 at 11 41 39 am

Note: Also fixes hot reload issue with `TabPanel` in dev mode.

### DIFF
--- a/packages/braid-design-system/lib/components/Tabs/Tabs.tsx
+++ b/packages/braid-design-system/lib/components/Tabs/Tabs.tsx
@@ -1,5 +1,4 @@
 import React, {
-  Children,
   useContext,
   useEffect,
   useRef,
@@ -8,6 +7,7 @@ import React, {
 } from 'react';
 import assert from 'assert';
 import flattenChildren from 'react-keyed-flatten-children';
+import { assignInlineVars } from '@vanilla-extract/dynamic';
 import { Box } from '../Box/Box';
 import { Divider } from '../Divider/Divider';
 import type { ResponsiveSpace } from '../../css/atoms/atoms';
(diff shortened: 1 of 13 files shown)


packages/braid-design-system/lib/components/Tabs/Tabs.tsx
```diff
@@ -58,8 +58,9 @@ export const Tabs = (props: TabsProps) => {
 
   const { dispatch, a11y } = tabsContext;
   const tabItems: Array<string | number> = [];
+  const childTabs = flattenChildren(children);
 
-  const tabs = Children.map(flattenChildren(children), (tab, index) => {
+  const tabs = Children.map(childTabs, (tab, index) => {
```
mrm007, line 63: Maybe I'm missing something, but I don't see the need for `Children.map` any more. Wouldn't this work OOTB, without `Children.map`?

```suggestion
  const tabs = childTabs.map((tab, index) => {
```

michaeltaranto: Yeah totally, previously would've been mapping over `children`, then it got flattened but not converted to a standard map.

Given my change to extract it to a variable, it makes sense to update it now. Good shout.

mrm007 (dismissed):

pnpm-lock.yaml
```diff
@@ -6616,6 +6624,44 @@ packages:
     dependencies:
       fill-range: 7.0.1
 
+  /braid-design-system/31.22.1_3cyl5sbjrbodepoesuincwdfla:
```
askoufis, line 6627: Confused as to why this is showing up in the lock file all of a sudden.

askoufis (commented):
Looks like there are some json files that weren't mean to be checked in.

askoufis (approved):

mrm007 (approved):

PR #1187, opened 2022-11-16
Refactor narrowing of `TextDropdownOption` to be compatible with typescript 4.9

askoufis:
On typescript versions >= 4.9, the `TextDropdown` component fails to typecheck [here](https://github.com/seek-oss/braid-design-system/blob/bcb987d486f946b43a95a47261d667bdf3784d35/packages/braid-design-system/lib/components/TextDropdown/TextDropdown.tsx#L34) with the error:

```
Type 'Value' is not assignable to type 'object'
```

This error is surfacing in typescript 4.9 for two reasons. Firstly, the narrowing conditions `typeof option !== 'string' && typeof option !== 'number'` were never strict enough to rule out the possibility that `option` could be a primitive value, so by the time the `in` operator was being executed, the type of `option` hadn't changed at all (pre typescript 4.9).

Secondly, the [changes to the `in` operator](https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/#unlisted-property-narrowing-with-the-in-operator) in typescript 4.9, in particular this change:
> ensuring that the left side is assignable to the type `string | number | symbol`, and the right side is assignable to `object`

resulted in a type error because `in` was being used on a RHS value (`option`) that wasn't yet of type `object`. I believe typescript 4.9 is a bit smarter about narrowing in general, as the intermediate type _does_ change, but it can't quite narrow the type all the way to `object` at this point.

The solution I went with was to flip the narrowing to explicitly check for an object (and perform a null check because `typeof null === 'object'`) first, and then narrow the type further. This change is backwards compatible with typescript <4.9.

I also updated the version of typescript within the repo to 4.9.3.

### DIFF
--- a/.changeset/shaggy-tools-trade.md
+++ b/.changeset/shaggy-tools-trade.md
@@ -3,11 +3,8 @@
 ---
 
 ---
-
 updated:
-
-- TextDropdown
-
+  - TextDropdown
 ---
 
 Fix a type error affecting consumers on TypeScript versions >=4.9.0
--- a/.changeset/shaggy-tools-trade.md
+++ b/.changeset/shaggy-tools-trade.md
@@ -10,4 +10,4 @@ updated:
 
 ---
 
-Fix a type issue affecting TypeScript versions >=4.9.0
+Fix a type error affectin consumers on TypeScript versions >=4.9.0
--- a/.changeset/shaggy-tools-trade.md
+++ b/.changeset/shaggy-tools-trade.md
@@ -1,0 +1,13 @@
+---
+'braid-design-system': patch
+---
+
+---
+
+updated:
+
+- TextDropdown
+
+---
+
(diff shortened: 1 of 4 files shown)


benjervis (approved):
👏

.changeset/shaggy-tools-trade.md
```diff
@@ -0,0 +1,13 @@
+---
+'braid-design-system': patch
+---
+
+---
+
+updated:
+
+- TextDropdown
+
```
mrm007, line 10:
```suggestion
updated:
  - TextDropdown
```

helping

mrm007 (approved):

.changeset/shaggy-tools-trade.md
askoufis: Prettier formats it like this, I think because it doesn't understand the double frontmatter.

PR #1190, opened 2022-11-17
Create shareable Playroom config

mrm007:
Metropolis uses Braid's internals and it would be very difficult to create a stand-alone package. See below for a few examples:

- https://github.com/seek-oss/braid-design-system/blob/35595823adc3d0b85d504b4a232439e4c3c00f50/packages/braid-design-system/lib/playroom/FrameComponent.tsx#L12
- https://github.com/seek-oss/braid-design-system/blob/2ba61622cc3ec9bfb1ec2082002171ddfeddc143/packages/braid-design-system/lib/playroom/components.ts#L4
- https://github.com/seek-oss/braid-design-system/blob/35595823adc3d0b85d504b4a232439e4c3c00f50/packages/braid-design-system/lib/playroom/snippets.ts#L4
- https://github.com/seek-oss/braid-design-system/blob/35595823adc3d0b85d504b4a232439e4c3c00f50/packages/braid-design-system/lib/playroom/useScope.ts#L7


askoufis (commented):
Should this be an entrypoint or a package? The risk with making it an entrypoint is that it'd be quite easy to auto-import the playroom version of a component. Making it a package is more effort though.

mrm007:
Making it a package is the goal, but it would be the odd one out in the repo's current state. `css`, `themes`, `reset` etc. are not packages, so I tried to stay consistent with that pattern.

In a crackle world it would 💯 be a package.

askoufis:
I worry about exposing this API as public though, even if it's temporary. However, making it a package would require the source macro to be a package too I guess. Also `useScope` and `FrameComponent` use internal braid APIs anyway. So do the playroom components. I feel like I've re-discovering things I wrote down a month ago.

packages/braid-design-system/playroom/components.ts
```diff
@@ -0,0 +1,15 @@
+import dedent from 'dedent';
+
+declare const global: any;
+if (global?.__IS_PLAYROOM_ENVIRONMENT__ !== 'clearly') {
+  throw new Error(dedent`
+    Playroom components were imported instead of Braid components.
+    ‎
```
mrm007, line 7: There are invisible characters here to prevent VS Code from trimming unused white space. Because then `dedent` will collapse subsequent empty lines and the message will be too compacted.

gremlins

Without the ["gremlins"](https://marketplace.visualstudio.com/items?itemName=nhoizey.gremlins) it would look like this:

```
Playroom components were imported instead of Braid components.
Make sure to import from the correct entry point.
-import { Component } from 'braid-design-system/playroom/components';
+import { Component } from 'braid-design-system';
```

packages/braid-design-system/lib/css/reset/resetTracker.ts
```diff
@@ -10,12 +10,12 @@ export const ensureResetImported = () => {
   if (!resetImported) {
     throw new Error(dedent`
       Braid components imported before reset.
-
-      Make sure to import the Braid reset module before importing any components. 
-      This ensures the CSS reset does not override the component styles. 
-    
+      ‎
```
mrm007, line 13: Same thing here

gremlins

site/sku.config.js
```diff
@@ -3,7 +3,7 @@ const routes = require('./sku.routes.js');
 const CircularDependencyPlugin = require('circular-dependency-plugin');
 const browserslist = require('../browserslist');
 
-const isGitHubPages = Boolean(process.env.IS_GITHUB_PAGES);
+const isGitHubPages = process.env.IS_GITHUB_PAGES === 'true';
```
mrm007, line 6: `Boolen("false") === true` 😬

site/sku.config.js
```diff
@@ -3,7 +3,7 @@ const routes = require('./sku.routes.js');
 const CircularDependencyPlugin = require('circular-dependency-plugin');
 const browserslist = require('../browserslist');
 
-const isGitHubPages = Boolean(process.env.IS_GITHUB_PAGES);
```
mrm007, line 6: `Boolen("false") === true` 😬

packages/braid-design-system/package.json
```diff
@@ -12,6 +12,7 @@
     "lib/css/reset/**/*",
     "themes/**/*",
     "reset/**/*",
+    "playroom/**/*",
```
mrm007, line 15: This ensures the error is thrown.

packages/braid-design-system/playroom/components.ts
```diff
@@ -0,0 +1,15 @@
...
+    Make sure to import from the correct entry point.
```
michaeltaranto, line 8:
```suggestion
    These components must not be used in production. Please import from the top level Braid entry point:
```

packages/braid-design-system/playroom/components.ts
```diff
@@ -0,0 +1,15 @@
+import dedent from 'dedent';
+
+declare const global: any;
+if (global?.__IS_PLAYROOM_ENVIRONMENT__ !== 'clearly') {
+  throw new Error(dedent`
+    Playroom components were imported instead of Braid components.
```
michaeltaranto, line 6:
```suggestion
    Playroom prototyping components are being imported instead of Braid components.
```

michaeltaranto (approved):
Nice work. Once tested via a snapshot deployment I'm happy get this in 👏

.changeset/brave-items-invite.md
```diff
@@ -0,0 +1,23 @@
+---
+'braid-design-system': minor
+---
+
+---
+new:
+  - Playroom
+---
```
michaeltaranto, line 8: The additional front matter is not required for playroom as its not a documented entry point.
```suggestion
```

mrm007:
Works [in Metropolis](https://metropolis.ssod.skinfra.xyz/en/?v=branch-use-playroom-config) 👌

askoufis (approved):

PR #1194, opened 2022-11-24
Update action versions and sku, preserve newlines in codemod

askoufis:
- Use looser action version for `pnpm/action-setup`. This addresses [a deprecation warning about using node 12](https://github.com/seek-oss/braid-design-system/actions/runs/3537904904).
- Use a major version tag instead of the `main` branch for `actions/checkout` and `actions/setup-node`
- Remove explicit `node-version` since `actions/setup-node` now honours the node version specified in our volta config
- Update sku to `11.6.2`
- As mentioned in comments below, likely due to changes in a recent babel version, the codemod is removing all newlines from transformed code. This isn't a huge deal right now, since most teams are likely already on a recent version of braid so they have no reason to use `braid-upgrade`, but future upgrades will be affected by this issue. This fix is to set [`retainLines`](https://babeljs.io/docs/en/options#retainlines) on the codemod, which appears to preserve newlines in all test cases.


.github/workflows/preview-site.yml
```diff
@@ -13,15 +13,14 @@ jobs:
...
-        uses: pnpm/action-setup@v2.2.2
+        uses: pnpm/action-setup@v2
 
-      - name: Set up Node.js 16.x
-        uses: actions/setup-node@main
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
         with:
-          node-version: 16.x
           cache: 'pnpm'
```
mrm007, line 24: We still need to [specify `node-version-file`](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file) otherwise it will use the default, which is [16.18.1](https://github.com/seek-oss/braid-design-system/actions/runs/3544167382/jobs/5951211572#step:4:2) and we want [16.13.0](https://github.com/seek-oss/braid-design-system/blob/action-versions/package.json#L77-L79).

```suggestion
          cache: 'pnpm'
          node-version-file: package.json
```

askoufis: Oh yes, you're right.

package.json
```diff
@@ -81,7 +81,7 @@
   "pnpm": {
     "patchedDependencies": {
       "panzoom@9.4.2": "patches/panzoom@9.4.2.patch",
-      "sku@11.6.0": "patches/sku@11.6.0.patch"
+      "sku@11.6.2": "patches/sku@11.6.2.patch"
```
mrm007, line 84: Should we do something about this in sku?

pnpm-lock.yaml
```diff
@@ -869,6 +920,20 @@ packages:
     transitivePeerDependencies:
       - supports-color
 
+  /@babel/helper-remap-async-to-generator/7.18.9_@babel+core@7.19.6:
```
mrm007, line 923: Looks like we have quite a few duplicates like this one. Looks like [there's a `pnpm-deduplicate`](https://www.npmjs.com/package/pnpm-deduplicate) (similar to the Yarn one) but I've never personally used it.

mrm007 (approved):
Nice clean-up 🧹

askoufis:
Codemod tests are failing, newlines are not being preserved in test cases. Babel in yarn lock is now at `7.20.0`, and I think that might have something to do with it. https://babeljs.io/blog/2022/10/27/7.20.0

If all newlines are no longer preserved, that could be bad for consumers as it will result in lots of undesirable formatting changes. [We run the codemod output through prettier](https://github.com/seek-oss/braid-design-system/blob/71f5b09fe1ff45ed0765017e4f45d46a02805787/packages/codemod/src/codemod.ts#L132), but I don't think that will add back in newlines that were removed.

Apparently [the solutions](https://github.com/babel/babel/issues/2226#issuecomment-267142901) to this are:
1. Set `retainLines` to `true`. Supposedly has issues, thought it seems to work fine with me.
2. Use sourcemaps?

mrm007:
@askoufis good sleuthing 🕵️

[Here's a branch](https://github.com/seek-oss/braid-design-system/compare/action-versions...action-versions-retainLines) with the necessary changes to retain lines (and the changes to enable usage of `satisfies` from TypeScript 4.9). Feel free to merge/remix/ignore.

PR #1201, opened 2023-01-10
Add a test for side-effects

mrm007:
This allows as to see what side-effects we need to keep when we ship compiled Braid (v32).


site/package.json
```diff
@@ -55,7 +55,7 @@
     "react-use": "^17.4.0",
     "recoil": "^0.7.2",
     "sku": "11.7.0",
-    "webpack": "^5.0.0"
+    "webpack": "^5.72.1"
```
mrm007, line 58: This is the same version already in the lockfile.

packages/consumer/__snapshots__/build.test.ts.snap
```diff
@@ -0,0 +1,108 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`build side-effects 1`] = `
```
mrm007, line 3: Having these as reference will be of great help in Crackle https://github.com/seek-oss/crackle/blob/6c0f1bc76167a3a8c418bffc30e82879395b4ce7/packages/core/src/package-utils/bundle.ts#L46-L59

packages/consumer/build.test.ts
```diff
@@ -0,0 +1,42 @@
...
+import path from 'path';
+import glob from 'fast-glob';
+import webpack from 'webpack';
+import braidPkg from 'braid-design-system/package.json'; // eslint-disable-line no-restricted-imports
+
+const { SideEffectsFlagPlugin } = webpack.optimize;
+
+describe('build', () => {
+  const cache = new Map();
+  const checkSideEffectts = (filePath: string): boolean => {
```
askoufis, line 11:
```suggestion
  const checkSideEffects = (filePath: string): boolean => {
```

askoufis:
Thoughts on moving this to a `fixtures` folder rather than putting it inside `packages`? I know it's private and will never be published anyway, but in terms of alignment with other repos (crackle, VE), I'd prefer it to be in `fixtures`, if it's not too much effort to do so.

mrm007:
`fixtures` is a good idea 👍

askoufis (approved):

PR #1208, opened 2023-02-01
Reorganize tasks

mrm007:


packages/braid-design-system/project.json
```diff
@@ -1,25 +1,54 @@
...
+      "inputs": ["^default"],
+      "outputs": ["{projectRoot}/codemod/dist"],
+      "executor": "nx:run-script",
+      "options": {
+        "script": "codemod"
+      }
+    },
+    "generate:icons": {
+      "inputs": ["{projectRoot}/icons/*.svg", "{projectRoot}/scripts"],
+      "outputs": ["{projectRoot}/src/**/*Svg.tsx"],
```
askoufis, line 25: I think it's worth scoping this to specifically the `icons` folder, just in case there's ever some other file that ends with `Svg.tsx`.

site/project.json
```diff
@@ -1,42 +1,69 @@
...
       "executor": "nx:run-script",
       "options": {
-        "script": "tsc"
+        "script": "lint:tsc"
       }
-    }
-  },
-  "implicitDependencies": ["braid-design-system"]
+    },
+    "format": {}
```
askoufis, line 67: Confused about this change in the most recent commit. Won't this result in format scripts not running?

michaeltaranto (approved):

site/project.json
mrm007: I encourage you to verify it locally.

This is a not-so-well documented feature that is briefly mentioned in passing [in the docs](https://nx.dev/reference/nx-json#target-defaults). If you look at Nx's own configuration you'll notice the same pattern.

Top-level `nx.json`: https://github.com/nrwl/nx/blob/master/nx.json#L88
Project config: https://github.com/nrwl/nx/blob/master/packages/next/project.json#L59

askoufis (approved):

PR #1219, opened 2023-02-10
Publish readme to npm with Braid

michaeltaranto:
Since restructuring the project to a monorepo we have no longer had a readme visible with the published artefact in npm.

This PR adds a little script to copy the root level readme into the braid package before publishing.


package.json
```diff
@@ -21,7 +21,7 @@
     "deploy": "node scripts/deploy.js",
     "post-commit-status": "node scripts/postCommitStatus.js",
     "release": "pnpm prepare-publish && changeset publish && pnpm build:site && pnpm run deploy",
-    "prepare-publish": "pnpm build",
+    "prepare-publish": "pnpm build && tsx scripts/copyReadme.ts",
```
mrm007, line 24: My only nitpick is to name the file `copyReadme.cts` to signal that it's a CommonJS module. `__dirname` is not available in ESM but these loaders (`tsx` and co.) usually paper over it with shims.

mrm007 (approved):

PR #1222, opened 2023-02-14
Drop support for React 16

mrm007:
And use latest Crackle


.changeset/breezy-fireants-yawn.md
```diff
@@ -0,0 +1,5 @@
+---
+'braid-design-system': major
+---
+
+Remove React 16 from `peerDependencies`
```
michaeltaranto, line 5:
```suggestion
Drop support for React 16.

As part of the work to pre-compile Braid, it is now a requirement that consumers are on React 17 or later. For more information on the differences between v16 and v17, [see the React blog].

[see the React blog]: https://reactjs.org/blog/2020/10/20/react-v17.html
```

site/src/types.d.ts
```diff
@@ -21,10 +21,10 @@ export interface RenderContext {
   helmetContext: HelmetData['context'] | Record<string, never>;
 }
 
-export interface Page extends RouteProps {
+export type Page = RouteProps & {
   title: string;
   badge?: 'New';
-}
+};
```
michaeltaranto, line 27: Just curious, was this a required change?

mrm007: Let's see what happens in CI, without this change 😊

mrm007: Looks like the build is failing now. https://github.com/seek-oss/braid-design-system/actions/runs/4170111393/jobs/7218757053#step:7:354

michaeltaranto: Sure, was curious as to why now? Is there a TS update in here, or something else. Just seemed a little random thats all

mrm007: I've deduplicated dependencies which upgraded `react-router` to latest which has the new types.

mrm007: Old: https://unpkg.com/browse/react-router@6.3.0/lib/components.d.ts#L42
New: https://unpkg.com/browse/react-router@6.4.3/dist/lib/components.d.ts#L80

packages/braid-design-system/src/lib/utils/flattenChildren.ts
```diff
@@ -0,0 +1,36 @@
+// Copied and modified from https://github.com/grrowl/react-keyed-flatten-children/blob/0b4fd6dad491905e73551c9d4496198f6b87eb41/index.ts
```
mrm007, line 1: Pulling in this function to [fix the build](https://buildkite.com/seek/metropolis/builds/13647#018623e9-978e-4f0e-bedf-4dd9b9bb73be/6-139) in consuming apps. I've copied the function and made slight tweaks to satisfy our ESLint config.

Another option would have been to move `react-keyed-flatten-children` which would have inlined the function, **as well as all of `react-is`**. `react-keyed-flatten-children` depends on `react-is@16` which is
1. a lot of dead code
2. not aligned with our version of React

.npmrc
```diff
@@ -0,0 +1 @@
+strict-peer-dependencies=false
```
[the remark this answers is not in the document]
mrm007: A long long time ago https://github.com/seek-oss/braid-design-system/commit/f21ec92d7603ea10e62cf5b77aaf566031f13b22#diff-e813096d69c49812e40e00be9ac8fa14d77f0ec1f97ab4c45cb096b3bedd35de

askoufis (approved):

packages/braid-design-system/package.json
```diff
@@ -171,6 +171,7 @@
     "@types/node": "^18.13.0",
     "@types/react": "^18.0.21",
     "@types/react-dom": "^18.0.6",
+    "@types/react-is": "^17.0.3",
```
michaeltaranto, line 174: Confirming, is this version correct? Not sure if the definitely typed package is just out of step with version numbers

michaeltaranto (approved):

packages/braid-design-system/package.json
mrm007: Yep, that's the most recent [version](https://www.npmjs.com/package/@types/react-is).

PR #1223, opened 2023-02-16
Add a script to generate the Playroom snippets index file

mrm007:
The task graphs now look like this:

### `generate`

graph-generate

### `build`
graph-build


packages/braid-design-system/src/lib/utils/source.macro.js
```diff
@@ -1,7 +1,8 @@
 const { createMacro } = require('babel-plugin-macros');
 const { default: generate } = require('@babel/generator');
 
-module.exports = createMacro(({ babel: { types: t }, references }) => {
+exports.__esModule = true;
+exports.default = createMacro(({ babel: { types: t }, references }) => {
```
mrm007, line 5: I'm having some trouble with this bad boy. It either works in Crackle or in Webpack,  but not both 😩

site/project.json
```diff
@@ -45,7 +45,7 @@
       }
     },
     "generate": {
-      "dependsOn": ["generate:component-docs", "^generate:icons"],
+      "dependsOn": ["generate:component-docs", "^generate"],
```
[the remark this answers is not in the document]
mrm007: This is `site:generate` which runs `site:generate:component-docs` and `braid-design-system:generate`.
`braid-design-system` has no knowledge of component docs.

If you have a look at the graphs above, it should be more clear.

packages/braid-design-system/src/lib/utils/source.macro.js
mrm007: Solved in [`4f9edc6` (#1223)](https://github.com/seek-oss/braid-design-system/pull/1223/commits/4f9edc66ac4b9185e0035ab98d86cbd486f2acd3)

packages/braid-design-system/src/lib/utils/source.macro.cjs
```diff
@@ -1,7 +1,8 @@
 const { createMacro } = require('babel-plugin-macros');
 const { default: generate } = require('@babel/generator');
```
mrm007, line 2: Right now, because these are `devDependencies` the source gets inlined and the compiled code is 35,000 LOC. If we're to have the snippets be loaded from Metropolis we'll likely have to make these `dependencies` so they don't get bundled up. Thoughts?

packages/braid-design-system/scripts/generateSnippets.ts
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable no-console */
```
mrm007, line 1: This file should be `.cts` but I discovered our ESLint config needs to be updated to lint files with `.cts`, `.cjs`, `.mjs` and `.mts` extensions.

packages/braid-design-system/src/lib/utils/source.macro.cjs
mrm007: Solved in [`9187af9` (#1223)](https://github.com/seek-oss/braid-design-system/pull/1223/commits/9187af9c0d756720e3ee878348c727ad1b2f52f4)

packages/braid-design-system/scripts/generateSnippets.ts
```diff
@@ -0,0 +1,93 @@
...
+        await fs.writeFile(transformedSnippetPath, transformedSnippetCode);
+        return transformedSnippetPath;
+      }),
+    )
+  )
+    .map((snippetsFile) => removeExtension(relativeTo(path.dirname(snippetsIndexFile), snippetsFile)))
+    .sort()
+    .map((relativePath) => ({
+      componentName: path.basename(relativePath),
+      importName: `snippets$${path.basename(relativePath)}`,
```
michaeltaranto, line 61: We probably dont need to name this anymore, can just use the `componentName`, which would make the output file:
```ts
import { snippets as Accordion } from './snippets/Accordion';
```

packages/braid-design-system/project.json
```diff
@@ -20,17 +21,32 @@
...
+        "{projectRoot}/src/**/components/icons/index.ts"
+      ],
       "executor": "nx:run-script",
       "options": {
         "script": "generate:icons"
       }
     },
+    "generate:snippets": {
+      "inputs": ["{projectRoot}/src/**/*.snippets.tsx", "scripts"],
+      "outputs": ["{projectRoot}/src/**/playroom/snippets.ts"],
```
michaeltaranto, line 41: I think this should take the new snippets directory into account as an output, e.g. `"{projectRoot}/src/**/playroom/snippets/*.ts"`

michaeltaranto (approved):

PR #1235, opened 2023-03-02
Upgrade playroom

benjervis:
We need to get these vital efficiency gains out asap

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -153,7 +153,7 @@ importers:
       html-validate: ^7.1.1
       is-mobile: ^2.2.2
       lodash: ^4.17.21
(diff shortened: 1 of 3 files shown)


pnpm-lock.yaml
```diff
@@ -15342,6 +15342,81 @@ packages:
       - uglify-js
       - utf-8-validate
       - webpack-cli
+    dev: true
+
+  /playroom/0.29.0_biqbaboplfbrettd7655fr4n2y:
```
mrm007, line 15347: Looks like `/playroom/0.28.2` is also in the lockfile (line 15273)

askoufis (approved):

PR #1240, opened 2023-03-06
Disclosure: Add weight and inline content support

michaeltaranto:
A few features and fixes for the `Disclosure` component:

### Add `weight` support

Provides support for reducing the visual weight of the Disclosures call to action. Follows the same contextual styling rules as `TextLink`.

**EXAMPLE USAGE:**
```jsx
<Disclosure weight="weak">
  ...
</Disclosure>
```

### Add inline content support
Provides support for using a Disclosure in the context of sentence, by allowing it to be nested within a typographic component, i.e. `Text`. This enables the call to action to sit inline with other copy.

All size and weight properties will inherit from the parent component.

**EXAMPLE USAGE:**
```jsx
<Text>
  Preceeding text content that is followed by a Disclosure.
  <Disclosure expandLabel="Read more">
    ...
  </Disclosure>
</Text>
```

### Ensure chevron does not wrap alone

Fixes the scenario where based on copy length and container size, the chevron could wrap independently of the rest of the label. By using a zero-width, non-breaking space, the chevron will now wrap together with the last word of the label.


fixtures/consumer/__snapshots__/build.test.ts.snap
```diff
@@ -12,6 +12,20 @@ exports[`build side-effects from dist 1`] = `
   "dist/side-effects/lib/css/reset/index.mjs",
   "dist/side-effects/lib/css/reset/resetTracker.cjs",
   "dist/side-effects/lib/css/reset/resetTracker.mjs",
+  "dist/side-effects/lib/playroom/FrameComponent.cjs",
```
mrm007, line 15: Looks like you need to do a `pnpm install` locally and rerun the test.

.changeset/eleven-bees-retire.md
```diff
@@ -0,0 +1,19 @@
...
+---
+
+---
+updated:
+  - Disclosure
+---
+
+**Disclosure:** Add `weight` support
+
+Provides support for reducing the visual weight of the Disclosures call to action. Follows the same contextual styling rules as `TextLink`.
```
askoufis, line 12:
```suggestion
Provides support for reducing the visual weight of the Disclosure's call to action. Follows the same contextual styling rules as `TextLink`.
```

.changeset/sour-fans-type.md
```diff
@@ -0,0 +1,24 @@
...
+**Disclosure:** Add inline content support
+
+Provides support for using a Disclosure in the context of sentence, by allowing it to be nested within a typographic component, i.e. `Text`. This enables the call to action to sit inline with other copy.
+
+All size and weight properties will inherit from the parent component.
+
+**EXAMPLE USAGE:**
+```jsx
+<Text>
+  Preceeding text content that is followed by a Disclosure.
```
askoufis, line 19:
```suggestion
  Preceding text content that is followed by a Disclosure.
```

fixtures/consumer/__snapshots__/build.test.ts.snap
michaeltaranto: Interesting. I did that before opening the PR and it still wanted to update the snapshot, so I thought I'd see what CI thought. Let me play around more 🤔

mrm007: Correction: [`71efb30` (#1240)](https://github.com/seek-oss/braid-design-system/pull/1240/commits/71efb30dc981d44bb69c9ec6a975d1afb3bf9556)

packages/braid-design-system/src/lib/components/Disclosure/Disclosure.tsx
```diff
@@ -40,6 +47,10 @@ export const Disclosure = ({
     "'collapseLabel' must be a string",
   );
 
+  const textContext = useContext(TextContext);
+  const headingContext = useContext(HeadingContext);
+  const isInline = textContext || headingContext;
```
mrm007, line 52: Just to be extra safe, since `textContext` can be an object:

```suggestion
  const isInline = Boolean(textContext || headingContext);
```

packages/braid-design-system/src/lib/components/Disclosure/Disclosure.screenshots.tsx
```diff
@@ -1,58 +1,187 @@
-import React, { useState } from 'react';
+import type { ReactNode } from 'react';
+import React from 'react';
 import type { ComponentScreenshot } from 'site/types';
 import { Disclosure, Text } from '../';
+import { Box } from '../Box/Box';
+
+const Container = ({ children }: { children: ReactNode }) => (
+  <Box style={{ maxWidth: 250 }}>{children}</Box>
+);
```
mrm007, line 9: Was this intentional? The [screenshots on large breakpoints][wide] are now clamped at 250px, but the 320px ones [are also clamped][320].

[wide]: https://www.chromatic.com/component?appId=5bd0cd3e0cc977002443f476&csfId=disclosure--apac&buildNumber=4944&k=6405b97df693085eb45086ec-320-interactive-true&h=17&b=-4
[320]: https://www.chromatic.com/component?appId=5bd0cd3e0cc977002443f476&csfId=disclosure--apac&buildNumber=4944&k=6405b97df693085eb45086ec-320-snapshot-true&h=18&b=-5

michaeltaranto: They are [only screenshot at 320] (see [screenshots]) and was done to deliberately exacerbate wrapping. At this size previously the chevron would wrap separately to the label, which was the bug that was fixed too.

[only screenshot at 320]: https://github.com/seek-oss/braid-design-system/pull/1240/files/e318dd33ad7831c668d2bc6fb3cf7770eabd7986#diff-e703586717ac85ca074180a6023ffdfa8e999e54cecfe4ca46a0f0141fda9a66R12
[screenshots]: https://www.chromatic.com/test?appId=5bd0cd3e0cc977002443f476&id=640579ae8bdb80c75963cd7b

mrm007 (approved):

.changeset/sour-fans-type.md
```diff
@@ -0,0 +1,24 @@
...
+---
+
+---
+updated:
+  - Disclosure
+---
+
+**Disclosure:** Add inline content support
+
+Provides support for using a Disclosure in the context of sentence, by allowing it to be nested within a typographic component, i.e. `Text`. This enables the call to action to sit inline with other copy.
```
askoufis, line 12: I don't think the comma is needed.

```suggestion
Provides support for using a Disclosure in the context of a sentence by allowing it to be nested within a typographic component, i.e. `Text`. This enables the call to action to sit inline with other copy.
```

Also more broadly, what does `in the context of a sentence` even mean? It honestly sounds more complicated than it needs to be. Feel like it could be simplified to `inline with other content` or something.

michaeltaranto: I agree actually, although to avoid confusion with the behaviour of our `Inline` layout component I'm thinking something more like "within a sentence" or similar.
```suggestion
Provides support for using a Disclosure within a sentence by allowing it to be nested within a typographic component, i.e. `Text`.
```

askoufis (approved):

PR #1242, opened 2023-03-08
Button, TextLink: Add support for trailing icons

michaeltaranto:
Provide support for choosing the position of the `icon` slot within a `Button` or `TextLink`.

By default, the `iconPosition` will be `leading` the label, but optionally, can be set to `trailing`.

**EXAMPLE USAGE:**
```jsx
<Button
  icon={<IconArrow direction="right" />}
  iconPosition="trailing"
>
  Next
</Button>
```


packages/braid-design-system/src/lib/components/Button/Button.tsx
```diff
@@ -370,6 +372,19 @@ export const ButtonText = ({
         ) : null}
         {children}
         {loading ? <ButtonLoader /> : null}
+        {!loading && icon && iconPosition === 'trailing' ? (
```
mrm007, line 375: This is what the example looks like in Firefox:

trailing

With `white-space: nowrap` on the immediate parent:

nowrap

packages/braid-design-system/src/lib/components/TextLink/TextLink.docs.tsx
```diff
@@ -71,6 +80,34 @@ const docs: ComponentDocs = {
...
+      description: (
+        <Text>
+          By default, an icon will be <Strong>leading</Strong> the label,
+          however this can be changed by providing <Strong>iconPosition</Strong>{' '}
+          of <Strong>trailing</Strong>.
+        </Text>
+      ),
+      Example: () =>
+        source(
+          <Inline space="large" alignY="center">
```
mrm007, line 93: This example feels like it should just be with `iconPosition="trailing"`, like the one above it. Instead of two `TextLink`s wrapped in an `Inline`.

Or remove the previous example and just keep this one showing both uses?

packages/braid-design-system/src/lib/components/TextLink/TextLink.docs.tsx
```diff
@@ -71,6 +80,34 @@ const docs: ComponentDocs = {
...
+            <Text>
+              <TextLink icon={<IconArrow direction="left" />} href="">
+                Back
+              </TextLink>
+            </Text>
+            <Text>
+              <TextLink
+                icon={<IconArrow direction="right" />}
+                iconPosition="trailing"
```
mrm007, line 102: Running this in Playroom works, but `iconPosition` doesn't appear in the autocomplete dropdown.

iconPosition

packages/braid-design-system/src/lib/components/TextLink/TextLink.tsx
```diff
@@ -102,9 +103,38 @@ export const useLinkStyles = ({
   );
 };
 
+export const TextLinkContent = ({
+  icon,
+  iconPosition = 'leading',
+  children,
+}: Pick<TextLinkProps, 'icon' | 'iconPosition' | 'children'>) => (
+  <>
```
mrm007, line 111: Do we want make sure the icon doesn't break on a new line?

wrap

packages/braid-design-system/src/lib/components/TextLink/TextLink.tsx
```diff
@@ -102,9 +103,38 @@ export const useLinkStyles = ({
...
+      <Box component="span" paddingRight="xxsmall">
+        {icon}
+      </Box>
+    ) : null}
+    {children}
+    {icon && iconPosition === 'trailing' ? (
+      <Box component="span" paddingLeft="xxsmall">
+        {icon}
+      </Box>
+    ) : null}
```
mrm007, line 122: Do we want make sure the icon doesn't break on a new line?

wrap

packages/braid-design-system/src/lib/components/Button/Button.docs.tsx
```diff
@@ -150,6 +151,31 @@ const docs: ComponentDocs = {
           </Inline>,
         ),
     },
+    {
+      background: 'surface',
+      description: (
+        <Text>
+          By default, an icon will be <Strong>leading</Strong> the label,
+          however this can be changed by providing <Strong>iconPosition</Strong>{' '}
```
askoufis, line 159:
```suggestion
          however this can be changed by providing an <Strong>iconPosition</Strong>{' '}
```

packages/braid-design-system/src/lib/components/Button/Button.tsx
```diff
@@ -355,21 +357,42 @@ export const ButtonText = ({
...
           >
             {icon}
+            {
+              '⁠' /* Word joiner character, a zero-width non-breaking character to prevent the icon wrapping onto its own line */
+            }
           </Box>
         ) : null}
         {children}
         {loading ? <ButtonLoader /> : null}
+        {!loading && icon && iconPosition === 'trailing' ? (
```
askoufis, line 379: What's the reasoning behind not showing a trailing icon when `loading`?

packages/braid-design-system/src/lib/components/Button/Button.tsx
```diff
@@ -355,21 +357,42 @@ export const ButtonText = ({
...
+            className={[
+              styles.nowrap,
+              shouldReducePaddingX || bleed
+                ? null
+                : negativeMargin('right', 'xxsmall'),
+            ]}
+          >
+            {
+              '⁠' /* Word joiner character, a zero-width non-breaking character to prevent the icon wrapping onto its own line */
+            }
```
mrm007, line 392:
```suggestion
            {/* Word joiner character, a zero-width non-breaking character to prevent the icon wrapping onto its own line */}
            &#8288;
```

This makes it easier to [look it up] on the web ([see also]).

[look it up]: https://www.fileformat.info/info/unicode/char/2060/index.htm
[see also]: https://stackoverflow.com/a/28405917

packages/braid-design-system/src/lib/components/TextLink/TextLink.docs.tsx
```diff
@@ -71,6 +80,34 @@ const docs: ComponentDocs = {
           </Text>,
         ),
     },
+    {
+      description: (
+        <Text>
+          By default, an icon will be <Strong>leading</Strong> the label,
+          however this can be changed by providing <Strong>iconPosition</Strong>{' '}
```
askoufis, line 87:
```suggestion
          however this can be changed by providing an <Strong>iconPosition</Strong>{' '}
```

packages/braid-design-system/src/lib/components/TextLink/TextLink.tsx
mrm007: Upon further testing, the above suggestion doesn't actually work 🤔

packages/braid-design-system/src/lib/components/TextLink/TextLink.docs.tsx
askoufis: Now that you mention it, `icon` isn't showing up either, both in `Button` and `TextLink`. This is the case in master as well, not specific to this branch.

askoufis: Seems like playroom just isn't seeing those props.
Logged during `playroom start`.
```ts
{
  ButtonTypes: {
    tone: [ 'brandAccent', 'critical', 'neutral' ],
    variant: [ 'transparent', 'solid', 'ghost', 'soft' ],
    keyboardFocusable: [],
    radius: [ 'large', 'full' ],
    forceActive: []
  },
  TextLinkTypes: {
    weight: [ 'regular', 'weak' ],
    showVisited: [],
    hitArea: [ 'large', 'standard' ],
    reset: [
      'symbol',   'object', 'strong',   'body',       'small',
      'text',     'link',   'slot',     'style',      'title',
      'clipPath', 'filter', 'mask',     'marker',     'progress',
      'center',   'ruby',   'table',    'data',       'cite',
      'form',     'label',  'pattern',  'span',       'summary',
      'button',   'meter',  'textarea', 'sub',        'circle',
      'embed',    'pre',    'caption',  'menu',       'menuitem',
      'div',      'a',      'abbr',     'address',    'area',
      'article',  'aside',  'audio',    'b',          'base',
      'bdi',      'bdo',    'big',      'blockquote', 'br',
      'canvas',   'code',   'col',      'colgroup',   'datalist',
      'dd',       'del',    'details',  'dfn',        'dialog',
      'dl',       'dt',     'em',       'fieldset',   'figcaption',
      'figure',   'footer', 'h1',       'h2',         'h3',
      'h4',       'h5',     'h6',       'head',       'header',
      'hgroup',   'hr',     'html',     'i',          'iframe',
      'img',      'input',  'ins',      'kbd',        'keygen',
      'legend',   'li',     'main',     'map',        'mark',
      'meta',     'nav',    'noindex',  'noscript',   'ol',
      'optgroup', 'option', 'output',   'p',          'param',
      ... 76 more items
    ]
  }
}
```

mrm007: `ButtonTypes` seems to match the props for `ButtonOverlays`
https://github.com/seek-oss/braid-design-system/blob/37c6d00ea34a78e068673d33eaa6dbf01187c3e7/packages/braid-design-system/src/lib/components/Button/Button.tsx#L224-L230

askoufis: Ahhh, `react-docgen-typescript` must be picking it preferentially for some reason.

packages/braid-design-system/src/lib/components/Button/Button.tsx
michaeltaranto: Yeah found this with Ben earlier, have avoided using `nowrap` deliberately as that has larger implications on the label and its wrapping. Personally I dont think Buttons should wrap, but they do, and we want to avoid a breaking change.

packages/braid-design-system/src/lib/components/TextLink/TextLink.docs.tsx
```diff
@@ -71,6 +80,34 @@ const docs: ComponentDocs = {
           </Text>,
         ),
     },
+    {
+      description: (
+        <Text>
+          By default, an icon will be <Strong>leading</Strong> the label,
```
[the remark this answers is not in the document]
michaeltaranto: You good shout, i'll update it.

packages/braid-design-system/src/lib/components/TextLink/TextLink.docs.tsx
michaeltaranto: Yep agree, I'll simplify it to just be the one button demoing `trailing`

packages/braid-design-system/src/lib/components/TextLink/TextLink.docs.tsx
askoufis: [This option](https://github.com/styleguidist/react-docgen-typescript#componentnameresolver) might less us specify that the component we want is the one with the same name as the file. Not ideal though.

packages/braid-design-system/src/lib/components/TextLink/TextLink.tsx
michaeltaranto: Yeah its missing the `nowrap` rule on the span. All good i'm working standardising that solution across the multiple usages now.

mrm007: I was playing around and must have added the `nowrap` at some point to make it work. Happy for you to take it from here 👍

packages/braid-design-system/src/lib/components/Button/Button.tsx
michaeltaranto: Nice one yeah, i'll add this to the shared component i created for this problem.

packages/braid-design-system/src/lib/components/Button/Button.tsx
michaeltaranto: Presented a strange UX when combining both, i.e. the animated ellipsis followed by an icon. Ideally I'd like to use the trailing icon slot for a better loading indicator. What are your thoughts?

packages/braid-design-system/src/lib/components/TextLink/TextLink.docs.tsx
michaeltaranto: I have noticed this in the passed with other prop/component combinations, so definitely a good one to look into separately 👏

packages/braid-design-system/src/lib/components/Button/Button.tsx
askoufis: Yeah I was thinking the behaviour looked a bit weird, but don't really have a better solution. I think replacing the icon with a loading icon is a good idea.

packages/braid-design-system/src/lib/components/private/AvoidWidowIcon/AvoidWidowIcon.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { Box } from '../../Box/Box';
+
+import * as styles from './AvoidWidowIcon.css';
+
+interface Props extends Pick<BoxProps, 'children' | 'className'> {
+  iconPosition: 'leading' | 'trailing';
+  space?: Space;
+}
+
+export const AvoidWidowIcon = ({
```
mrm007, line 13: I'll batch a few suggestions into one and you can pick and choose what you like:

- make the `children` prop more restrictive
- rename the `children` prop to `icon` inside the render function, to make it clear (to humans) what it is
- move the comment above the HTML entity


```tsx
import type { UseIconProps } from '../../../hooks/useIcon';
import React from 'react';
import type { Space } from '../../../css/atoms/atoms';
import type { BoxProps } from '../../Box/Box';
import { Box } from '../../Box/Box';

import * as styles from './AvoidWidowIcon.css';

interface Props extends Pick<BoxProps, 'className'> {
  iconPosition: 'leading' | 'trailing';
  space?: Space;
  children: React.ReactElement<UseIconProps>;
}

export const AvoidWidowIcon = ({
  children: icon,
  iconPosition,
  space = 'xxsmall',
  className,
}: Props) => (
  <Box
    component="span"
    paddingRight={iconPosition === 'leading' ? space : undefined}
    paddingLeft={iconPosition === 'trailing' ? space : undefined}
    className={[styles.nowrap, className]}
    aria-hidden
  >
    {iconPosition === 'leading' ? icon : undefined}
    {/* Word joiner character, a zero-width non-breaking character to prevent the icon wrapping onto its own line */}
    &#8288;
    {iconPosition === 'trailing' ? icon : undefined}
  </Box>
);
```

mrm007 (approved):
Nice work with the `AvoidWidowIcon` component 🎩

PR #1247, opened 2023-03-17
Use latest (fastest) Crackle

mrm007:
This includes [Crackle 0.20.0](https://github.com/seek-oss/crackle/releases/tag/%40crackle%2Fcore%400.20.0) and the `.d.ts` fixes from https://github.com/seek-oss/crackle/pull/82

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "@babel/runtime": "^7.18.0",
     "@changesets/cli": "^2.22.0",
     "@changesets/get-github-info": "^0.5.0",
-    "@crackle/cli": "^0.10.6",
+    "@crackle/cli": "0.0.0-redirect-dts-20230317053050",
     "@manypkg/cli": "^0.19.1",
     "@octokit/rest": "^18.12.0",
     "@types/fs-extra": "^9.0.13",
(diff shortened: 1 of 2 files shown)


.prettierignore
```diff
@@ -7,4 +7,14 @@ CHANGELOG.md
...
+# generated snippets
 **/playroom/snippets/
+
+# backwards-compatible entry points (with package.json)
+/color-mode/
+/css/
+/playroom/
+/reset/
+/test/
+/themes/
```
askoufis, line 20: Confused about this change. Wouldn't this, for example, exclude [this folder](https://github.com/seek-oss/braid-design-system/tree/001b9ae2e7ee9f8f87fd928e1f08bc755a1231a4/packages/braid-design-system/src/lib/css) from formatting?

mrm007: Because the patterns start with `/` they only apply to folders from the root.


https://user-images.githubusercontent.com/3297808/226215957-c5a875c8-353b-4605-bd4e-176ed497ed12.mp4

askoufis: Oh of course. So in this case you want to apply specifically to the root of the `braid-design-system` package?

mrm007: `.prettierignore` and `.eslintignore` are symlinked from the root of the monorep, so adding `braid-design-system` would make Prettier and ESLint work from the root of the monorepo, but not from each package. At the moment we're running Prettier, ESLint, TypeScript etc. from each package, not from the root.

askoufis: Ah, forgot about the symlinking. Makes sense now.

askoufis (approved):

michaeltaranto (approved):

PR #1250, opened 2023-03-24
codemod: Fix build and enable running on Braid itself

michaeltaranto:
The codemod appears to have been broken through the course of converting to a monorepo, and once fixed we discovered that a downstream dependency is broken (`yoga-layout-prebuilt`) which appears unlikely to be fixed.

Worked with @mrm007 to internalise this dependency from the codemod build, so that we could patch it to unblock features coming through.

Also added support for running codemod over Braid repo itself, rather than just consumer repos.

### DIFF
--- a/packages/codemod/src/codemod.ts
+++ b/packages/codemod/src/codemod.ts
@@ -85,7 +85,7 @@ export function babelRecast({
 
     return {
       // @ts-expect-error
-      warnings: metadata && metadata.warnings ? metadata.warnings : [],
+      warnings: metadata?.warnings ?? [],
       // @ts-expect-error
       hasChanged: metadata && metadata.hasChanged ? metadata.hasChanged : false,
       // @ts-expect-error
--- a/packages/codemod/src/codemod.ts
+++ b/packages/codemod/src/codemod.ts
@@ -87,7 +87,7 @@ export function babelRecast({
       // @ts-expect-error
       warnings: metadata?.warnings ?? [],
       // @ts-expect-error
-      hasChanged: metadata && metadata.hasChanged ? metadata.hasChanged : false,
+      hasChanged: metadata?.hasChanged ?? false,
       // @ts-expect-error
       source: print(transformedAST).code,
     };
--- a/packages/codemod/src/codemod.ts
+++ b/packages/codemod/src/codemod.ts
@@ -85,9 +85,9 @@ export function babelRecast({
 
     return {
       // @ts-expect-error
-      warnings: metadata ? metadata.warnings : [],
+      warnings: metadata && metadata.warnings ? metadata.warnings : [],
       // @ts-expect-error
-      hasChanged: metadata ? metadata.hasChanged : false,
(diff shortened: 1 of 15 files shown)


packages/codemod/src/codemod.ts
```diff
@@ -85,9 +85,9 @@ export function babelRecast({
 
     return {
       // @ts-expect-error
-      warnings: metadata ? metadata.warnings : [],
+      warnings: metadata && metadata.warnings ? metadata.warnings : [],
```
mrm007, line 88:
```suggestion
      warnings: metadata?.warnings ?? [],
```

Gold plating (🏅) is my passion

packages/codemod/src/codemod.ts
```diff
@@ -85,9 +85,9 @@ export function babelRecast({
...
       // @ts-expect-error
-      hasChanged: metadata ? metadata.hasChanged : false,
+      hasChanged: metadata && metadata.hasChanged ? metadata.hasChanged : false,
```
mrm007, line 90:
```suggestion
      hasChanged: metadata?.hasChanged ?? false,
```

packages/codemod/src/plugin-deprecate/isBraidImport.test.ts
```diff
@@ -0,0 +1,37 @@
...
+describe('isBraidImport', () => {
+  describe('valid cases', () => {
+    [
+      'braid-design-system',
+      'braid-design-system/css',
+      'braid-src/lib',
+      'braid-src/lib/components',
+      'braid-src/lib/css',
+    ].forEach((source) =>
+      test(`${source}`, () => expect(isBraidImport(source)).toBeTruthy()),
```
mrm007, line 12: 🏅 Do you prefer this form over `test.each`?

packages/codemod/src/plugin-deprecate/isBraidImport.ts
```diff
@@ -0,0 +1,4 @@
+export const isBraidImport = (importSource: string) =>
+  /(braid-design-system(?:\/css)?|(braid-src\/lib(?:\/(css|components))?))$/.test(
```
mrm007, line 2: Given that this regular expression will be executed a few hundred times, maybe it's worth extracting it to a constant in the module scope. What do you think?

mrm007: 🏅 I'd split it into 2 regular expressions, for readability. Something like

```js
regexForBraid.test(importSource) || regexForInternal.text(importSource)
```

For most cases (consumers), the first branch is the one that will return early.

mrm007 (approved):
Nice one! A few [gold plating] suggestions marked with 🏅

packages/codemod/src/plugin-deprecate/isBraidImport.test.ts
michaeltaranto: Honestly I just forget its there 😄 . I'll update.

mrm007 (approved):

askoufis (approved):

PR #1251, opened 2023-03-27
RadioItem, Toggle: Use formAccent border when selected

michaeltaranto:
**RadioItem, Toggle:** Use formAccent border when selected

Switch to using the `formAccent` border colour, rather than the `field` border colour, when in the selected state (e.g. `checked` for `RadioItem`, `on` for `Toggle`).

### RadioItem
RadioItem before

RadioItem after

### Toggle
Toggle before

Toggle after

### DIFF
--- a/.changeset/hungry-donuts-suffer.md
+++ b/.changeset/hungry-donuts-suffer.md
@@ -1,0 +1,14 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated:
+  - Radio
+  - RadioItem
+  - Toggle
+---
+
+**RadioItem, Toggle:** Use formAccent border when selected
+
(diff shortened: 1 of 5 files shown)


mrm007 (approved):
😚👌

askoufis:
Any concerns about the `on` `Toggle`'s hover state? It's quite a subtle change now that the colour change is gone. The hover effect all but disappears at larger zoom levels since the box shadow is a fixed pixel width.

https://user-images.githubusercontent.com/5663042/227832611-1a2e7c27-1248-49d4-8190-984ed022d839.mov

michaeltaranto:
> Any concerns about the `on` `Toggle`'s hover state? It's quite a subtle change now that the colour change is gone. The hover effect all but disappears at larger zoom levels since the box shadow is a fixed pixel width.

Nice, that was super subtle based on browser zoom level for me, I think toggling the hover to only apply on the `off` position feels nice. I'll update that

askoufis (approved):

PR #1257, opened 2023-03-29
Badge: Adjust height to support different typographic scales

michaeltaranto:
**Badge:** Adjust height to support different typographic scales

Adjusts the height of `Badge` to be driven by the capHeight of `xsmall` text plus vertical padding, rather than `xsmall` line height. This enables different typographic scales across themes, while ensuring the `Badge` height is relative.


packages/braid-design-system/src/lib/components/Badge/Badge.css.ts
askoufis, line 1: 💥

askoufis (approved):
👌

PR #1259, opened 2023-03-30
Button, ButtonLink: Improve support for different typographic scales

michaeltaranto:
Ensure the height of a `small` sized `Button` is not reliant on the text line height.

This enables different typographic scales across themes, while ensuring the `Button` height is relative.

### DIFF
--- a/packages/braid-design-system/src/lib/components/Button/Button.css.ts
+++ b/packages/braid-design-system/src/lib/components/Button/Button.css.ts
@@ -72,8 +72,8 @@ export const small = style(
 );
 
 export const bleedVerticallyToCapHeight = style({
-  marginTop: calc(capHeightToMinHeight).negate().toString(),
-  marginBottom: calc(capHeightToMinHeight).negate().toString(),
+  marginTop: calc.negate(capHeightToMinHeight),
+  marginBottom: calc.negate(capHeightToMinHeight),
 });
 
 export const padToMinHeight = style({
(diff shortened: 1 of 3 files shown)


packages/braid-design-system/src/lib/components/Button/Button.css.ts
```diff
@@ -39,45 +40,45 @@ export const focusOverlay = style({
...
-      mobile: stylesForBreakpoint('mobile', 'standard'),
-      tablet: stylesForBreakpoint('tablet', 'standard'),
-    }),
-    [`${small}&`]: responsiveStyle({
-      mobile: stylesForBreakpoint('mobile', 'small'),
-      tablet: stylesForBreakpoint('tablet', 'small'),
-    }),
-  },
+  marginTop: calc(capHeightToMinHeight).negate().toString(),
+  marginBottom: calc(capHeightToMinHeight).negate().toString(),
```
mrm007, line 76:
```suggestion
  marginTop: calc.negate(capHeightToMinHeight),
  marginBottom: calc.negate(capHeightToMinHeight),
```
💅

packages/braid-design-system/src/lib/components/Button/Button.tsx
```diff
@@ -306,11 +305,8 @@ export const ButtonText = ({
   iconPosition = 'leading',
   variant = 'solid',
   tone,
-  labelSpacing = true,
   bleed,
-}: ButtonProps & {
-  labelSpacing?: boolean;
```
mrm007, line 312: Do we want to call out the removal of this prop? Although I don't actually see it referenced anywhere in this repo.

michaeltaranto: It is an internal component that we have refactored away from using and never cleaned up. So no comms needed.

PR #1260, opened 2023-03-30
Adding Crackle to renovate config

mrm007:
This gives us a faster feedback loop (to see if things break in compilation) and an easy way to create Braid snapshots for testing in apps


askoufis (approved):
💥

michaeltaranto (approved):

PR #1262, opened 2023-03-30
Add `xxxlarge` to the space scale

michaeltaranto:
Extends the range of the space scale to include `xxxlarge`.
This addition is to support an upcoming design uplift that requires greater fidelity in the scale.
Note, the new value is also available as a responsive property.

**EXAMPLE USAGE:**
```jsx
<Stack space="xxxlarge">...</Stack>

{/* Or responsively: */}
<Stack space={{ mobile: 'large', tablet: 'xxxlarge' }}>...</Stack>
```

```ts
import { atoms } from 'braid-design-system/css';

atoms({ paddingY: 'xxxlarge' })
```

```ts
import { vars } from 'braid-design-system/css';

const space = vars.space.xxxlarge;
```

### DIFF
--- a/.changeset/heavy-foxes-dress.md
+++ b/.changeset/heavy-foxes-dress.md
@@ -1,0 +1,47 @@
+---
+'braid-design-system': minor
+---
+
+---
+updated:
+  - atoms
+  - Bleed
+  - Box
+  - BoxRenderer
+  - Columns
+  - Disclosure
+  - Inline
+  - List
+  - MenuRenderer
+  - Stack
+  - Tabs
+  - Tiles
+  - useSpace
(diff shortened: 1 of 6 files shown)


askoufis (approved):
Bleeding edge meme right here, probably only displays properly on iOS 16.4.
```
🤏
🫸        🫸
```

https://emojipedia.org/rightwards-pushing-hand/
https://emojipedia.org/leftwards-pushing-hand/

michaeltaranto:
Updated to drop `xxxsmall` as there is currently no use case our prototypes.

askoufis:
Updated meme to drop `xxxsmall`

PR #1268, opened 2023-04-03
TooltipRenderer: Remove custom background

michaeltaranto:
Use the correct semantic token for the background of tooltips.
While there is no visual change, this is just a clean up to ensure the palette usage remains consistent.

### DIFF
--- a/packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
+++ b/packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
@@ -72,10 +72,10 @@ export const TooltipContent = ({
   >
     <Box
       boxShadow="large"
-      background="customDark"
+      background="neutral"
       borderRadius={borderRadius}
       padding="small"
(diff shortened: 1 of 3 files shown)


packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
```diff
@@ -72,10 +72,10 @@ export const TooltipContent = ({
   >
     <Box
       boxShadow="large"
-      background="customDark"
+      background="neutral"
```
mrm007, line 75: Does this affect dark mode in any way? Text legibility/contrast?

mrm007 (approved):

packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
michaeltaranto: Nope, exact some colour under the covers and "neutral" also changes the foreground context to use inverted text.

PR #1269, opened 2023-04-03
Alert, useToast: Design update

michaeltaranto:
### `Alert`
Refine the design to use consistent horizontal container inset, aligning content with elements like `Card`, as well as simplifying the design by removing the sidebar/keyline (consistent with `useToast`).

Before

After

### `useToast`
As a means to increase constrast against page content, the design has been updated to be presented on inverted backgrounds based on the color mode.
The design has also be refined to remove the sidebar/keyline (consistent with `Alert`), while also applying the `tone` to the provided `message`.

Before

After

Before

After

### DIFF
--- a/.changeset/honest-ants-sneeze.md
+++ b/.changeset/honest-ants-sneeze.md
@@ -1,0 +1,13 @@
+---
+'braid-design-system': minor
+---
+
+---
+updated:
+  - useToast
+---
+
+**useToast:** Design uplift with increased page contrast
+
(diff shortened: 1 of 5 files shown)


askoufis:
Is the toast action label weight changing to `weak` intentional? I can't see the change in the code, so I'm not sure how it's happening.

michaeltaranto:
Yep, it is moving to a `weak` link as a result inline with the contextual rules built into `TextLink` — now that it is on `surfaceDark` it defaults to `weak`.

We could force `regular` weight, but the design decision is to run with `weak`

askoufis (approved):

PR #1271, opened 2023-04-04
Add support for `caution` tone to form fields

michaeltaranto:
Add support for `caution` tone to form fields

Provide support for applying a `caution` tone to form fields.
Specifying this `tone` will provide the `IconCaution` alongside the provided `message`.

**EXAMPLE USAGE:**
```jsx
<TextField tone="caution" message="Caution message" />
```

### DIFF
--- a/.changeset/mighty-apes-grow.md
+++ b/.changeset/mighty-apes-grow.md
@@ -1,0 +1,24 @@
+---
+'braid-design-system': minor
+---
+
+---
+updated:
+  - Autosuggest
+  - Dropdown
+  - FieldMessage
+  - MonthPicker
+  - PasswordField
+  - Textarea
+  - TextField
+---
+
+Add support for `caution` tone to form fields
+
+Provide support for applying a `caution` tone to form fields.
+Specifying this `tone` will provide the `IconCaution` alongside the provided `message`.
+
+**EXAMPLE USAGE:**
+```jsx
+<TextField tone="caution" message="Caution message" />
+```
--- a/.changeset/mighty-apes-grow.md
+++ b/.changeset/mighty-apes-grow.md
@@ -16,7 +16,7 @@ updated:
 Add support for `caution` tone to form fields
 
 Provide support for applying a `caution` tone to form fields.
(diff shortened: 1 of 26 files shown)


.changeset/mighty-apes-grow.md
```diff
@@ -0,0 +1,24 @@
...
+  - MonthPicker
+  - PasswordField
+  - Textarea
+  - TextField
+---
+
+Add support for `caution` tone to form fields
+
+Provide support for applying a `caution` tone to form fields.
+Specifying this `tone` will provide the `IconCaution` alongside the provided `message`.
```
askoufis, line 19: `provide` sounds a bit weird here IMO.
```suggestion
Specifying this `tone` will show the `IconCaution` alongside the provided `message`.
```

packages/braid-design-system/src/lib/components/RadioGroup/RadioGroup.tsx
```diff
@@ -22,6 +22,7 @@ export type RadioGroupBaseProps<Value = NonNullable<string | number>> =
     onChange: (event: FormEvent<HTMLInputElement>) => void;
     name?: string;
     size?: InlineFieldProps['size'];
+    tone?: Exclude<FieldGroupBaseProps['tone'], 'caution'>;
```
askoufis, line 25: Should we use `Extract` instead of `Exclude` here? If a new tone was added tomorrow, `RadioGroup` would automatically have it as one of its `tone`s. If the new tone was never intended to be supported by `RadioGroup`, then you'd have to do a breaking change to remove it from the prop type.

Unrelated to this change, but should [the docs](https://seek-oss.github.io/braid-design-system/components/RadioGroup#message-and-tone) mention `positive` the tone here since it's on the prop type?

michaeltaranto: Yeah I like that point. Will update, and interesting catch on `positive` being missing too!

askoufis (approved):

PR #1280, opened 2023-04-11
Checkbox, RadioItem: Fix alignment with updated `Badge` layout

michaeltaranto:
Improve layout to work with updated `Badge` layout which is no longer driven by line height.


mrm007 (approved):
✅ 📻 🦡

PR #1281, opened 2023-04-11
Add `seekJobs` theme

michaeltaranto:
## Details

The `seekJobs` theme encapsulates the system changes necessary to apply and deliver the updated visual design language for SEEK Jobs.
Through the development of this theme, we have been able improve the fidelity of the various scales in our tokens, while also ensuring that the tokens themselves are consumed and applied more consistently throughout the system itself.

### Palette change

There is an update to the centralised palette around the hues of `violet` and `purple`. For this reason, the existing `violet` swatch has been inlined with the base `apac` theme to maintain backwards compatibility.

### Legacy theme token

Introduces a `legacy` flag on the theme to be used as a landmark for delineating design uplift without disturbing how existing apps use the existing themes today. We will be sunsetting some concepts and spacing with the existing themes, and look to support the new patterns going forward.

This flag will be removed when all themes have adopted common usage patterns.

## Preview

Side by side comparison of some key elements to visualise how `apac` (left) differs from `seekJobs` (right).
Comparison of themes


.changeset/long-gifts-protect.md
```diff
@@ -0,0 +1,31 @@
...
+import seekJobs from 'braid-design-system/themes/seekJobs';
+
+<BraidProvider theme={seekjobs}>
+  ...
+</BraidProvider>
+```
+
+**MIGRATION**
+
+Consumers of the apac theme are not recommended to migrate independently. The seekJobs theme represents an uplifted visual identity that is part of a wider visual uplift.
```
askoufis, line 26:
```suggestion
Consumers of the `apac` theme are not recommended to migrate independently. The `seekJobs` theme represents an uplifted visual identity that is part of a wider visual uplift.
```

.changeset/long-gifts-protect.md
```diff
@@ -0,0 +1,31 @@
...
+Instead, we’ll be guiding the initial teams through a staged migration in coordination with the centralised team process.
+There are some differences in how certain concepts are applied, whether it's the space scale, or `Card` usage, etc. and we will be documenting these in due course.
```
askoufis, line 28:
```suggestion
There are some differences in how certain concepts are applied, whether it's the space scale, or `Card` usage, etc., and we will be documenting these in due course.
```

packages/braid-design-system/src/lib/components/Pagination/Pagination.tsx
```diff
@@ -107,9 +109,10 @@ const Page = ({ number, current }: { number: number; current: boolean }) => {
         boxShadow="borderFormAccent"
         className={{
           [styles.lightModeCurrentKeyline]:
-            parentBackground.lightMode !== 'surface' && current,
-          [styles.darkModeCurrentKeyline]:
-            parentBackground.darkMode !== 'surface' && current,
+            isLegacyTheme &&
+            parentBackground.lightMode !== 'surface' &&
+            current,
+          [styles.darkModeCurrentKeyline]: current,
```
askoufis, line 115: Is [this statement on the `Pagination` docs](https://seek-oss.github.io/braid-design-system/components/Pagination/#contextual-design) still correct with this change?

packages/braid-design-system/src/lib/components/Card/Card.tsx
```diff
@@ -68,16 +70,19 @@ export const Card = ({
     ]);
   }
 
+  const roundingForTheme = !isLegacyTheme ? borderRadius : resolvedRounding;
```
mrm007, line 73: I would flip the condition to make the exception more obvious i.e.

> If (exception) do something
> Else do the default

It's also consistent with the other ternaries introduced in this PR.

packages/braid-design-system/src/lib/themes/wireframe/tokens.ts
```diff
@@ -50,6 +50,7 @@ const getHoverColor = (x: string) =>
 const tokens: BraidTokens = {
   name: 'wireframe',
   displayName: 'Wireframe',
+  legacy: false,
```
mrm007, line 53: Does this mean once `apac` is dropped this theme will become the wireframe for `seekJobs`?

packages/braid-design-system/src/lib/themes/index.ts
mrm007, line 1: Should we get rid of this file?

.changeset/long-gifts-protect.md
```diff
@@ -0,0 +1,31 @@
...
+Add `seekJobs` theme
+
+The `seekJobs` theme encapsulates the system changes necessary to apply and deliver the updated visual design language for SEEK Jobs.
+Through the development of this theme, we have been able improve the fidelity of the various scales in our tokens, while also ensuring that the tokens themselves are consumed and applied more consistently throughout the system itself.
+
+**EXAMPLE USAGE:**
+```tsx
+import seekJobs from 'braid-design-system/themes/seekJobs';
+
+<BraidProvider theme={seekjobs}>
```
mrm007, line 19:
```suggestion
<BraidProvider theme={seekJobs}>
```

.changeset/long-gifts-protect.md
mrm007, line 26:
```suggestion
Consumers of the apac theme are not recommended to migrate independently. The `seekJobs` theme represents an uplifted visual identity that is part of a wider visual uplift.
```

mrm007 (approved):

.changeset/long-gifts-protect.md
michaeltaranto: Taking the above suggestion which covers both theme references.

packages/braid-design-system/src/lib/components/Pagination/Pagination.tsx
michaeltaranto: Both this change and the `Alert` change make that statement theme specific. Will have to consider how to approach this.

packages/braid-design-system/src/lib/themes/index.ts
michaeltaranto: Maybe, I think this was in service of the `BraidLoadableProvider` which was removed a while ago. Happy to consider this separate to this PR though.

askoufis (approved):

PR #1285, opened 2023-04-18
Card, Column: Add support for full height

michaeltaranto:
### `Card`
Provide support for making a `Card` use all available vertical space. This is useful when laying out rows of `Card` elements and having them all be equal height.

**EXAMPLE USAGE:**
```jsx
<Card height="full">
  ...
</Card>
```

---
 
### `Column`
Provides support for full height content by growing the all `Column` elements to be uniform in height.

This will have no effect on the content itself, but enables consumers to create designs of uniform content, such as rows of `Card` elements.

### DIFF
--- a/.changeset/warm-terms-film.md
+++ b/.changeset/warm-terms-film.md
@@ -1,0 +1,15 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated:
+  - Columns
+  - Column
+---
+
+**Column:** Support full height content
+
+Provides support for full height content by growing the all `Column` elements to be uniform in height.
+
+This will have no effect on the content itself, but enables consumers to create designs of uniform content, such as rows of `Card` elements.
--- a/.changeset/warm-terms-film.md
+++ b/.changeset/warm-terms-film.md
@@ -10,6 +10,6 @@ updated:
 
 **Column:** Support full height content
 
-Provides support for full height content by growing the all `Column` elements to be uniform in height.
+Provide support for full height content by growing all `Column` elements to be uniform in height.
 
 This will have no effect on the content itself, but enables consumers to create designs of uniform content, such as rows of `Card` elements.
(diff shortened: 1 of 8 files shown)


.changeset/warm-terms-film.md
```diff
@@ -0,0 +1,15 @@
...
+
+---
+updated:
+  - Columns
+  - Column
+---
+
+**Column:** Support full height content
+
+Provides support for full height content by growing the all `Column` elements to be uniform in height.
```
askoufis, line 13:
```suggestion
Provide support for full height content by growing all `Column` elements to be uniform in height.
```

packages/braid-design-system/src/lib/components/Card/Card.docs.tsx
```diff
@@ -21,6 +21,42 @@ const docs: ComponentDocs = {
...
+      label: 'Controlling the height',
+      description: (
+        <>
+          <Text>
+            By default, <Strong>height</Strong> will be set to{' '}
+            <Strong>content</Strong> — growing to accomodate what is inside it.
+          </Text>
+          <Text>
+            Alternatively, choose a height of <Strong>full</Strong> to use all
+            available vertical space. This is useful when composing rows of Card
```
askoufis, line 34: Is it worth linking to the `Card` docs here?

packages/braid-design-system/src/lib/components/Column/Column.screenshots.tsx
```diff
@@ -122,5 +122,49 @@ export const screenshots: ComponentScreenshot = {
...
+                </Box>
+              </Column>
+            </Columns>
+          ))}
+        </Stack>
+      ),
+    },
+    {
+      label:
+        'Full height column, where column with shorter content has specified width',
```
askoufis, line 149:
```suggestion
        'Full height column, where column with taller content has specified width',
```

packages/braid-design-system/src/lib/components/Card/Card.docs.tsx
askoufis: Oh, this is in the card docs 🤦

askoufis (approved):

PR #1288, opened 2023-04-19
TextLink, TextLinkButton: Update underline design

michaeltaranto:
Uplift the design of the the text underline used on `TextLink` and `TextLinkButton` components.

Increasing the thickness of the underline based on the font size, as well as the text offset.


askoufis (approved):
LGTM.

Non-blocking suggestion: should we have some Thai text screenshots in addition to English?

michaeltaranto:
> LGTM.
> 
> Non-blocking suggestion: should we have some Thai text screenshots in addition to English?

Yeah I think this is a good idea and to run it across the typography screenshots too.

Would definitely prefer considering this as a separate PR more broadly.

PR #1289, opened 2023-04-20
Text, Heading: Only show truncate deprecation message when true

michaeltaranto:
Only show the truncate deprecation message when `truncate` is provided and set to `true`

Given we default it to `false` in the Typography component, we are currently showing the message a few too many times 🤦‍♂️

### DIFF
--- a/packages/braid-design-system/src/lib/components/private/Typography/Typography.tsx
+++ b/packages/braid-design-system/src/lib/components/private/Typography/Typography.tsx
@@ -43,7 +43,7 @@ export const Typography = ({
     );
 
   if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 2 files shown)


packages/braid-design-system/src/lib/components/private/Typography/Typography.tsx
```diff
@@ -43,7 +43,7 @@ export const Typography = ({
     );
 
   if (process.env.NODE_ENV !== 'production') {
-    if (typeof truncate !== 'undefined') {
+    if (typeof truncate !== 'undefined' && truncate) {
```
askoufis, line 46: Can't this just be a truthy check?
```suggestion
    if (truncate) {
```

askoufis (approved):

PR #1294, opened 2023-04-26
Hide field borders in dark containers

michaeltaranto:
Reduce visual noise when a form field is displayed in a dark container by hiding the default border.
As fields are light on light backgrounds, the border is used to delineate its bounds against the container, which is not relevant in a dark container.

### Previews

Before:
Before

After:
After

### DIFF
--- a/packages/braid-design-system/src/lib/components/private/InlineField/StyledInput.tsx
+++ b/packages/braid-design-system/src/lib/components/private/InlineField/StyledInput.tsx
@@ -9,6 +9,7 @@ import type { BoxProps } from '../../Box/Box';
 import { Box } from '../../Box/Box';
 import * as styles from './InlineField.css';
 import type { Size } from './InlineField.css';
+import { useBackgroundLightness } from '../../Box/BackgroundContext';
 
 const tones = ['neutral', 'critical'] as const;
 export type InlineFieldTone = (typeof tones)[number];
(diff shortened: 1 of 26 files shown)


packages/braid-design-system/src/lib/components/private/InlineField/StyledInput.tsx
```diff
@@ -210,6 +213,10 @@ export const StyledInput = forwardRef<
             variant={disabled ? 'disabled' : defaultBorder}
             borderRadius={fieldBorderRadius}
             visible={tone !== 'critical' || disabled}
+            className={{
+              [styles.hideBorderOnDarkBackgroundInLightMode]:
+                lightMode === 'dark',
```
askoufis, line 218: What is `lightMode === 'dark'` conveying here? Just reading it is a bit confusing.

askoufis (approved):

packages/braid-design-system/src/lib/components/private/InlineField/StyledInput.tsx
michaeltaranto: `useBackgroundLightness()` returns and object containing whether the current background colour is `light` or `dark`, and for each colour mode.
So the logic is, if the current background colour applied in `lightMode` is a *dark* colour...

Does that make sense?

askoufis: Ah, that's a lot clearer. 👍

PR #1309, opened 2023-05-29
Update renovate config

mrm007:
I was looking through [the logs](https://developer.mend.io/github/seek-oss/braid-design-system/-/job/01886460-371c-7418-99bf-c90590adb035) trying to figure out why `@crackle/cli` wasn't picked up by Renovate and I noticed that the config was using an old format.
[`matchPackagePatterns`](https://docs.renovatebot.com/configuration-options/#matchpackagepatterns) is a regex, which is probably why it wasn't picking up `@crackle/*`, so I replaced `packageNames` with [`matchPackagePrefixes`](https://docs.renovatebot.com/configuration-options/#matchpackageprefixes) 🤞

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -3,12 +3,12 @@
   "commitMessagePrefix": null,
   "packageRules": [
     {
-      "packageNames": ["playroom", "@capsizecss/*", "@crackle/*"],
+      "matchPackagePrefixes": ["playroom", "@capsizecss/", "@crackle/"],
       "enabled": true
     },
     {
       "groupName": "@testing-library packages",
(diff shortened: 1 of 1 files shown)


askoufis (approved):

askoufis:
Not sure why the UI snapshots aren't being accepted, I've accepted these changes twice already 🤔

mrm007:
It looks like the job for the last commit to master was incorrectly [skipped](https://github.com/seek-oss/braid-design-system/actions/runs/5107377818/jobs/9180611893). The workflow needs a bit of work, so I'll do that before merging this PR.

PR #1316, opened 2023-06-21
Drawer updates

michaeltaranto:
A suite of related changes to `Drawer`.

Aligning the horizontal gutter with `PageBlock` gutters to ensure content on mobile has the same available space both in and outside a `Drawer`.

With the above comes some other minor fixes/updates. See changesets for details.

### DIFF
--- a/packages/braid-design-system/src/lib/components/private/Modal/Modal.css.ts
+++ b/packages/braid-design-system/src/lib/components/private/Modal/Modal.css.ts
@@ -1,3 +1,4 @@
+import { rgba } from 'polished';
 import { style } from '@vanilla-extract/css';
 import { calc } from '@vanilla-extract/css-utils';
 import { externalGutter } from './ModalExternalGutter';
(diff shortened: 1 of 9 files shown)


mrm007 (approved):

packages/braid-design-system/src/lib/components/private/Modal/Modal.css.ts
```diff
@@ -12,9 +13,16 @@ export const fixedStackingContext = atoms({
...
-export const backdrop = style({
-  background: 'rgba(0, 0, 0, .4)',
-});
+export const backdrop = style(
+  colorModeStyle({
+    lightMode: {
+      background: 'rgba(0, 0, 0, .4)',
+    },
+    darkMode: {
+      background: 'rgba(0, 0, 0, .6)',
```
askoufis, line 22: Should we be using hex codes (polished `rgba`) for better browser support?

michaeltaranto: Hmm, possibly. Docs seems a little vague, but maybe missing support Safari 9-12. If that is the case, might be worth sweeping the code base as a follow up, as i can see a few around.

michaeltaranto: Just confirming for this instance though, we cant use polished as we need transparency. Given its the backdrop as a standalone element, i'll split it into `background` and `opacity` properties.

askoufis (approved):

PR #1318, opened 2023-06-23
Drawer, Dialog: Support validation blocking closure of modal

michaeltaranto:
To prevent a `Dialog` or `Drawer` from closing, e.g. due to validation, the `onClose` function can now return **false**.

**EXAMPLE USAGE:**
```jsx
<Drawer
  onClose={() => {
    const valid = runValidation();
    if (!valid) {
      return false;
    }

    closeDrawer();
  }}
/>
```

### DIFF
--- a/.changeset/spotty-wasps-poke.md
+++ b/.changeset/spotty-wasps-poke.md
@@ -1,0 +1,27 @@
+---
+'braid-design-system': minor
+---
+
+---
+updated:
+  - Drawer
+  - Dialog
+---
+
+**Drawer, Dialog:** Support validation blocking closure of modal
+
+To prevent a `Dialog` or `Drawer` from closing, e.g. due to validation, the `onClose` function can now return **false**.
+
+**EXAMPLE USAGE:**
+```jsx
+<Drawer
+  onClose={() => {
+    const valid = runValidation();
+    if (!valid) {
+      return false;
+    }
+
+    closeDrawer();
+  }}
+/>
+```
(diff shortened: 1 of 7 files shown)


.changeset/spotty-wasps-poke.md
```diff
@@ -0,0 +1,27 @@
...
+**EXAMPLE USAGE:**
+```jsx
+<Drawer
+  onClose={() => {
+    const valid = runValidation();
+    if (!valid) {
+      return false;
+    }
+
+    closeDrawer();
```
mrm007, line 24: Is this necessary? Without looking at the implementation, my expectation would be that this would let the drawer close normally.

michaeltaranto: It is just an example. That would be the consumers controlled state function that drives the `open` prop

mrm007: Because the `open` prop is not provided in the example, I assumed it's not a controlled component.

mrm007, line 24:
```suggestion
  open={open}
  onClose={() => {
    const valid = runValidation();
    if (!valid) {
      return false;
    }

    setOpen(false);
```

michaeltaranto: Yeah all good, we tend to highlight the changed prop, as using valid code with all required props can be very noisy/too verbose for a lot of components

packages/braid-design-system/src/lib/components/Drawer/Drawer.docs.tsx
mrm007, line 1: I noticed the drawer maintains the "validation" state between opening/closing. Is this something we'd want to call out and provide a way to mitigate potential bugs?

michaeltaranto: Not really something I think Braid should get involved in. Consumers need to maintain state in the most sensical way for their use case.

IMHO, if its state that Braid doesn't own, documenting common pitfalls around state management would be a slippery slope for us to be tackling.

askoufis (approved):

PR #1320, opened 2023-07-03
IconEnlarge: Add new component

michaeltaranto:
**IconEnlarge:** Add new component

**EXAMPLE USAGE:**

```jsx
<IconEnlarge />
```

The `active` prop can be used to toggle the icon into the "reduce" state:

```jsx
<IconEnlarge active={true} />
```


packages/braid-design-system/src/lib/components/icons/IconEnlarge/IconEnlarge.docs.tsx
```diff
@@ -0,0 +1,53 @@
...
+    ),
+  alternatives: [],
+  additional: [
+    {
+      label: 'Toggling active state',
+      description: (
+        <Text>
+          Often used to toggle between englarged and reduced states, setting the{' '}
+          <Strong>active</Strong> prop to <Strong>true</Strong> will reverse the
+          icons intent.
```
askoufis, line 34:
```suggestion
          icon's intent.
```

packages/braid-design-system/src/lib/components/icons/IconEnlarge/IconEnlarge.docs.tsx
```diff
@@ -0,0 +1,53 @@
...
+      Example: ({ getState, toggleState }) =>
+        source(
+          <Stack space="large" dividers>
+            <IconEnlarge active={getState('active')} />
+            <Inline space="small">
+              <Button onClick={() => toggleState('active')}>
+                {getState('active') ? 'inactive' : 'active'}
+              </Button>
+            </Inline>
+          </Stack>,
```
askoufis, line 46: Do you think an example using a `ButtonIcon` be better here? It's likely more reflective of a real-world use case compared to this which is a bit too trivial IMO.

michaeltaranto: Probably a good point, I followed the existing state toggle examples from the other icon docs (e.g. Bookmark, Heart, etc).

Let me look at updating the set of them

michaeltaranto: Updated all 3 now.

askoufis (approved):

PR #1343, opened 2023-08-17
Page: Add component

michaeltaranto:
The new `Page` component establishes a consistent page-level layout, by managing the relationship between the footer and the main content.

By default, for pages with limited content the `footer` will at a minimum be placed at the bottom of the screen, pushed beyond as the page content grows.

For pages with dynamic content, it is recommended to place the footer out of view by setting the `footerPosition` prop to `belowFold` to prevent the footer from popping in and out of view when the page content changes, e.g. toggling between a loading indicator and content.

**EXAMPLE USAGE:**
```jsx
<Page footer={<Footer />}>
  <Header />
  {/* page content... */}
</Page>
```

### DIFF
--- a/packages/braid-design-system/src/lib/components/Page/Page.tsx
+++ b/packages/braid-design-system/src/lib/components/Page/Page.tsx
@@ -40,7 +40,7 @@ export const Page = ({
           className={
             footerPosition === 'belowFold' ? styles.fullHeight : undefined
           }
-          paddingBottom="xxlarge"
+          paddingBottom="xxxlarge"
         >
           {children}
         </Box>
--- a/packages/braid-design-system/src/lib/components/Page/Page.tsx
+++ b/packages/braid-design-system/src/lib/components/Page/Page.tsx
@@ -1,9 +1,10 @@
-import type { ReactNode } from 'react';
-import React from 'react';
+import assert from 'assert';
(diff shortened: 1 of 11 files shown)


packages/braid-design-system/src/lib/components/Page/Page.tsx
```diff
@@ -20,20 +21,31 @@ export const Page = ({
...
+        flexDirection="column"
+        className={styles.fullHeight}
+        {...buildDataAttributes({ data, validateRestProps: restProps })}
+      >
+        <Box
+          flexGrow={footerPosition !== 'belowFold' ? 1 : undefined}
+          className={
+            footerPosition === 'belowFold' ? styles.fullHeight : undefined
+          }
+          paddingBottom="xxlarge"
```
askoufis, line 43: Are consumers going to want to customize this padding?

jahredhope (approved):

packages/braid-design-system/src/lib/components/Page/Page.tsx
michaeltaranto: Actually think we should start at `xxxlarge` given that is the loosest size used for page stacks, and then we decide if we should open up control later only if required.

askoufis (approved):

PR #1345, opened 2023-08-21
TabPanel: Align focus outline radius to scale

michaeltaranto:
Increase the radius of the focus outline to better align to the scale. A `TabPanel` is typically a "large" element containing entire sections of UI, so using the `large` radius to better align to the radius scale.

## Screenshots

Example of before and after with the tab panel focused:


jahredhope (approved):
Very clear reasoning, thanks.

PR #1353, opened 2023-08-31
Bump esbuild in codemod and generate-component-docs

mrm007:
Fixes the codemod not working after https://github.com/seek-oss/braid-design-system/pull/1339


askoufis (approved):

mrm007:
Not sure what's with the Chromatic diffs @michaeltaranto. Do you mind confirming those are valid changes?

PR #1361, opened 2023-09-12
Upgrade TypeScript and ESLint config

mrm007:
Most changes are the result of `eslint --fix`, with a few highlighted exceptions.


site/src/App/routes/guides/playroom-prototyping.tsx
```diff
@@ -215,7 +215,7 @@ const PlayroomPrototyping = () => (
         return {
           ...result,
           code: result.code
-            .replace(/ as Array<MockJob>/g, '')
+            .replace(/ as MockJob\[\]/g, '')
```
mrm007, line 218: This will be handled by the `source.macro` package (in a separate PR)

package.json
```diff
@@ -76,13 +76,17 @@
...
   },
   "volta": {
     "node": "18.15.0"
   },
   "packageManager": "pnpm@7.30.0",
   "pnpm": {
+    "overrides": {
+      "@braid-design-system/generate-component-docs": "workspace:*",
+      "braid-design-system": "workspace:*"
+    },
```
mrm007, line 89: Had to do these shenanigans until `manypkg` supports the `workspace:` protocol https://github.com/Thinkmill/manypkg/pull/186

packages/generate-component-docs/src/__snapshots__/contract.test.ts.snap
```diff
@@ -8207,12 +8209,12 @@ exports[`filterSuggestions 1`] = `
   props: {
```
mrm007, line 8209: Looks like this export is incorrectly treated as a component, hence the bogus definition for "props".

packages/braid-design-system/package.json
```diff
@@ -200,7 +200,7 @@
     "fs-extra": "^10.1.0",
     "html-validate": "^7.1.1",
     "playroom": "0.32.1",
-    "prettier": "^2.8.2",
+    "prettier": "^2.8.8",
```
mrm007, line 203: GitHub wouldn't let me merge because it thought `pnpm-lock.yaml` had conflicts (it didn't) so I made an unrelated change to update the lockfile.

michaeltaranto (approved):
👏

PR #1362, opened 2023-09-13
Create `source.macro` package

mrm007:
The source macro allows code snippets to be extracted at build time. For a live example, [see the first example and code snippet for `Button`](https://seek-oss.github.io/braid-design-system/components/Button) (and [the source code for that example](https://github.com/seek-oss/braid-design-system/blob/c36e6843c219c8174a2fb402e824ed695119cacf/packages/braid-design-system/src/lib/components/Button/Button.docs.tsx#L24-L34)).

This PR extracts the existing macro into its own package, plus a few dependency bumps. Recommended to review commits individually because there's one noisy commit that changes an import across 261 files.


fixtures/consumer/build.test.ts
```diff
@@ -34,7 +34,7 @@ describe('build', () => {
     // eslint-disable-next-line no-console
     console.log('Running `pnpm build`...');
     await promisify(exec)('pnpm build');
-  });
+  }, 60_000);
```
michaeltaranto, line 37: Was this for debugging, or is it needed for something im missing?

mrm007: I noticed it was sometimes timing out. Maybe it was just my machine, but the default 30 seconds. I can't seem to find an instance of it failing in CI...

michaeltaranto (approved):

PR #1368, opened 2023-09-26
docs: Improve Dialog and Drawer documentation

michaeltaranto:
Updating the Dialog and Drawer documentation with:
- Better static visuals, beyond the interaction previews.
- Documented `title` and `description` impact on screen readers
- Updated “Design considerations” section to no longer discourage use, but rather encourage for the right behaviours


packages/braid-design-system/src/lib/components/Dialog/Dialog.docs.tsx
mrm007, line 1: A bizarre thing happens in Firefox and Chromium/Arc, but not in Safari.

Using the [first example], when I open multiple dialogs this is what I see:

https://github.com/seek-oss/braid-design-system/assets/3297808/709e5e55-e42f-434f-9205-df87871551c1


[first example]: https://braid-design-system--4953c2777c2971b6bef6ff24429a4bba059f8286.surge.sh/playroom/#?code=N4Igxg9gJgpiBcIA8BJAdgGwJZpgAgGcAHAQzBgF4AdEAgWxIwxoD4q088kAhAVwBd%2BEDsIDC2MAGsKwABQBKPBRZ4hAczUYYAZX4l%2BMWTShZGENTXkBfFgHkiMDibNqkAej6DhbNO-TZcH3YkABFTDHN2Tn4sfi1qEABRAA8SOiItPAAVWK0aKLxYAjAAJywiGOEZAs4kLJhk-lVhShoCGEg0KBISgE9WGs48AEERCqxhRkKYYrLx4TwIADNVAAt8MJc8ToM0fkH3esafTisCiAc0GTUYfl19Q2NwyJBrc7RxCHaZBSUVdU0Oj0BiMIGcEQsrzOaBOXAAChgyDBVhAMLASnh1lg1Kt%2BDIAIwABkJVjwAHcsFB%2BKsEkTCQBSGh4REAIxgGASmwheFEwl2%2BxAeDcPncXPMLBAABoQNSYHQZggANoAXWlFKpqwISuVViAA

michaeltaranto: Ah this is actually something I want to fix. The scroll lock is aggressively locking your focus to that open Dialog. You'll notice you cant even edit the code!

michaeltaranto: To clarify that fix will be a separate piece of work though as its more than the scope of this docs update.

packages/braid-design-system/src/lib/components/Dialog/Dialog.docs.tsx
```diff
@@ -70,13 +103,96 @@ const docs: ComponentDocs = {
...
+            <Text>
+              Open in Playroom and enable your screen reader to preview the
+              announcements.
+            </Text>
+          </Alert>
+        </>
+      ),
+      background: false,
+      Example: ({ id, getState, toggleState }) => {
+        const { code } = source(
```
mrm007, line 127: I'm not very excited about all the duplication. It's very easy for the code and example to drift apart and it's very easy to miss in a review.

Perhaps we can avoid it by using the `flattenChildren` util? Something like this (I tested the code locally):

```tsx
const { code, value } = source(
  <>
    <Inline /* with all its props */ />

    <Dialog /* with all its props */ />
  </>
);
const [toggle, dialog] = flattenChildren(value); // => [<Inline />, <Dialog />];

return {
  code,
  value: (
    <Box borderRadius="xlarge" overflow="hidden">
      <DialogPreview>
        <DialogContent
          {...(dialog as React.ReactElement).props}
          open={() => {}}
          onClose={() => {}}
          scrollLock={false}
        />
        <Screen />
      </DialogPreview>
    </Box>
  ),
};
```

What do you think?

michaeltaranto: Yeah i felt that. I like the work around idea. I'll riff on that a little bit and update 👏

michaeltaranto: Actually, pivoted a bit here. Given the preview is just the Dialog or Drawer, having the Button in the code was confusing. So switching these examples to demo with the `open={true}` prop.

mrm007: Nice 👌

mrm007 (approved):

PR #1369, opened 2023-09-27
docs: Move theme selector to header

michaeltaranto:
Moving the theme selector into the header on the docs site, so that it its always available regardless of scroll position in the side bar navigation.

Also adding some affordance to the top of the side bar when scrolled to improve affordance of the current scroll position not being at the top of the navigation.

| | Top of page | Scrolled navigation |
| --- | --- | --- |
|  Before | ![Top Before]  | ![Scrolled Before] |
|  After | ![Top After] | ![Scrolled After] |

[Top Before]: https://github.com/seek-oss/braid-design-system/assets/912060/d6937d39-272e-47ca-b99a-3107b31ce655
[Top After]: https://github.com/seek-oss/braid-design-system/assets/912060/a59639ea-52f4-43d4-808b-f7229f0e3634
[Scrolled Before]: https://github.com/seek-oss/braid-design-system/assets/912060/32d8897d-a9db-4453-b61e-8e7d88a95ade
[Scrolled After]: https://github.com/seek-oss/braid-design-system/assets/912060/9b1c6181-4c03-4232-b4e1-9bc14765c8c9

### DIFF
--- a/site/src/App/Navigation/Navigation.css.ts
+++ b/site/src/App/Navigation/Navigation.css.ts
@@ -60,6 +60,11 @@ const backgroundVar = createVar();
 const shadowVar = createVar();
 const transparent = 'rgba(0, 0, 0, 0)';
 const scrollShadows = style([
+  {
+    backgroundPosition: 'center top',
+    backgroundRepeat: 'no-repeat',
+    backgroundAttachment: 'local, scroll',
+  },
   colorModeStyle({
     lightMode: {
       vars: {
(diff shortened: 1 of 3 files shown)


site/src/App/Navigation/Navigation.css.ts
```diff
@@ -56,10 +56,46 @@ const subNavOffsetAboveMobile = style(
...
+      vars: {
+        [backgroundVar]: vars.backgroundColor.bodyDark,
+        [shadowVar]: palette.grey['800'],
+      },
+      backgroundImage: [
+        `linear-gradient(45deg, ${backgroundVar}, ${backgroundVar})`,
+        `linear-gradient(90deg, ${transparent}, ${shadowVar} 15%, ${shadowVar} 85%, ${transparent})`,
+      ].join(', '),
+      backgroundSize: '100% 2px',
+    },
```
mrm007, line 90: Just checking if this intentionally looks different from the light mode version.

dark

vs

light

michaeltaranto: Yep. Shadows arent really a recommended way of differentiating layers in dark mode. So I opted for a different treatment.

mrm007: Personally, I'd like to have it a bit bigger, especially for larger screens where it can be hard to notice.

mrm007 (approved):

PR #1374, opened 2023-10-13
TooltipRenderer: Re-evaluate position when trigger or children changes

michaeltaranto:
Fixes an issue where the tooltip would not re-evaluate it's position when the `trigger` or `children` prop changed while the tooltip was already open.

https://github.com/seek-oss/braid-design-system/assets/912060/72080336-719d-47fc-bb53-5582d81d83c9

### Playrooms
[Before] / [After]

[Before]: https://braid-design-system--bb744806600c5337caa6b8ba2e99d26b01139dab.surge.sh/playroom/#?code=N4Igxg9gJgpiBcIA8AhCAPABABwIZSgEsA7AcwF4AdEAWxiIFcbqA%2BS4zTJAYQmIBcYAlABsIYANaYA7oSj8AFlRDoAzjVwiRrdpyQBlfrkmZVeMDGXp0I3ACdSMHR05dDxqWeOXq6zdpA2F1cuABUYdH4WAHlsIUxFQlVMPhgZQkVMCRgATwAjCHsoJAB6cMigkM5gR353QQAKahoIADcnEABKTAB%2BTCbgqqRy-gTU5TA7DMIwTVZQqdJHO0wW9qgAGjGIEX5CbEwoCBhVAUx4BtKRyqru%2BExiBi0AX10hlAZ%2Bfj4ASUhiN5VUyEABePhAtgcHUBVRmfHIwCQfz4AEE7HYINJDoQ7DAwHt4dQjtIASBMCUWK9BiFbHkYCJlABZNrQ6muOTKVS4drUGEhPjcEQzCQIhrdcgsMZLEQweowJq0FnUTpUoEUmGldySIIarWeczguiMZhkzSEUjEZSLBT8ZxA4YRKKC4VjKWkGWYWn0q6Om4hVCfb7EZEAtmcVSg8GQxy8sOYOGWxEhtAQCQaOxSYx7doI2pyhVc9bK57kyl81xehk1GB1IyNXzc%2BjK3qYaj6RtQajnVsgAByEFGORrpg7WzAQpM3xHPJAqqB8agnMbsfnKWIzskovFku%2B0tldflDaLXTnIXVoaB1druHrIELTa6mAAhORyJgGMRYAAzEj0FsDVcHUiMZiHBSZplmAJMDNC0rXNG1WAAGVwOkREwMAFFwMgTi2bAIAjAkOCOE4zguH0KnLO4HieERTy4Eo9RuTUjG1UpeAEIR%2BFEcQJCCUo0HQIIQA2EBFBgOhVAQABtO8YBgCQACkIDySSAF0RNkeQFEk%2BApIAdgANgADlU54gA
[After]: https://braid-design-system--f25c0d9ba991feeb5b94027016d0d4d4553c1b13.surge.sh/playroom/#?code=N4Igxg9gJgpiBcIA8AhCAPABABwIZSgEsA7AcwF4AdEAWxiIFcbqA%2BS4zTJAYQmIBcYAlABsIYANaYA7oSj8AFlRDoAzjVwiRrdp05IAyv1yTMqvGBjL06EbgBOpGDo579Rk1PMmr1dZu0QNlc3fQAVGHR%2BFgB5bCFMRUJVTD4YGUJFTAkYAE8AIwgHKCQAegio4NC9YCd%2BD0EACmoaCAA3ZxAASkwAfi4K6LD7QlIne0xWjqgAQjLBlkx4TGIGLQBfXWquFAZ%2Bfj4ASUhiLe2zQgAvXxA7R06z7cIT8mAkY74AQXt7CGlMIj2GBgfiEPjKKB-U4gTClFibELbOz5GAiZQAWXaD0R1TkylUuA61Ee1T43BEzwkr0aPXIiwOYxEMAaMGatCx1C6CPOcJJNTqLLZU06PX6zRxoSQg0SaWU2AgqkyhCJQQMEABckSCnSBwgIlB2Dm5Ui0T5nB6y1WGz5ZQ8kmCNrtXgsNzojGYMM0o2IyhGpAU-Bc5ylJpY5MpMpljPSyNR81DZp2ewOxA%2BpwlbkV12UdycxIzeme4LeabQEAkGnsUhMoI6rwFxiafkJ9E561h8MTnFjaNqMHqjdZzemnL6mGoBhbUGoS3HIAAchB%2BJhcv2zFOADSYMAU0wHdcq7nnTB44fY4%2BpYjhyTU2n0iDRwVn6fdI-VXnp859ge4JsgAkjt0mAzOQ5CYAwxCwAAZiQ9BjuKF4hlEMrEDc8qKrWnSYF6pA%2BtQfoBkGF6YAAMrgKIiNuCi4GQMCqFu6FKnw4HYFAv50TMXZcMalSJhaKxrCIb56Laxj2o8omeFU3G8AIQj8KI4gSMEZRoOgwQgBuICKDAdCqAgADa-4wDAEgAFIQPk%2BkALpabI8gKPp8AGQA7AAbAAHNZ6xAA

### DIFF
--- a/.changeset/late-socks-double.md
+++ b/.changeset/late-socks-double.md
@@ -1,0 +1,12 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated:
+  - TooltipRenderer
+---
+
+**TooltipRenderer:** Re-evaluate position when `trigger` or `children` changes
+
+Fixes an issue where the tooltip would not re-evaluate it's position when the `trigger` or `children` prop changed while the tooltip was already open.
--- a/.changeset/late-socks-double.md
+++ b/.changeset/late-socks-double.md
@@ -9,4 +9,4 @@ updated:
 
 **TooltipRenderer:** Re-evaluate position when `trigger` or `children` changes
 
-Fixes an issue where the tooltip would not re-evaluate it's position when the `trigger` or `children` prop changed while the tooltip was already open.
+Fixes an issue where the tooltip would not re-evaluate its position when the `trigger` or `children` prop changed while the tooltip was already open.
--- a/packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
+++ b/packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
@@ -5,6 +5,7 @@ import React, {
   useEffect,
   useContext,
   useRef,
+  useLayoutEffect,
 } from 'react';
 import { usePopperTooltip } from 'react-popper-tooltip';
 import isMobile from 'is-mobile';
@@ -205,23 +206,25 @@ export const TooltipRenderer = ({
     },
   );
 
-  // If the tooltip is visible and the size or position of either the trigger
-  // or the tooltip has changed, then update the tooltip size and position.
-  if (
-    controlledVisible &&
-    update &&
-    (doesBoundingBoxNeedUpdating(triggerRef, triggerBoundingBoxRef.current) ||
-      doesBoundingBoxNeedUpdating(tooltipRef, tooltipBoundingRectRef.current))
-  ) {
-    triggerBoundingBoxRef.current = normaliseRect(
-      triggerRef?.getBoundingClientRect(),
-    );
-    tooltipBoundingRectRef.current = normaliseRect(
-      tooltipRef?.getBoundingClientRect(),
-    );
+  useLayoutEffect(() => {
+    // If the tooltip is visible and the size or position of either the trigger
+    // or the tooltip has changed, then update the tooltip size and position.
+    if (
+      controlledVisible &&
+      update &&
+      (doesBoundingBoxNeedUpdating(triggerRef, triggerBoundingBoxRef.current) ||
+        doesBoundingBoxNeedUpdating(tooltipRef, tooltipBoundingRectRef.current))
+    ) {
+      triggerBoundingBoxRef.current = normaliseRect(
+        triggerRef?.getBoundingClientRect(),
+      );
+      tooltipBoundingRectRef.current = normaliseRect(
+        tooltipRef?.getBoundingClientRect(),
+      );
 
-    update();
-  }
+      update();
+    }
+  });
 
   useEffect(() => {
     if (visible) {
--- a/packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
+++ b/packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
@@ -4,6 +4,7 @@ import React, {
   useState,
   useEffect,
   useContext,
+  useRef,
 } from 'react';
 import { usePopperTooltip } from 'react-popper-tooltip';
 import isMobile from 'is-mobile';
@@ -108,6 +109,21 @@ export interface TooltipRendererProps {
   children: (renderProps: { triggerProps: TriggerProps }) => ReactNode;
 }
 
+const normaliseRect = (domRect?: DOMRect) => ({
+  top: Math.round(domRect?.top || 0),
+  left: Math.round(domRect?.left || 0),
+  height: Math.round(domRect?.height || 0),
+  width: Math.round(domRect?.width || 0),
+});
+
+const doesBoundingBoxNeedUpdating = (
+  element: HTMLElement | null,
+  previousRect: ReturnType<typeof normaliseRect>,
+) => {
+  const currentRect = normaliseRect(element?.getBoundingClientRect());
+  return JSON.stringify(currentRect) !== JSON.stringify(previousRect);
+};
+
 export const TooltipRenderer = ({
   id,
   tooltip,
@@ -125,6 +141,12 @@ export const TooltipRenderer = ({
   const [controlledVisible, setControlledVisible] = useState(false);
   const [opacity, setOpacity] = useState<0 | 100>(0);
   const { grid, space } = useSpace();
+  const triggerBoundingBoxRef = useRef<ReturnType<typeof normaliseRect>>(
+    normaliseRect(),
+  );
+  const tooltipBoundingRectRef = useRef<ReturnType<typeof normaliseRect>>(
+    normaliseRect(),
+  );
 
   const {
     visible,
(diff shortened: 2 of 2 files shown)


.changeset/late-socks-double.md
```diff
@@ -0,0 +1,12 @@
...
+---
+
+---
+updated:
+  - TooltipRenderer
+---
+
+**TooltipRenderer:** Re-evaluate position when `trigger` or `children` changes
+
+Fixes an issue where the tooltip would not re-evaluate it's position when the `trigger` or `children` prop changed while the tooltip was already open.
```
benjervis, line 12:
```suggestion
Fixes an issue where the tooltip would not re-evaluate its position when the `trigger` or `children` prop changed while the tooltip was already open.
```

packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
```diff
@@ -175,6 +206,24 @@ export const TooltipRenderer = ({
     },
   );
 
+  // If the tooltip is visible and the size or position of either the trigger
+  // or the tooltip has changed, then update the tooltip size and position.
+  if (
```
mrm007, line 211: Should this be wrapped in a `useEffect`, [or even `useLayoutEffect`](https://react.dev/reference/react/useLayoutEffect#measuring-layout-before-the-browser-repaints-the-screen)?

packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
```diff
@@ -125,6 +141,12 @@ export const TooltipRenderer = ({
   const [controlledVisible, setControlledVisible] = useState(false);
   const [opacity, setOpacity] = useState<0 | 100>(0);
   const { grid, space } = useSpace();
+  const triggerBoundingBoxRef = useRef<ReturnType<typeof normaliseRect>>(
+    normaliseRect(),
+  );
+  const tooltipBoundingRectRef = useRef<ReturnType<typeof normaliseRect>>(
+    normaliseRect(),
+  );
```
mrm007, line 149: This means `normaliseRect()` will be called (twice) on every render.

packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
mrm007, line 1: Just thinking out loud: would this component break if I, as a consumer, were to wrap it in `React.memo`? Before or after the current change.

packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
```diff
@@ -134,12 +156,21 @@ export const TooltipRenderer = ({
...
     getArrowProps,
+    update,
   } = usePopperTooltip(
     {
       placement,
       trigger: [isMobile() ? 'click' : 'hover', 'focus'],
       visible: isStatic || controlledVisible,
-      onVisibleChange: setControlledVisible,
+      onVisibleChange: (newState) => {
+        setControlledVisible(newState);
```
mrm007, line 166: Should we move this to the end of the callback, after the refs are updated? IIRC under some circumstances `set*State` can trigger a rerender.

mrm007:
Can you please share the Playroom link you used to record the video? It would be useful to run the profiler to see what insights we can gain.

packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
michaeltaranto: I'll create a default rect statically outside the component and use that instead.

packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
michaeltaranto: Can move it to end of the callback.

michaeltaranto:
> Can you please share the Playroom link you used to record the video? It would be useful to run the profiler to see what insights we can gain.

I wanted to include it in the PR description, but surge is having another Bad Gateway kind of afternoon 😞

mrm007:
> > Can you please share the Playroom link you used to record the video? It would be useful to run the profiler to see what insights we can gain.
> 
> I wanted to include it in the PR description, but surge is having another Bad Gateway kind of afternoon 😞

Yeah, one of those days 😞. I was going to play with it locally anyway

packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
michaeltaranto: I dont think so. The issue is the tooltip or the trigger are changing and the position is not being re-evaluated. So if its memo'd and they update either of those props it should be re-evaluated.

The guard of the first check being that the tooltip is already visible, will short circuit this for irrelevant cases.

packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
michaeltaranto: Deliberately not, as not wanting to delay rendering the actual component, or have an effect that is driven by specific data changing.

The thinking being if this component is re-rendering, and the tooltip is already visible, check if it needs to be repositioned. Happy to discuss when you have time.

askoufis:
Surge is still 502ing sometimes, but it seems like it's just the root page. [The playroom page](https://braid-design-system--f25c0d9ba991feeb5b94027016d0d4d4553c1b13.surge.sh/playroom/#?code=N4XyA) and docs pages work fine for me.

packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
mrm007: In that case, this `if` block should be wrapped in `useLayoutEffect`, like so:

```ts
// If the tooltip is visible and the size or position of either the trigger
// or the tooltip has changed, then update the tooltip size and position.
useLayoutEffect(() => {
  if (
    controlledVisible &&
    update &&
    (doesBoundingBoxNeedUpdating(triggerRef, triggerBoundingBoxRef.current) ||
      doesBoundingBoxNeedUpdating(tooltipRef, tooltipBoundingRectRef.current))
  ) {
    triggerBoundingBoxRef.current = normaliseRect(
      triggerRef?.getBoundingClientRect(),
    );
    tooltipBoundingRectRef.current = normaliseRect(
      tooltipRef?.getBoundingClientRect(),
    );

    update();
  }
});
```

Notice how the dependency array is missing, which makes the code run on every render, which is equivalent to what you had before.
Since the code has a side-effect, it should be wrapped in `use*Effect`.

michaeltaranto: Haha so it was less of a question then 😄 

Let's chat more about the right direction for the requirements and test it out.

michaeltaranto:
@askoufis @mrm007 updated the description with Playroom links and wrapped logic in `useLayoutEffect`

mrm007 (approved):

PR #1382, opened 2023-10-27
Fix navigator check for BraidProvider

benjervis:
Node 21 has introduced a `navigator` object, where we could previously rely on there not being one in a Node environment.

This fixes that issue (which has been reported internally already). See changeset for full detail.

### DIFF
--- a/packages/braid-design-system/src/lib/components/BraidProvider/BraidProvider.tsx
+++ b/packages/braid-design-system/src/lib/components/BraidProvider/BraidProvider.tsx
@@ -70,6 +70,8 @@ export interface BraidProviderProps {
   children: ReactNode;
 }
 
+// Node 21 introduces an empty navigator object, so checks need a little more logic to them
+
 export const BraidProvider = ({
   theme,
   styleBody = true,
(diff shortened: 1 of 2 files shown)


packages/braid-design-system/src/lib/components/BraidProvider/BraidProvider.tsx
```diff
@@ -82,10 +82,12 @@ export const BraidProvider = ({
...
   useHideFocusRings(!(alreadyInBraidProvider || inTestProvider));
 
+  // Node 21 introduces an empty navigator object, so checks need a little more logic to them
   assert(
-    typeof navigator !== 'object' ||
-      navigator.userAgent.indexOf('jsdom') === -1 ||
-      inTestProvider,
+    inTestProvider ||
+      typeof navigator === 'undefined' ||
+      typeof navigator.userAgent === 'undefined' ||
```
jahredhope, line 89: Why a typoeof comparison over a value comparison?

```suggestion
      navigator.userAgent === undefined ||
```

askoufis (approved):

PR #1392, opened 2023-11-13
Skip NX cache for generating icons

felixhabib:
Currently, in lint on Github Actions, icons are being marked as cached incorrectly, and are not being generated. This causes the validate step to fail (see example [here](https://github.com/seek-oss/braid-design-system/actions/runs/6820110993/job/18548686830)).

This fix disables NX cache entirely for generating icon steps, causing the validate step to pass.

This PR is a temporary change to get builds passing, but a real solution will need to be implemented later.


benjervis:
@mrm007 has [fixed this]

[fixed this]: https://github.com/seek-oss/braid-design-system/pull/1393

PR #1390, opened 2023-11-10
Add optional tooltipPlacement prop to ButtonIcon

felixhabib:
Add optional `tooltipPlacement` prop to `ButtonIcon`, which defaults to `top`.
Add section on tooltip placement to docs.

This change is in response to [this request](https://seekchat.slack.com/archives/C02A7JC07B3/p1698983908288259).

PR #1393, opened 2023-11-14
Solve some task dependency issues

mrm007:
Sometimes the build mysteriously fails in CI when icons are added, e.g. #1390 has been failing on every commit.

There are two issues that have been intermittently observed since we moved to a monorepo with Nx:

1. Linting fails because Prettier expects but doesn't find the generated icon files. This is caused by a race condition where the linting task runs before the icon generation task ([example][lint]).
2. Tests fail because the generated icon files are missing ([example][test]). This is because an integration test runs the `build` script which wipes out all icon component files.

[lint]: https://github.com/seek-oss/braid-design-system/actions/runs/6856814056/job/18644723861#step:7:361
[test]: https://github.com/seek-oss/braid-design-system/actions/runs/6857128352/job/18645633637#step:8:421

This PR

- fixes the task dependency for 1. so that the generated icon files are always available for linting
- ~~skips the Nx cache for 2. so that the generated icon files are always generated when tests run~~
- runs fixture/integration tests separately so they don't interfere with the other tests

### Before

See how `lint:prettier` and `lint:eslint` can run at the same time as `generate`

braid tasks before

### After

braid tasks after

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "version": "1.0.0",
   "scripts": {
     "test": "pnpm run generate --skip-nx-cache && jest",
+    "test:fixtures": "jest --config fixtures/jest.config.js",
     "start": "nx run site:start",
     "start:playroom": "nx run site:start:playroom",
     "build": "nx run braid-design-system:build",
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "private": true,
   "version": "1.0.0",
   "scripts": {
-    "test": "pnpm run generate && jest",
+    "test": "pnpm run generate --skip-nx-cache && jest",
     "start": "nx run site:start",
     "start:playroom": "nx run site:start:playroom",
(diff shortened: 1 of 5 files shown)


benjervis (approved):

package.json
```diff
@@ -3,7 +3,8 @@
   "private": true,
   "version": "1.0.0",
   "scripts": {
-    "test": "pnpm run generate && jest",
+    "test": "pnpm run generate --skip-nx-cache && jest",
```
benjervis, line 6: Is the skip cache still required when the task graph is properly wired up?

mrm007: With the discovery of the third bullet, no, not really. Do you think we should revert it or keep it?

benjervis: If we don't really need it then might be worth reverting. Not exactly a high effort step though, so not a huge deal.
```suggestion
    "test": "pnpm run generate && jest",
```

PR #1395, opened 2023-11-15
Manually update lint config

benjervis:
Pulls in the new eslint-config-seek version, so that anyone who's accidentally putting in curly braces gets them automatically fixed


pnpm-lock.yaml
```diff
@@ -741,7 +753,29 @@ packages:
     transitivePeerDependencies:
       - supports-color
 
-  /@babel/eslint-parser@7.22.10(@babel/core@7.22.10)(eslint@8.21.0):
+  /@babel/core@7.23.3:
```
askoufis, line 756: Worth running `pnpm dedup` IMO

askoufis (approved):

PR #1401, opened 2023-11-24
Create docs-ui package with LinkableHeading component

felixhabib:
This package is to house components for building docs sites like Braid. It is intended to be used by other sites with a similar layout and UX to the Braid site.

There will be a follow up PR to remove the 'LinkableHeading' component from the `braid-design-system` package and replace all usages with the component from the `docs-ui` package.


packages/docs-ui/README.md
```diff
@@ -0,0 +1,39 @@
...
+<LinkableHeading>Section heading</LinkableHeading>;
+```
+
+#### Props
+
+| props     | value            | description                                                                                                                                                               |
+| --------- | ---------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| level     | [`HeadingLevel`] | The heading size (defaults to `3`)                                                                                                                                        |
+| component | `string`         | HTML element to render, e.g. &ldquo;h1&rdquo;, &ldquo;h2&rdquo;, etc.                                                                                                     |
+| children  | `string`         | Visible heading content, converted to slug (e.g. &ldquo;section-heading&rdquo;) to be used as hash link.<br/>Note: If the heading                                         |
```
mrm007, line 35: If the heading what?

felixhabib: Oh whoops, will add the fix in the follow up PR

packages/docs-ui/src/components/LinkableHeading/LinkableHeading.tsx
```diff
@@ -0,0 +1,48 @@
...
+    <Box display="flex">
+      <Box className={styles.linkableHeading} display="inlineBlock">
+        <Box id={slug} position="absolute" />
+        <Link href={`#${slug}`}>
+          <Heading level={level} component={component}>
+            {restProps.children}{' '}
+            <Box
+              component="span"
+              transition="fast"
+              marginLeft="xxsmall"
```
mrm007, line 37: Do we need the space _and_ the margin?

felixhabib: Hmm not sure, this is just a direct copy of the original component. What's giving space other than the margin?

mrm007: There a space character on line 33.

felixhabib: Oo yep nice catch! Have made a fix in the follow up PR.

PR #1397, opened 2023-11-22
Fix all themes being imported by the `reset` entry

mrm007:
Some changes in preparation for a Crackle release that handles all the CSS being imported by the `reset` entry. ~~I'll update the Crackle snapshot to the final version once released.~~ I'll create a new PR when the new Crackle is released.

Master:

```
packages/braid-design-system/dist/reset.mjs:
  1  import "./side-effects/lib/css/reset/index.mjs";
  2: import "./styles/lib/themes/apac/apacTheme.css.mjs";
  3: import "./styles/lib/themes/docs/docsTheme.css.mjs";
  4: import "./styles/lib/themes/seekBusiness/seekBusinessTheme.css.mjs";
  5: import "./styles/lib/themes/seekJobs/seekJobs.css.mjs";
  6: import "./styles/lib/themes/wireframe/wireframeTheme.css.mjs";

packages/braid-design-system/dist/playroom/components.mjs:
  1  import dedent from "dedent";
  2  import { jsx, jsxs } from "react/jsx-runtime";
  3  import { mapResponsiveValue, sprinkles } from "../styles/lib/css/atoms/sprinkles.css.mjs";
  4: import "../styles/lib/themes/apac/apacTheme.css.mjs";
  5: import "../styles/lib/themes/seekJobs/seekJobs.css.mjs";
  6: import "../styles/lib/themes/seekBusiness/seekBusinessTheme.css.mjs";
  7  import { wireframe } from "../styles/lib/themes/wireframe/wireframeTheme.css.mjs";
  8: import "../styles/lib/themes/docs/docsTheme.css.mjs";

packages/braid-design-system/dist/side-effects/lib/css/reset/index.mjs:
  1  import { markResetImported } from "./resetTracker.mjs";
  2: import "../../../../styles/lib/css/reset/reset.css.mjs";
  3: import "../../../../styles/lib/css/atoms/sprinkles.css.mjs";
  4  if (process.env.NODE_ENV === "development") {
  5    markResetImported();
  6  }

packages/braid-design-system/dist/side-effects/lib/themes/apac/index.mjs:
  1: import "../../../../styles/lib/themes/apac/apacTheme.css.mjs";

packages/braid-design-system/dist/side-effects/lib/themes/docs/index.mjs:
  1: import "../../../../styles/lib/themes/docs/docsTheme.css.mjs";

packages/braid-design-system/dist/side-effects/lib/themes/seekBusiness/index.mjs:
[...]

PR #1405, opened 2023-11-28
Fix theme side effects

mrm007:
This release fixes a bug where the reset module included all the tokens for all the themes. It builds upon #1397 and https://github.com/seek-oss/crackle/pull/136.

Additionally, this release contains significant compilation fixes to ensure that only styles for the imported components are included in the generated CSS. Previously, due to the default bundling behavior of Vite, styles for all Braid components were included in the generated CSS, regardless of which components were imported.

This should speed up compilation times and reduce the size of the generated CSS in consuming apps. Running `sku build` on a project created with `sku init` (without minification):

```
➜ hyperfine --setup 'pnpm install' --parameter-list BRAID_VERSION '32.14.0,fix-theme-side-effects' --prepare 'rm -rf node_modules/.cache; pnpm install braid-design-system@{BRAID_VERSION}' --command-name '{BRAID_VERSION}' 'pnpm build'

Benchmark 1: 32.14.0
  Time (mean ± σ):     23.679 s ±  1.601 s    [User: 32.462 s, System: 3.793 s]
  Range (min … max):   21.190 s … 26.328 s    10 runs

Benchmark 2: fix-theme-side-effects
  Time (mean ± σ):     21.823 s ±  1.619 s    [User: 30.121 s, System: 3.464 s]
  Range (min … max):   20.066 s … 25.553 s    10 runs

Summary
  fix-theme-side-effects ran
    1.09 ± 0.11 times faster than 32.14.0
```

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -46,8 +46,8 @@ importers:
         specifier: ^0.5.0
         version: 0.5.0
       '@crackle/cli':
-        specifier: 0.0.0-vite-5-20231123061728
-        version: 0.0.0-vite-5-20231123061728(@types/node@18.13.0)(@types/react@18.0.28)(babel-plugin-macros@3.1.0)(typescript@5.1.6)
+        specifier: ^0.12.5
+        version: 0.12.5(@types/node@18.13.0)(@types/react@18.0.28)(babel-plugin-macros@3.1.0)(typescript@5.1.6)
       '@manypkg/cli':
         specifier: ^0.19.1
         version: 0.19.1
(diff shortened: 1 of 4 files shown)


askoufis (approved):

pnpm-lock.yaml
```diff
@@ -18256,6 +18585,17 @@ packages:
       fsevents: 2.3.3
     dev: false
 
+  /tsx@4.3.0:
```
michaeltaranto, line 18588: Might be worth a dedupe, not sure these duplicates are necessary or not

mrm007: The one above is `tsx@3`

PR #1407, opened 2023-11-29
Compute tab line position within `useIsomorphicLayoutEffect` to avoid layout thrashing

mrm007:


packages/braid-design-system/src/lib/components/Tabs/Tabs.tsx
```diff
@@ -134,7 +155,10 @@ export const Tabs = (props: TabsProps) => {
       : selectedIndex;
   const selectedTabButtonEl =
     tabsContext.tabButtonElements[selectedTabIndex.toString()];
-  const activeTab = getActiveTabLinePosition(selectedTabButtonEl);
+
+  useIsomorphicLayoutEffect(() => {
+    setActiveTabPosition(getActiveTabLinePosition(selectedTabButtonEl));
```
mrm007, line 160: This is the recommended way: https://react.dev/reference/react/useLayoutEffect#measuring-layout-before-the-browser-repaints-the-screen

mrm007:
Hopefully the last diff we'll ever see: https://www.chromatic.com/build?appId=5bd0cd3e0cc977002443f476&number=5721

.changeset/healthy-news-cry.md
```diff
@@ -0,0 +1,5 @@
+---
+'braid-design-system': patch
+---
+
+DRAFT: Fix tab line position
```
michaeltaranto, line 5: We should flesh out some changesets and ensure the correct front matter is there for attribution in the docs site:

This one can be for the tabs active underline change:

```suggestion
---
- Tabs
---

**Tabs**: Improve positioning of the active underline
```


And maybe a separate one for the TooltipRenderer warning fix.

michaeltaranto (approved):

PR #1408, opened 2023-12-04
Don't bundle every syntax highlighter language

askoufis:
Same deal as the issue I noticed in metropolis. We were bundling every language supported by the syntax highlighter, instead of just the ones we need: `jsx`, `tsx` and `diff` (`jsx` and `tsx` implicitly handle `js` and `ts`).

This cuts ~200kb from the biggest chunk in the docs site bundle.

Before:
image

After:
image

The only downside of this is that we now need to explicitly register languages we use for docs/markdown formatting (the releases page). We could throw a runtime error if we find an unsupported language, that way the static render would fail during the build. Open to suggestions if it's a road we want to go down.

### DIFF
--- a/site/src/App/Markdown/Markdown.tsx
+++ b/site/src/App/Markdown/Markdown.tsx
@@ -20,7 +20,7 @@ const Code = ({
   language,
   value,
 }: {
-  language: SupportedLanguage;
+  language: SupportedLanguage | null;
   value: string;
 }) => (
   <Box paddingBottom="medium">


site/src/App/Code/Code.tsx
```diff
@@ -26,15 +26,32 @@ import { FieldOverlay } from 'braid-src/lib/components/private/FieldOverlay/Fiel
...
+} as const;
+type SupportedLanguage = keyof typeof supportedLanguageHighlighters;
+
+(Object.keys(supportedLanguageHighlighters) as SupportedLanguage[]).forEach(
+  (language) =>
+    SyntaxHighlighter.registerLanguage(
+      language,
+      supportedLanguageHighlighters[language],
+    ),
+);
```
mrm007, line 53:
```suggestion
for (const [name, language] of Object.entries({ diff, jsx, tsx })) {
  SyntaxHighlighter.registerLanguage(name, language);
}
```

site/src/App/Code/Code.tsx
```diff
@@ -26,15 +26,21 @@ import { FieldOverlay } from 'braid-src/lib/components/private/FieldOverlay/Fiel
...
+import { editorTheme } from './editorTheme';
 import { ThemedExample } from '../ThemeSetting';
 import usePlayroomScope from 'braid-src/lib/playroom/useScope';
 import { PlayroomStateProvider } from 'braid-src/lib/playroom/playroomState';
 
 type ReactElementOrString = ReactElement | string;
 
+for (const [name, language] of Object.entries({ diff, jsx, tsx })) {
+  SyntaxHighlighter.registerLanguage(name, language);
+}
```
mrm007, line 42: Or perhaps 

```suggestion
(async () => {
  for (const name of ['diff', 'jsx', 'tsx']) {
    SyntaxHighlighter.registerLanguage(
      name,
      await import(
        /* webpackChunkName: "syntax-highlighter-languages" */ `react-syntax-highlighter/dist/esm/languages/prism/${name}`
      ),
    );
  }
})();
```

mrm007:
> We could throw a runtime error if we find an unsupported language, that way the static render would fail during the build. Open to suggestions if it's a road we want to go down.

I like this idea!

site/src/App/Code/Code.tsx
askoufis: This makes a bundle for every single syntax since webpack can't tell we only want 3 of them. It _does_ work, but means we ship a whole bunch of tiny chunks.

PR #1410, opened 2023-12-05
Rename fixtures to integration tests

mrm007:
Also renamed a couple of jobs/steps ~~because I couldn't help myself~~ for consistency. I'll update branch protection with the new job name before merging.


.github/workflows/validate.yml
```diff
@@ -44,16 +44,15 @@ jobs:
       - name: Install Dependencies
         run: pnpm i
 
-      - name: Build
-        run: pnpm build
-
       - name: Lint
         run: pnpm lint
```
askoufis, line 48: I thought lint depended on build because we want to lint generated files, e.g. icons?

mrm007: I don't remember that. But if lint depends on build, the relationship should be defined in the task graph.

askoufis: Yeah good point. So the cache should handle it and not rebuild. 👍

packages/braid-design-system/codemod/bin.js
```diff
@@ -1,2 +1,2 @@
 #!/usr/bin/env node
-require('./dist/index');
+require('./dist/index'); // eslint-disable-line import/no-unresolved
```
askoufis, line 2: What changed the necessitated adding this?

mrm007: Before the change, it was running build before lint

askoufis (approved):

PR #1417, opened 2023-12-15
On docs site, invert colours of favicon if the browser is set to dark mode

felixhabib:
This is just a little extension of a hackathon project.

On the Braid docs site, the favicon should be a white Braid logo in dark mode, and a black Braid logo in light mode.


site/src/App/Seo/AppMeta.tsx
```diff
@@ -26,20 +33,21 @@ export function AppMeta() {
           rel="icon"
           type="image/png"
           sizes="16x16"
-          href="https://seek-oss.github.io/braid-design-system/favicon-16x16.png"
+          href={`https://seek-oss.github.io/braid-design-system/${faviconFileName}-16x16.png`}
```
michaeltaranto, line 36: Noticed that these werent working on the preview site, because they are pointing to production urls that dont exist yet.

Wondering these have any need to be absolute rather than path relative and can't think of a reason. Making them path relative means they work in dev and preview deployments too.

```suggestion
          href={`/${faviconFileName}-16x16.png`}
```

michaeltaranto (approved):

PR #1420, opened 2023-12-21
Autosuggest: Fix `aria-label` and `aria-labelledby` features

michaeltaranto:
**Autosuggest:** Fix `aria-label` and `aria-labelledby` features

Fixes an issue where the `aria-label` and `aria-labelledby` props provided by a consumer were being overidden internally by the `Autosuggest` component.

#### Notes

Also refactoring internally to remove the private API of `labelId` on `Field`, which is no longer used as it was only in service of `Autosuggest`. This has no impact on consumers.

### DIFF
--- a/packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.docs.tsx
+++ b/packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.docs.tsx
@@ -11,6 +11,7 @@ import {
   Alert,
   List,
   Stack,
+  Heading,
 } from '../';
 import { IconHelp, IconLanguage } from '../icons';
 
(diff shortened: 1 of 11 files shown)


packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.docs.tsx
```diff
@@ -662,6 +663,53 @@ const docs: ComponentDocs = {
...
+                { text: 'Apples', value: 1 },
+                { text: 'Bananas', value: 2 },
+                { text: 'Broccoli', value: 3 },
+                { text: 'Carrots', value: 4 },
+              ])}
+            />
+
+            <Autosuggest
+              aria-label="Hidden label for field"
+              id={id}
```
mrm007, line 699: This id clashes with the one above

mrm007 (approved):

PR #1430, opened 2024-01-22
Rating: Add `weight` support

michaeltaranto:
Provide a `weight` prop to customise the weight of the text rating alongside the stars.

**EXAMPLE USAGE:**
```jsx
<Rating rating={3} weight="strong" />
```

### DIFF
--- a/packages/braid-design-system/src/lib/components/Rating/Rating.tsx
+++ b/packages/braid-design-system/src/lib/components/Rating/Rating.tsx
@@ -50,19 +50,29 @@ const RatingStar = ({ percent, ...restProps }: RatingStar) => {
 };
 
 const ratingArr = [...Array(5)];
-export interface RatingProps {
+interface RatingBaseProps {
   rating: number;
   size?: TextProps['size'];
   /** @deprecated Use `variant="starsOnly"` instead */
   showTextRating?: boolean;
-  variant?: 'full' | 'starsOnly' | 'minimal';
   'aria-label'?: string;
   data?: TextProps['data'];
 }
+type RatingVariants = 'full' | 'starsOnly' | 'minimal';
+
+export type RatingProps = RatingBaseProps &
+  (
+    | { weight?: never; variant?: RatingVariants }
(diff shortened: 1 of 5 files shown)


benjervis (approved):
🏋️

packages/braid-design-system/src/lib/components/Rating/Rating.tsx
```diff
@@ -89,14 +103,25 @@ export const Rating = ({
...
       );
     }
-  }
 
-  const variant = variantProp || 'full';
-  const resolvedVariant =
-    showTextRating === false && !variantProp ? 'starsOnly' : variant;
+    if (typeof weight !== 'undefined' && resolvedVariant === 'starsOnly') {
+      // eslint-disable-next-line no-console
+      console.warn(
```
mrm007, line 109: Should we do an `assert` here? `console.warn` will appear in production.

mrm007: Or check for `NODE_ENV`? Prior art: https://github.com/seek-oss/braid-design-system/blob/aa43dfb03ea0032b46016feac42b885fdbd62b6f/packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.tsx#L318-L320

askoufis: I had the same thought as you, but this block is actually inside a `NODE_ENV` check. The git diff makes it a bit hard to see.

michaeltaranto: This wont appear in production, the build strips out blocks wrapped in [NODE_ENV checks].

[NODE_ENV checks]: https://github.com/seek-oss/braid-design-system/pull/1430/files/97d8466af9ae2bb28e308cbfb69af6117c145af6#diff-da536ce32f035b4a7b2b7603012bfec9838413ca07dd368be41a5cbd48d99b06R90

PR #1436, opened 2024-02-01
Fix routing for Icon page tabs

felixhabib:
Props are release paths were not included in the static render for icons, causing links such as https://seek-oss.github.io/braid-design-system/components/IconDownload/props to return a 404.
This change includes the correct paths in the static render.

Also, retire usage of `flatten` in favour of `.flat()` and `.flatMap`.

### DIFF
--- a/site/sku.routes.js
+++ b/site/sku.routes.js
@@ -1,6 +1,5 @@
 const fs = require('fs');
 const path = require('path');
-const flatten = require('lodash/flatten');
 
 const extractExports = require('./extractExports');
 const undocumentedExports = require('./src/undocumentedExports.json');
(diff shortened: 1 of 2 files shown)


site/sku.routes.js
```diff
@@ -61,5 +61,11 @@ module.exports = [
...
     ]),
   ),
-  ...iconNames.map((name) => ({ route: `/components/${name}`, name })),
+  ...flatten(
+    iconNames.map((name) => [
+      { route: `/components/${name}`, name },
+      { route: `/components/${name}/props` },
+      { route: `/components/${name}/releases` },
+    ]),
+  ),
```
askoufis, line 70: I think it's safe to use the native [`.flat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) nowadays (worth doing for the other usages in this file too).

benjervis: Probably even just [`flatMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap)

askoufis (approved):
👏

site/sku.routes.js
```diff
@@ -1,6 +1,5 @@
 const fs = require('fs');
 const path = require('path');
-const flatten = require('lodash/flatten');
```
michaeltaranto, line 3: Can this dependency be removed now?

felixhabib: This is the full `lodash` dependency and we still use other functions in other places

michaeltaranto (approved):

PR #1447, opened 2024-02-09
Tab: Remove cropping of the icon slot

michaeltaranto:
Previously the `icon` slot on a `Tab` was cropped on the left to improve alignment with the active tab indicator.
For most icons in a `Tab`, this was subtle polish, but for others it had the undesirable side effect of clipping the side of the icon.

Removing the cropping until we have a better solution for trimming whitespace around icons.

Before

### DIFF
--- a/.changeset/cool-tomatoes-attend.md
+++ b/.changeset/cool-tomatoes-attend.md
@@ -1,0 +1,15 @@
+---
+'braid-design-system': patch
+---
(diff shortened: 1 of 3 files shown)


askoufis (approved):

mrm007 (approved):
One finds shrinkflation in the most unexpected places

PR #1448, opened 2024-02-09
Import playroom exports from actual entries

mrm007:
Turns out [there was a bug](https://github.com/seek-oss/crackle/pull/184) which we never encountered because we're importing from `braid-src` in the site


site/src/playroom.components.ts
```diff
@@ -1,3 +1,3 @@
 import 'braid-src/entries/reset';
```
michaeltaranto, line 1: Is there a reason to not use the public reset too?

mrm007: Yes, but the explanation is long and complicated. @askoufis has the results from the Playroom dependency analysis. TLDR: Playroom needs changes to ensure the same reset (not ESM and CJS) is imported only once per entry point.

michaeltaranto: Thanks sir 🙏

michaeltaranto (approved):

PR #1451, opened 2024-02-15
Text: Remove descender crop on truncated text

adenj:
A combination of webkit-box, line-height and font-size can lead to instances where descender in the last line are cropped. 
Screenshot 2024-02-16 at 10 43 06 am

This change focuses on preventing any potential cropping on text nodes

### DIFF
--- a/.changeset/itchy-mayflies-allow.md
+++ b/.changeset/itchy-mayflies-allow.md
@@ -9,3 +9,5 @@ updated:
 ---
 
 **Text, Heading:** Fix `maxLines` cropping of decending characters
+
+Fixes a bug when using -webkit-box, where the descender on the last line of text could be cropped based on the combination of line height and font size.
--- a/.changeset/itchy-mayflies-allow.md
+++ b/.changeset/itchy-mayflies-allow.md
@@ -1,0 +1,11 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated:
+  - Text
+  - Heading
+---
+
+**Text, Heading:** Fix `maxLines` cropping of decending characters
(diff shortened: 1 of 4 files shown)


packages/braid-design-system/src/lib/components/private/MaxLines/MaxLines.css.ts
```diff
@@ -1,6 +1,17 @@
 import { createVar, style } from '@vanilla-extract/css';
 import { atoms } from '../../../../entries/css';
 
+/*
+Fixes a bug when using -webkit-box, where the descender of the last line
```
michaeltaranto, line 5:
```suggestion
Fixes a bug when using -webkit-box, where the descender on the last line
```

packages/braid-design-system/src/lib/components/private/MaxLines/MaxLines.css.ts
```diff
@@ -10,6 +21,11 @@ export const base = style([
   {
     textOverflow: 'ellipsis',
     whiteSpace: 'nowrap',
+    selectors: {
+      [`${descenderCropFixForWebkitBox} &`]: {
```
michaeltaranto, line 25: Not that we allow nesting our typography components, but maybe a more full proof way is always ensure the `base` class is a direct child of the parent.
```suggestion
      [`${descenderCropFixForWebkitBox} > &`]: {
```

.changeset/itchy-mayflies-allow.md
```diff
@@ -0,0 +1,11 @@
...
+'braid-design-system': patch
+---
+
+---
+updated:
+  - Text
+  - Heading
+---
+
+**Text, Heading:** Fix `maxLines` cropping of decending characters
```
michaeltaranto, line 11: We encourage adding a bit of context about the fix after the summary for those wanting to know more, something like this maybe:

```suggestion
**Text, Heading:** Fix `maxLines` cropping of decending characters

Fixes a bug when using -webkit-box, where the descender on the last line of text could be cropped based on the combination of line height and font size.
```

michaeltaranto (approved):

PR #1481, opened 2024-04-04
New and updated icons

michaeltaranto:
Adding a few new icons to the suite:
  - IconAI
  - IconCareer
  - IconCategory
  - IconExperience
  - IconGift
  - IconGlobe
  - IconRocket
  - IconSentiment
  - IconSkills

And updating some semantic icons:
  - IconCritical
  - IconLanguage


.changeset/moody-readers-fly.md
```diff
@@ -0,0 +1,20 @@
...
+
+Update semantic icon assets.
+
+`IconCritical`: Move from circle to diamond outline. Increase the visual distinction from `IconInfo`.
+`IconLanguage`: Move from globe to characters. Better represents the concept of language. Previous asset available as `IconGlobe`.
+
+
+**MIGRATION GUIDE**
+
+As the above are updates to semantics icons, consumers are unaffected if their usage follows the icons semantic intent.
```
askoufis, line 19:
```suggestion
As the above are updates to semantics icons, consumers are unaffected if their usage follows the icon's semantic intent.
```

packages/braid-design-system/src/lib/components/icons/IconSentiment/IconSentiment.tsx
```diff
@@ -0,0 +1,28 @@
...
+  neutral: IconSentimentSvg,
+};
+
+export const IconSentiment = ({
+  feeling = 'neutral',
+  ...props
+}: IconSentimentProps) => {
+  const iconProps = useIcon(props);
+
+  return <Box component={feelingToIcon[feeling]} {...iconProps} />;
```
askoufis, line 27: Should we fall back to a sentiment if an invalid `feeling` is provided? Feel like it's worth it to prevent an unexpected empty icon.

```suggestion
  return <Box component={feelingToIcon[feeling] || IconSentimentSvg} {...iconProps} />;
```

[`IconChevron`](https://github.com/seek-oss/braid-design-system/blob/fb60dfcbfba5699ddd4db0c78ef21092f448a370/packages/braid-design-system/src/lib/components/icons/IconChevron/IconChevron.tsx) is the only other icon that has a non-boolean prop that affects the icon appearance, but it implements that using CSS, so an invalid input will just result in the `down` appearance.

michaeltaranto, line 27: Yeah i dont mind that, maybe prefer defaulting to the value from the lookup instead of the SVG directly.

```suggestion
  return (
    <Box
      component={feelingToIcon[feeling] || feelingToIcon.neutral}
      {...iconProps}
    />
  );
```

What do you think of this?

askoufis: Ah yep, that makes more sense.

askoufis (approved):

PR #1489, opened 2024-05-09
Fix Animation handling in Modal component

felixhabib:
While `Modal` currently has an exit transition, it looks like a fall-through in a switch statement has prevented this from ever showing.

This change:
- Prevents fall through behaviour in the switch statement
- Smooths out the behaviour of the `Drawer` and `Dialog` exit animations
- Refactors animation style logic
- Fixes typo in `horizontalTransition`
- Updates `Dialog` docs page to add seperate state for the Dialog's width and open status

### DIFF
--- a/.changeset/calm-fireants-exist.md
+++ b/.changeset/calm-fireants-exist.md
@@ -1,0 +1,10 @@
+---
+'braid-design-system': major
+---
+
+---
+updated:
+  - Drawer
+---
+
+Fix minor bug which prevented `Drawer` exit animation.
--- a/.changeset/calm-fireants-exist.md
+++ b/.changeset/calm-fireants-exist.md
@@ -1,5 +1,5 @@
 ---
-'braid-design-system': major
+'braid-design-system': patch
 ---
 
 ---
--- a/.changeset/calm-fireants-exist.md
+++ b/.changeset/calm-fireants-exist.md
@@ -7,4 +7,4 @@ updated:
   - Drawer
 ---
 
-Fix minor bug which prevented `Drawer` exit animation.
+Fix minor bug which disabled `Drawer` and `Dialog` exit animations.
--- a/.changeset/calm-fireants-exist.md
+++ b/.changeset/calm-fireants-exist.md
@@ -7,4 +7,4 @@ updated:
   - Drawer
 ---
 
-Fix minor bug which prevented the `Drawer` and `Dialog` exit animations from occurring.
+Fix minor bug which prevented the `Drawer` exit animation from occurring.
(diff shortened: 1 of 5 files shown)


packages/braid-design-system/src/lib/components/private/Modal/Modal.tsx
```diff
@@ -221,13 +226,12 @@ export const Modal = ({
...
           inset={0}
           zIndex="modalBackdrop"
-          transition={position === 'center' ? 'fast' : undefined}
           opacity={state !== OPEN ? 0 : undefined}
           pointerEvents={state === CLOSING ? 'none' : undefined}
           className={[
-            styles.backdrop,
-            position === 'left' ||
-              (position === 'right' && styles.horiztontalTransition),
+            (state === OPEN || state === OPENING) && styles.backdrop,
```
michaeltaranto, line 232: Not sure the backdrop styling should be conditional. I think the possible improvement here is triggering the opacity change (line 229 above) during the opening phase. Thoughts?

packages/braid-design-system/src/lib/components/private/Modal/Modal.tsx
```diff
@@ -236,15 +240,20 @@ export const Modal = ({
...
           zIndex="modal"
           pointerEvents="none"
-          transition="fast"
-          opacity={state !== OPEN ? 0 : undefined}
+          transition={
+            // Disable exit transition for center modals
+            position === 'center' && state !== OPEN && state !== OPENING
+              ? undefined
+              : 'fast'
+          }
```
michaeltaranto, line 248: Let's pair on this one. Notice that we are still transitioning the backdrop on center exits, might be an opportunity to improve some things.

.changeset/calm-fireants-exist.md
```diff
@@ -0,0 +1,10 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated:
+  - Drawer
+---
+
+Fix minor bug which disabled `Drawer` and `Dialog` exit animations.
```
michaeltaranto, line 10:
```suggestion
Fix minor bug which prevented the `Drawer` and `Dialog` exit animations from occurring.
```

packages/braid-design-system/src/lib/components/private/Modal/Modal.css.ts
```diff
@@ -34,56 +32,43 @@ const reducedMotion = style({
...
-  left: [
-    reducedMotion,
-    style(
-      responsiveStyle({
-        mobile: { opacity: 1, transform: 'translateX(-110%)' },
-        tablet: { opacity: 0, transform: 'translateX(-10px)' },
-      }),
-    ),
-  ],
+  center: centerAnimation,
```
michaeltaranto, line 65: Might be worth adding a changeset for the introduction of the `Dialog` exit animation.

michaeltaranto (approved):

PR #1497, opened 2024-05-23
Tag: Add `small` size

michaeltaranto:
Introduce a new `small` size for `Tag` component. This size sits alongside the existing `standard` size, which is the default.

**EXAMPLE USAGE:**
```jsx
<Tag size="small">
  Tag
</Tag>
```

### DIFF
--- a/packages/braid-design-system/src/lib/components/Tag/Tag.tsx
+++ b/packages/braid-design-system/src/lib/components/Tag/Tag.tsx
@@ -8,18 +8,28 @@ import buildDataAttributes, {
   type DataAttributeMap,
 } from '../private/buildDataAttributes';
 import type { AllOrNone } from '../private/AllOrNone';
+import type { Space } from '../../css/atoms/atoms';
 import * as styles from './Tag.css';
+
+export const tagSizes = ['small', 'standard'] as const;
 
 export type TagProps = {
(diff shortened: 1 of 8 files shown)


packages/braid-design-system/src/lib/components/Tag/Tag.tsx
```diff
@@ -47,27 +57,31 @@ export const Tag = ({
...
+        paddingX={paddingXForSize[size]}
+        paddingRight={onClear ? 'xsmall' : paddingXForSize[size]}
         borderRadius="full"
       >
         <Box minWidth={0} title={children}>
-          <Text baseline={false} maxLines={1} icon={icon}>
-            {children}
+          <Text size={size} baseline={false} maxLines={1}>
+            {icon} {children}
           </Text>
```
askoufis, line 68: What's the reason to not use the `icon` slot here?

michaeltaranto: We decided the slot layout was not necessary and resulted in the content feeling too loose particularly inside smaller versions. Opting for the inline approach which ensure better relationship between icon and label particularly when the clear button is shown.

askoufis (approved):

PR #1504, opened 2024-06-04
PageBlock: Add `small` and `full` width options

michaeltaranto:
Add `small` to available `width` options of `PageBlock` to support narrower max width for page content.

Also introducing `full` as a `width` option to enable full width content, while still maintaining consistent screen gutters.

**EXAMPLE USAGE:**
```jsx
<PageBlock width="small">
  ...
</PageBlock>
```

### DIFF
--- a/packages/braid-design-system/src/lib/components/PageBlock/PageBlock.gallery.tsx
+++ b/packages/braid-design-system/src/lib/components/PageBlock/PageBlock.gallery.tsx
@@ -6,6 +6,15 @@ import source from '@braid-design-system/source.macro';
 
 export const galleryItems: ComponentExample[] = [
   {
+    label: 'Small width',
+    Example: () =>
+      source(
+        <PageBlock width="small">
+          <Placeholder height={100} />
+        </PageBlock>,
+      ),
+  },
+  {
     label: 'Medium width',
     Example: () =>
       source(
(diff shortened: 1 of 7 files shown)


packages/braid-design-system/src/lib/components/PageBlock/PageBlock.gallery.tsx
```diff
@@ -5,6 +5,15 @@ import { PageBlock } from '../';
...
 export const galleryItems: ComponentExample[] = [
+  {
+    label: 'Small width',
+    Example: () =>
+      source(
+        <PageBlock width="small">
+          <Placeholder height={100} />
+        </PageBlock>,
+      ),
+  },
```
felixhabib, line 16: Can't really see the different sizes here because the gallery frame is too narrow

felixhabib (approved):

packages/braid-design-system/src/lib/components/PageBlock/PageBlock.gallery.tsx
michaeltaranto: Yeah its the same for `ContentBlock` too. I wonder if we should consider removing them from Gallery entirely. They arent exactly a visual pattern people browse for in that way

felixhabib: Mm yep could do. Unless people like to use Gallery to grab code snippets for these components, but I'm not sure that's happening

michaeltaranto: We'll do a Gallery sweep separately. For now we'll proceed with following the pattern in there.

PR #1509, opened 2024-06-07
Add `togglePosition` prop to `Toggle` component

felixhabib:
Add `togglePosition` prop to `Toggle` component:

* `togglePosition="leading"` - positions the toggle before the label text
* `togglePosition="trailing"` - positions the toggle after the label text

If no `togglePosition` prop is provided:

* If `align` is set to `left`,  `togglePosition` is set to `leading`
* If `align` is not set to `left`, `togglePosition` is set to `trailing`

This is to preserve the current behaviour with `align` to avoid a breaking change

See changeset for more details

### DIFF
--- a/packages/braid-design-system/src/lib/components/Toggle/Toggle.docs.tsx
+++ b/packages/braid-design-system/src/lib/components/Toggle/Toggle.docs.tsx
@@ -59,6 +59,34 @@ const docs: ComponentDocs = {
         ),
     },
     {
+      label: 'Toggle position',
+      description: (
+        <Text>
+          Toggles can be positioned before or after the label via the{' '}
+          <Strong>togglePosition</Strong> prop.
+        </Text>
+      ),
+      Example: ({ id, getState, toggleState }) =>
+        source(
+          <Stack space="large" dividers>
+            <Toggle
+              label="Leading"
+              id={`${id}_leading`}
+              on={getState('toggleLeading')}
+              onChange={() => toggleState('toggleLeading')}
+              togglePosition="leading"
+            />
+            <Toggle
+              label="Trailing"
+              id={`${id}_trailing`}
+              on={getState('toggleTrailing')}
+              onChange={() => toggleState('toggleTrailing')}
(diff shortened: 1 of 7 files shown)


packages/braid-design-system/src/lib/components/Toggle/Toggle.docs.tsx
```diff
@@ -58,6 +58,34 @@ const docs: ComponentDocs = {
           </Stack>,
         ),
     },
+    {
+      label: 'Toggle position',
+      description: (
+        <Text>
+          Toggles can be positioned before or after the label via the{' '}
+          <Strong>togglePosition</Strong> prop.
```
michaeltaranto, line 66: As above, it would be good to reword to incorporate the names of the values here. 

Maybe something like this:

> By default, the position of the toggle relative to the label text will be determined by its `<Strong>align</Strong>` prop. This can be overridden by setting the `<Strong>togglePosition</Strong` prop as either `<Strong>leading</Strong>` or `<Strong>trailing</Strong>`

Thoughts?

packages/braid-design-system/src/lib/components/Toggle/Toggle.tsx
```diff
@@ -43,6 +44,7 @@ export const Toggle = forwardRef<HTMLInputElement, ToggleProps>(
       onChange,
       label,
       align = 'left',
+      togglePosition = align === 'left' ? 'leading' : 'trailing',
```
michaeltaranto, line 47: Not sure about using a ternary as a default value like this, and this may be the reason for the Playroom note below.

Given it is an override for the layout (previously controlled solely by `align`), knowing if a value has been passed in explicitly versus a default would be my recommended approach.

felixhabib: I've moved the default setting logic into the component itself, but not 100% clear how much simpler this approach is. The Playroom issue is actually present in prod with the `algin` prop as well.

askoufis (approved):

PR #1511, opened 2024-06-13
On docs site, add 'wide' width for Gallery Component sections. Use for `PageBlock` and `ContentBlock`

felixhabib:
This change makes it easier to visually compare the layout options of wider components in Gallery on the docs site

* Add an optional `itemWidth` prop for galleryComponents, with a `wide` option (`1500px`). This width is configured in the `.gallery.tsx` file for each component, similar to `screenshotWidths` in `screenshots.tsx` files
* Update all `gallery.tsx` files to conform to the new `GalleryComponent` type. The diffs are pretty unhelpful for these, but they are all essentially a 4 line change
* Use `wide` option for `PageBlock` and `ContentBlock` components

Currently, only `PageBlock` and `ContentBlock` really benefit from a wide view, so it makes more sense to have this configurable per component, rather than applying `wide` to all Layout components

Want to keep these components in Gallery to help with discoverability. It's nice to have visual comparison for the different widths that you can't see on the standard docs page

Should explore separating all Layout components into their own Gallery section as a seperate piece of work

### DIFF
--- a/site/src/App/routes/gallery/Gallery.tsx
+++ b/site/src/App/routes/gallery/Gallery.tsx
@@ -80,13 +80,10 @@ const DefaultContainer = ({ children }: { children: ReactNode }) => (
 
 const COLUMN_SIZE = 4;
 
-const galleryComponents = allGalleryComponents.map(
-  ({ itemWidth, examples, ...rest }) => ({
-    ...rest,
-    itemWidth,
-    examples: chunk(examples, COLUMN_SIZE),
-  }),
-);
+const galleryComponents = allGalleryComponents.map(({ examples, ...rest }) => ({
+  ...rest,
+  examples: chunk(examples, COLUMN_SIZE),
+}));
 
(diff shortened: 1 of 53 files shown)


site/src/App/routes/gallery/Gallery.tsx
```diff
@@ -80,41 +80,47 @@ const DefaultContainer = ({ children }: { children: ReactNode }) => (
...
 const COLUMN_SIZE = 4;
 
-const galleryComponents = allGalleryComponents.map(({ examples, ...rest }) => ({
-  ...rest,
-  examples: chunk(examples, COLUMN_SIZE),
-}));
+const galleryComponents = allGalleryComponents.map(
+  ({ itemWidth, examples, ...rest }) => ({
+    ...rest,
+    itemWidth,
```
askoufis, line 86: Won't `itemWidth` be included in `rest` if you just don't split it out?

askoufis (approved):

PR #1514, opened 2024-06-17
On docs site, update Gallery grid alignment and `Icon` visual treatment

felixhabib:
For Gallery grids:

* Ensures the last row in each Gallery grid is left aligned, for more consistent spacing between components

For Gallery icons:

* Adds a border around each icon in the Icons section so that "Copy code" button is more obviously aligned to the icon frame.
* Increases space between icon heading and icon preview
* Moves the "Copy code" button below the icon preview for better visual balance

This change is primarily for the 'SEEK Jobs' theme, where the lack of borders makes it unclear what the icon title and "Copy code" button are aligned to and may appear uneven


site/src/App/routes/gallery/Gallery.tsx
```diff
@@ -136,43 +137,52 @@ const getRowsFor = memoize((type: SetName) => {
...
-        {value ? (
-          <ThemedExample background={background || undefined}>
-            <Container>
-              <Box height="full" width="full" style={{ cursor: 'auto' }}>
-                {value}
-              </Box>
-            </Container>
-          </ThemedExample>
-        ) : null}
+        {isAnIcon ? children.slice().reverse() : children}
```
askoufis, line 185: Not sure if the `slice` is necessary here. `children` is recreated every render, so it _should_ be unique for every example, so it shouldn't matter if you `reverse` the original array.

site/src/App/routes/gallery/Gallery.tsx
```diff
@@ -136,43 +137,52 @@ const getRowsFor = memoize((type: SetName) => {
 interface RenderExampleProps {
   id: string;
   example: ComponentExample;
+  isAnIcon: boolean;
```
askoufis, line 140: Even though this was an existing variable before your change, while you're here IMO it should be renamed to `isIcon`:
```suggestion
  isIcon: boolean;
```

site/src/App/routes/gallery/Gallery.tsx
```diff
@@ -346,28 +357,35 @@ interface StageProps {
...
+        <Box padding="xxlarge">
+          <Heading component="h2" level="1">
+            <span style={{ fontSize: '3em' }}>{title}</span>
+          </Heading>
+        </Box>
+        <Box>
+          {items.map((row, index) => (
+            <Columns space="none" key={index}>
+              {range(rowLength).map((item) => (
+                <Column key={item} width={row[item] ? 'content' : undefined}>
```
askoufis, line 376: Won't this `key` potentially clash with the `key` 2 lines above? Might be worth prefixing with `column` and `row` or something.

askoufis (approved):

PR #1518, opened 2024-06-18
Toggle: fix label text vertical centering for small size

felixhabib:
Label text can be vertically misaligned in some scenarios - [Playroom](https://seek-oss.github.io/braid-design-system/playroom/#?code=N4Igxg9gJgpiBcIA8AhCAPABAIwIZgGsBzAJwgFcA7KAXgB1wSBLAFybFwBsAZJogCxYMAfHUqZMSAJKVOTSjExc%2BlAJr1wMSixgkGmAM4AHfDA3p0BgLZdOIsRIlIAKhCJFOizrmwxOG7h8-fQMmAC8zBmtbfQB6UXFHFxh0FkNwyJBozjsQYWzOTB1UpFjnFJYEp1iZOQUE0rR0BJAAGhAWfhgrGAMEAG0QXBMwNqyYGAIAKQhsPoBddoB3JihOvvh%2BgGYAJgAGeYBfIA)
With a small size `Toggle`, the text sits slightly too high

This change resolves these issues

### DIFF
--- a/packages/braid-design-system/src/lib/components/Toggle/Toggle.screenshots.tsx
+++ b/packages/braid-design-system/src/lib/components/Toggle/Toggle.screenshots.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import type { ComponentScreenshot } from 'site/types';
-import { Toggle, Box, Tiles, Inline } from '../';
+import { Toggle, Box, Tiles, Inline, Text } from '../';
 import { BackgroundContrastTest } from '../../utils/BackgroundContrastTest';
 import { debugTouchableAttrForDataProp } from '../private/touchable/debugTouchable';
 
@@ -188,7 +188,36 @@ export const screenshots: ComponentScreenshot = {
         />
       ),
     },
-
+    {
+      label: 'With Inline text (standard size)',
+      Container: ({ children }) => (
+        <div style={{ maxWidth: '300px' }}>{children}</div>
+      ),
+      Example: ({ id, handler }) => (
+        <Inline space="xsmall" alignY="center">
+          <Toggle on={true} label="Toggle" id={id} onChange={handler} />
+          <Text>Inline text</Text>
+        </Inline>
+      ),
+    },
+    {
+      label: 'With Inline text (small size)',
+      Container: ({ children }) => (
+        <div style={{ maxWidth: '300px' }}>{children}</div>
+      ),
+      Example: ({ id, handler }) => (
+        <Inline space="xsmall" alignY="center">
+          <Toggle
+            on={true}
+            label="Toggle"
+            id={id}
+            onChange={handler}
+            size="small"
+          />
+          <Text size="small">Inline text</Text>
+        </Inline>
+      ),
+    },
     {
       label: 'Virtual touch target',
       Example: ({ id, handler }) => (
--- a/packages/braid-design-system/src/lib/components/Toggle/Toggle.tsx
+++ b/packages/braid-design-system/src/lib/components/Toggle/Toggle.tsx
@@ -170,6 +170,9 @@ export const Toggle = forwardRef<HTMLInputElement, ToggleProps>(
           paddingRight={
             appliedTogglePosition === 'leading' ? undefined : 'xsmall'
           }
+          display="flex"
+          flexDirection="column"
+          justifyContent="center"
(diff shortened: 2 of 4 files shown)


packages/braid-design-system/src/lib/components/Toggle/Toggle.screenshots.tsx
```diff
@@ -188,7 +188,36 @@ export const screenshots: ComponentScreenshot = {
...
+            on={true}
+            label="Toggle"
+            id={id}
+            onChange={handler}
+            size="small"
+          />
+          <Text size="small">Inline text</Text>
+        </Inline>
+      ),
+    },
```
askoufis, line 220: These screenshot tests don't look like how I'd expect them to look.

image

I would've though the small toggle screenshot above would've changed, so there shouldn't be a need to add these new ones.

packages/braid-design-system/src/lib/components/Toggle/Toggle.tsx
```diff
@@ -170,6 +170,9 @@ export const Toggle = forwardRef<HTMLInputElement, ToggleProps>(
           paddingRight={
             appliedTogglePosition === 'leading' ? undefined : 'xsmall'
           }
+          display="flex"
+          flexDirection="column"
+          justifyContent="center"
```
askoufis, line 175: You could do `alignItems: center` instead and then you wouldn't need to change the flex direction. Not a huge deal, but it communicates the intent that you're trying to vertically center something that's being laid out horizontally, as opposed to spacing out something that's laid out vertically (current behaviour).

Feel like there might be an opportunity to abstract something here to ensure consistency with other inline labels like the ones used for `Checkbox` and `RadioGroup`. Not sure though. Regardless it's worth tackling separately.

packages/braid-design-system/src/lib/components/Toggle/Toggle.screenshots.tsx
felixhabib: image
Hmm, I'm not seeing the same results
Yep the small toggle screenshot has also changed slightly, but even checking that this alignment issue was still missed. It might be worth adding this test because it's easy to see if the text here is misaligned. With the `Toggle` on its own, it can be hard to notice the text being slightly off centre if you aren't paying close attention. Thoughts?

packages/braid-design-system/src/lib/components/Toggle/Toggle.tsx
felixhabib: Good point. I mentally picture it about as two columns, but this is simpler

packages/braid-design-system/src/lib/components/Toggle/Toggle.screenshots.tsx
askoufis: Weird. Maybe it was because I was looking in the chromatic storybook specifically.

It _is_ hard to spot, but IMO one of the main values of these screenshots is catching regressions, rather than being a source of truth for layout or to show usage examples. The small toggle screenshot will catch any regressions in this alignment in the future, so IMO these extras aren't that useful.

packages/braid-design-system/src/lib/components/Toggle/Toggle.tsx
askoufis: I don't think this worked as intended, the screenshots have changed 😅

packages/braid-design-system/src/lib/components/Toggle/Toggle.screenshots.tsx
felixhabib: True, but I suspect this came about when adding the SEEK Jobs theme as the error is not in older themes. I guess having this test makes it easier to ensure alignment if a new theme is added in future. What do you think?

packages/braid-design-system/src/lib/components/Toggle/Toggle.tsx
felixhabib: Hmm ok, I think it's because the default value for `justifyContent` is `flexStart`, which moves all the children horizontally to the left of the container (in a row). We can add `justifyContent: flexEnd` to fix this, but now it's back to three lines anyway. Probably will just revert this change because the defaults happen to work for the case of column and it makes sense to me to think of the vertical alignment as a column

askoufis (approved):

PR #1529, opened 2024-07-07
Update pnpm

michaeltaranto:
Upgrade to latest pnpm.


askoufis (approved):
With [a recent manypkg version](https://github.com/Thinkmill/manypkg/releases/tag/%40manypkg%2Fcli%400.21.3), it's worth moving the overrides to regular workspace deps where they're needed. Can be a separate PR though.

PR #1530, opened 2024-07-08
site: Move 'Layout' components into separate section in Gallery

felixhabib:
Layout components are moved into their own section in Gallery, making them more discoverable and adding some more hierarchy to the components view

### DIFF
--- a/site/src/App/routes/gallery/Gallery.tsx
+++ b/site/src/App/routes/gallery/Gallery.tsx
@@ -80,14 +80,24 @@ const DefaultContainer = ({ children }: { children: ReactNode }) => (
 
 const COLUMN_SIZE = 4;
 
-const galleryComponents = allGalleryComponents.map(({ examples, ...rest }) => ({
-  ...rest,
-  examples: chunk(examples, COLUMN_SIZE),
-}));
-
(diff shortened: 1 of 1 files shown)


site/src/App/routes/gallery/Gallery.tsx
```diff
@@ -119,16 +129,21 @@ export const galleryIcons: typeof galleryComponents = Object.keys(icons).map(
...
+  const items = componentTypeMap[type];
 
   const ratio = Math.max(
-    (window.innerWidth / window.innerHeight) *
-      (type === 'components' ? 0.65 : 0.5),
+    (window.innerWidth / window.innerHeight) * (type === 'icons' ? 0.5 : 0.65),
     1,
   );
-  const rowLength = Math.floor(Math.sqrt(items.length) * ratio);
+  const rowLength = Math.ceil(Math.sqrt(items.length) * ratio);
```
felixhabib, line 146: `ceil` seems to work better to for all sections. Using `floor` here would have `rowLength` a bit too short for the Layout section

michaeltaranto (approved):
Works better than we imagined! Nice work 👏

PR #1531, opened 2024-07-08
site: Switch default theme from 'apac' to 'seekJobs'

felixhabib:
Switching the default docs theme to 'seekJobs', as it is now the standard flagship theme


michaeltaranto (commented):
🚀

michaeltaranto (approved):

PR #1534, opened 2024-07-15
Migrate to `sku` version 13, and fix minor bug in useToast for React 18.3.0

felixhabib:
Migrating to sku 13, involving a few supporting changes:

* Handles Storybook configuration, replacing `dist-storybook` uses with the default `storybook-static`
* Updates component contracts to support new HTML properties
* Fixes minor bug in `useToast`. Remove `key` as a provided value in one use of `addToast` to prevent [warning in React 18.3.0](https://github.com/facebook/react/blob/5dcf3ca8d45a276a8b4cee0cedd234967661ca35/CHANGELOG.md#react) (Warn when spreading key)


.changeset/strange-otters-carry.md
```diff
@@ -0,0 +1,10 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated:
+  - Toast
```
michaeltaranto, line 7: This should match the public API, e.g. `useToast`

```suggestion
  - useToast
```

packages/braid-design-system/.eslintignore
```diff
@@ -5,10 +5,11 @@
 
 node_modules/
 codemod/dist/
-dist-storybook/
+storybook-static/
 src/lib/components/icons/*/*Svg.tsx
 src/lib/playroom/snippets/*
 README.md
+storybook-static
```
michaeltaranto, line 12: Appear to have duplicate entries here (and in `gitignore` too)

site/playroom.config.js
```diff
@@ -1,4 +1,6 @@
 const path = require('path');
+// Todo - shouldn't need eslint disable
+// eslint-disable-next-line import/no-unresolved
```
michaeltaranto, line 3: Any ideas on what is happening here?

felixhabib: Not sure currently, we had the same issue in Metropolis. Even though the IDE is able to find the plugin it shows a lint error

askoufis: I think it's an `eslint-plugin-import` issue https://github.com/import-js/eslint-plugin-import/issues/1810.

Supposedly using `eslint-import-resolver-typescript` fixes it, but we already use it in `eslint-config-seek`, so I'm not sure why it's not working. Could be because this is in a `.js` file, so the typescript resolver doesn't get used.

Not worth fussing over IMO.

askoufis (approved):

PR #1542, opened 2024-07-26
changesets: Ignore internal packages and site

michaeltaranto:
Noticed these packages appearing in the Version Packages PR when they should be ignored by Changesets


askoufis (approved):
IMO it's not a huge deal if these _do_ end up in version packages, but happy to ignore them.

PR #1547, opened 2024-07-30
Badge: Enable usage inside typographic components

michaeltaranto:
A `Badge` can now be nested inside typographic components, e.g. `Text` and `Heading`, as an inline element alongside text.
Previously a `Badge` had to be aligned against text using an `Inline` component, which would result in the `Badge` dropping below the text when the copy wrapped.

**EXAMPLE USAGE:**
```jsx
<Text>
  Lorem ipsum velit in <Badge>amet</Badge>.
</Text>
```

### DIFF
--- a/.changeset/fast-kangaroos-switch.md
+++ b/.changeset/fast-kangaroos-switch.md
@@ -1,0 +1,29 @@
+---
+'braid-design-system': minor
+---
+
+---
+new:
+  - Column
+---
+
+**Column:** Add support for hide above/below breakpoint
+
+Introduce new `hideAbove` and `hideBelow` props on column for responsively hiding columns across breakpoint.
+
+These props can be used either separately or in combination to optimise content display across different screen sizes.
+
+**EXAMPLE USAGE:**
+```jsx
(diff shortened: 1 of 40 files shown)


askoufis (commented):
Is it worth updating [the `Badge` vertical bleed docs](https://braid-design-system--fc45a801e2598361a2a1b1d35627b9a7197defaa.surge.sh/components/Badge#vertical-bleed), i.e. do we want encourage putting the badge in the heading rather than bleeding?

Also, is the badge meant to be vertically centered at all heading sizes? It doesn't appear to be the case.
**EDIT**: The screenshot below is of the `apac` theme. It's perfectly aligned in `seekJobs`.
image

michaeltaranto:
Good shout on the docs example. I have updated with the following
Docs preview

In terms of the position of the badge, because it is being used  "inline" we are at the mercy of the line box created by the font itself as we are using `vertical-align: middle`. I have tweaked the spacing a touch to further improve across the current themes, plus added docs for this.

askoufis (approved):

PR #1550, opened 2024-08-01
Tabs: Add `size` support

michaeltaranto:
Introduces the ability to customise the `size` of the `Tab` components in the tab list.
Available sizes are `standard` (default) and `small`.

**EXAMPLE USAGE:**
```jsx
<Tabs size="small">
  <Tab>First tab</Tab>
  <Tab>Second tab</Tab>
  <Tab>Third tab</Tab>
</Tabs>
```


askoufis (approved):
🥇

PR #1572, opened 2024-09-13
Autosuggest: Fix bug in `suggestionHighlight` prop when using `remaining`

felixhabib:
Currently, the `suggestionHighlight` prop when set to `remaining` will set every match range to end with the length of the suggestion.

This does not work when matches are found in multiple words, where the match ranges may end at the end of a word rather than the end of the total string.

### DIFF
--- a/.changeset/forty-coins-wait.md
+++ b/.changeset/forty-coins-wait.md
@@ -7,7 +7,7 @@ updated:
   - Autosuggest
 ---
 
-**Autosuggest**: Improve handing of `suggestionHighlight` prop when set to `remaining`
+**Autosuggest**: Improve handling of `suggestionHighlight` prop when set to `remaining`
 
 Fixes a bug in `Autosuggest` when using `suggestionHighlight` prop set to `remaining`.
 If the input contained multiple words, the highlighted portion would be appended to the end of matching suggestions.
--- a/.changeset/forty-coins-wait.md
+++ b/.changeset/forty-coins-wait.md
@@ -1,0 +1,13 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated:
+  - Autosuggest
+---
+
+**Autosuggest**: Improve handing of `suggestionHighlight` prop when set to `remaining`
+
+Fixes a bug in `Autosuggest` when using `suggestionHighlight` prop set to `remaining`.
+If the input contained multiple words, the highlighted portion would be appended to the end of matching suggestions.
--- a/packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.docs.tsx
+++ b/packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.docs.tsx
@@ -14,7 +14,6 @@ import {
   Column,
   Columns,
   TextField,
-  Inline,
 } from '../';
 import { IconHelp, IconLanguage } from '../icons';
 import { highlightSuggestions } from './Autosuggest';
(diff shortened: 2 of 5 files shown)


packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.docs.tsx
```diff
@@ -474,11 +473,12 @@ const docs: ComponentDocs = {
           each suggestion.
         </Text>
       ),
-      Example: ({ id, setDefaultState, setState, getState }) =>
-        source(
+      Example: ({ id, setDefaultState, setState, getState }) => {
+        const suggestion = 'Apples and Bananas';
```
michaeltaranto, line 477: I would suggest reverting this to default state as the content of passed into `source` is used to generate the Playroom snippet, im guessing opening this example in Playroom errors with `suggestion` being undefined?

.changeset/forty-coins-wait.md
```diff
@@ -0,0 +1,13 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated:
+  - Autosuggest
+---
+
+**Autosuggest**: Improve handing of `suggestionHighlight` prop when set to `remaining`
```
michaeltaranto, line 10:
```suggestion
**Autosuggest**: Improve handling of `suggestionHighlight` prop when set to `remaining`
```

packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.docs.tsx
felixhabib: It would error, but we hide the code and option to open in Playroom since it doesn't actually use Autosuggest. But yeah happy to revert

packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.tsx
```diff
@@ -346,14 +347,12 @@ export function highlightSuggestions(
   value: string,
   variant: HighlightOptions = 'matching',
 ): SuggestionMatch {
-  const matches = matchHighlights(suggestion, value);
+  const matches =
+    variant === 'matching'
+      ? matchHighlights(suggestion, value)
+      : reverseMatches(suggestion, matchHighlights(suggestion, value));
```
michaeltaranto, line 353: I suggest extracting the matched highlights to a shared variable across both variants to guard against accidentally updating only one code path in the future

```suggestion
  const matchedHighlights = matchHighlights(suggestion, value);
  const matches =
    variant === 'matching'
      ? matchedHighlights
      : reverseMatches(suggestion, matchedHighlights);
```

packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.docs.tsx
michaeltaranto: Ah good shout! No probs

packages/braid-design-system/src/lib/components/Autosuggest/reverseMatches.ts
```diff
@@ -0,0 +1,22 @@
+export type Match = [number, number];
+
+export function reverseMatches(suggestion: string, matches: Match[]): Match[] {
+  const suggestionLength = suggestion.length;
```
michaeltaranto, line 4: Given the matches are zero based, feel like this should be `length - 1`?

E.g. in the tests above, reversing `[]` match on `Apple` produces `[0,5]`. Should that be `[0,4]`?

packages/braid-design-system/src/lib/components/Autosuggest/reverseMatches.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    const matches: Match[] = [[0, 5]];
+    const expected: Match[] = [];
+
+    expect(reverseMatches(suggestion, matches)).toEqual(expected);
+  });
+
+  it('should handle matches at the start of the suggestion', () => {
+    const suggestion = 'Apple';
+    const matches: Match[] = [[0, 2]];
+    const expected: Match[] = [[2, 5]];
```
michaeltaranto, line 38: Looking at these numbers, im not sure I agree. If we match `A P P` (e.g. `[0,2]`) and we reverse that, wouldnt we be expecting `L E` (e.g. `[3, 4]`)?

PR #1574, opened 2024-09-13
Stack, Tiles: Deprecate dividers prop

michaeltaranto:
In preparation for migrating Braid layout components to use [CSS gap], the `dividers` prop has been deprecated on `Stack` and `Tiles`.

Consumers are encouraged to migrate now in advance of its removal in v33.

[CSS gap]: https://developer.mozilla.org/en-US/docs/Web/CSS/gap


docs/Removing dividers support from layout components.md
```diff
@@ -0,0 +1,95 @@
+# Removing `dividers` support from layout components
+
+Following the update to SEEK&rsquo;s Browser Support Policy, we are now able to leverage [CSS gap] in Braid, and more importantly, our layout components. This will be landing in Braid in the v33 release.
+
+Previously, to enable gap-like behaviour layout components iterated over their children, wrapping them in container elements and applying padding. The trade off of this technique was an increased number of DOM elements and the introduction of unwanted space if the child element was hidden or return `null` — requiring developers to hoist logic to the parent component.
```
felixhabib, line 5: nit - not sure on the tense here?
```suggestion
Previously, to enable gap-like behaviour layout components iterated over their children, wrapping them in container elements and applying padding. The trade off of this technique was an increased number of DOM elements and the introduction of unwanted space if the child element was hidden or returned `null` — requiring developers to hoist logic to the parent component.
```

packages/braid-design-system/src/lib/components/Inline/Inline.docs.tsx
```diff
@@ -93,17 +93,19 @@ const docs: ComponentDocs = {
       ),
       Example: () =>
         source(
-          <Stack space="medium" dividers align="center">
+          <Stack space="medium" align="center">
```
felixhabib, line 96: The `align="center"` here makes the dividers disappear. Wrapping in Box width full fixes, or there might be a cleaner way. Maybe worth calling out in the migration guide too?

felixhabib: I think this will be working when we get gap Stack though?

packages/braid-design-system/src/lib/components/Inline/Inline.gallery.tsx
```diff
@@ -23,17 +23,19 @@ export const galleryItems: GalleryComponent = {
       label: 'Vertical alignment',
       Example: () =>
         source(
-          <Stack space="medium" dividers align="center">
+          <Stack space="medium" align="center">
```
felixhabib, line 26: Same as above comment

.changeset/five-ads-warn.md
```diff
@@ -0,0 +1,22 @@
...
+
+**Stack, Tiles:** Deprecate `dividers` prop
+
+In preparation for migrating Braid layout components to use [CSS gap], the `dividers` prop has been deprecated on `Stack` and `Tiles`.
+
+Consumers are encouraged to migrate now in advance of its removal in v33.
+
+#### Migration Guide
+
+See [migration guide] for details on how to migrate off the from `dividers` prop.
```
askoufis, line 19:
```suggestion
See [the migration guide] for details on how to migrate off the `dividers` prop.
```

docs/Removing dividers support from layout components.md
```diff
@@ -0,0 +1,95 @@
+# Removing `dividers` support from layout components
+
+Following the update to SEEK&rsquo;s Browser Support Policy, we are now able to leverage [CSS gap] in Braid, and more importantly, our layout components. This will be landing in Braid in the v33 release.
+
+Previously, to enable gap-like behaviour layout components iterated over their children, wrapping them in container elements and applying padding. The trade off of this technique was an increased number of DOM elements and the introduction of unwanted space if the child element was hidden or returned `null` — requiring developers to hoist logic to the parent component.
+
+Migrating to CSS gap mitigates all of the above trade offs.
+
+However, because `Stack` and `Tiles` no longer iterate over their children, the `dividers` feature is no longer feasible to implement centrally.
```
askoufis, line 9:
```suggestion
However, because `Stack` and `Tiles` no longer iterate over their children, implementing the `dividers` feature centrally is no longer feasible.
```

docs/Removing dividers support from layout components.md
```diff
@@ -0,0 +1,95 @@
...
+
+While we could have conditionally maintained this behaviour, it would have resulted in inconsistent edge cases when using `dividers` and not, e.g. if a child component rendered nothing or a hidden element, the `divider` would still be rendered, resulting in an inconsistent layout.
```
askoufis, line 11: Is `and not` meant to be here? Could also be split into a separate sentence.
```suggestion
While we could have conditionally maintained this behaviour, it would have resulted in inconsistent edge cases when using `dividers`. E.g. if a child component rendered nothing or a hidden element, the `divider` would still be rendered, resulting in an inconsistent layout.
```

packages/braid-design-system/src/lib/components/Inline/Inline.docs.tsx
michaeltaranto: As discussed we have improved `Divider` to prevent it collapsing when its a flex item and added it to this PR too.

docs/Removing dividers support from layout components.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Migration guide
+
+Braid already provides the [`Divider` component], so migrating away from the `dividers` prop is a matter of inserting a `Divider` as required into your layout.
+How you migrate will depend on whether the children of your layout component are static or being iterable.
```
askoufis, line 16:
```suggestion
How you migrate will depend on whether the children of your layout component are static or being iterated over.
```

felixhabib (approved):

askoufis (approved):

PR #1580, opened 2024-09-17
Text, Heading: Ensure `maxLines` truncates correctly in flex container

michaeltaranto:
Provide `min-width` to `Text` and `Heading` components to ensure `maxLines` truncates as expected when typography components are used as a direct child of a flex container.

### DIFF
--- a/.changeset/hungry-ads-cheer.md
+++ b/.changeset/hungry-ads-cheer.md
@@ -8,6 +8,4 @@ updated:
   - Heading
 ---
 
-**Text, Heading:** Ensure `maxLines` truncates correctly in flex container
-
-Provide `min-width` to `Text` and `Heading` components to ensure `maxLines` truncates as expected when typography components are used as a direct child of a flex container.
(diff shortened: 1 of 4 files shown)


.changeset/hungry-ads-cheer.md
```diff
@@ -0,0 +1,13 @@
...
+
+---
+updated:
+  - Text
+  - Heading
+---
+
+**Text, Heading:** Ensure `maxLines` truncates correctly in flex container
+
+Provide `min-width` to `Text` and `Heading` components to ensure `maxLines` truncates as expected when typography components are used as a direct child of a flex container.
```
askoufis, line 13: Feel like this is implementation detail. Could honestly confuse users into thinking they need to make a change.

michaeltaranto, line 13:
```suggestion
**Text, Heading:** Ensure `maxLines` truncates correctly when used as the direct child of a flex container.
```

askoufis (approved):

PR #1585, opened 2024-09-18
icons: Support rendering inline as child of a flex container

michaeltaranto:
When rendering our icons with [inline sizing], i.e. inside of `Text` or `Heading` components, they use negative margins and specific height and width values to improve alignment with the surrounding text. However, if when used as a direct child in a flex container, the values behave differently.

This PR wraps the icon in a `inline-block` element (when using [inline sizing]) to ensure predictable sizing and spacing when used in our layout components or as a flex item.

[inline sizing]: https://seek-oss.github.io/braid-design-system/foundations/iconography/#size-inline

### DIFF
--- a/.changeset/spicy-olives-travel.md
+++ b/.changeset/spicy-olives-travel.md
@@ -1,0 +1,105 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated:
+  - IconAI
+  - IconAdd
+  - IconArrow
+  - IconBookmark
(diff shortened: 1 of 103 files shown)


askoufis (approved):

askoufis:
Is it worth adding a screenshot test for this?

michaeltaranto:
Yeah good call. I'll add this in:

image

PR #1602, opened 2024-10-01
Stack, Tiles: Remove `divider` support

michaeltaranto:
As part of migrating our layout components to leverage flex gap, the `Stack` & `Tiles` no longer iterate over their children, making `dividers` no longer feasible to implement centrally.

While we could have conditionally maintained this behaviour, it would have resulted in inconsistent edge cases when using `dividers` and not, e.g. if a child component rendered nothing or a hidden element, the `divider` would still be rendered, resulting in an inconsistent layout.

### MIGRATION GUIDE: 
See changeset or [migration guide](https://github.com/seek-oss/braid-design-system/blob/0d694f136c4f0ab3c98196e81af93f7f15158855/docs/Removing%20dividers%20support%20from%20layout%20components.md)


.changeset/empty-numbers-live.md
```diff
@@ -0,0 +1,101 @@
...
+
+---
+updated:
+  - Stack
+  - Tiles
+---
+
+**Stack, Tiles:** Remove `divider` support
+
+As part of migrating our layout components to leverage flex gap, the `Stack` & `Tiles` no longer iterate over their children, making `dividers` no longer feasible to implement centrally.
```
askoufis, line 13:
```suggestion
As part of migrating our layout components to leverage flex gap, the `Stack` & `Tiles` components no longer iterate over their children, making `dividers` no longer feasible to implement centrally.
```

.changeset/empty-numbers-live.md
```diff
@@ -0,0 +1,101 @@
...
+
+While we could have conditionally maintained this behaviour, it would have resulted in inconsistent edge cases when using `dividers` and not, e.g. if a child component rendered nothing or a hidden element, the `divider` would still be rendered, resulting in an inconsistent layout.
```
askoufis, line 15: Feel like the `dividers` prop usage is implied here. Have also reworded it a bit. Does it still describe the problem accurately?

```suggestion
While we could have exposed an API to apply this behaviour conditionally, there are edge cases that cannot be handled correctly without consumer-side rendering logic, such as when child components render nothing or a hidden element.
```

packages/braid-design-system/src/lib/components/Tiles/Tiles.tsx
```diff
@@ -20,106 +15,46 @@ export interface TilesProps {
...
-            minWidth={0}
-            className={resolveResponsiveProp(
-              columns,
-              styles.columnsMobile,
-              styles.columnsTablet,
-              styles.columnsDesktop,
-              styles.columnsWide,
-            )}
+            height="full"
+            // This needs to be a separate element to support IE11.
```
askoufis, line 50: Can we collapse this element now that we don't support IE11?

michaeltaranto: Deliberately left this as divider removal only, as a subsequent PR refactors the `Tiles` component to use CSS grid.

askoufis (approved):

PR #1603, opened 2024-10-01
Spread: Narrow `component` options to valid layout elements

michaeltaranto:
Not all HTML elements make sense to be a layout container, e.g. `input`, so scoping the `component` prop to only surface relevant element types.


.changeset/three-squids-explode.md
```diff
@@ -0,0 +1,12 @@
+---
+'braid-design-system': minor
```
askoufis, line 2: This is technically a `major` right? It'll be combined with other `major`s on the `next` branch though, so not fussed.

askoufis (approved):

.changeset/three-squids-explode.md
michaeltaranto: Correct! I had an unsaved change locally for that 🤦 

```suggestion
'braid-design-system': major
```

PR #1604, opened 2024-10-02
Tiles: Migrate to CSS gap

michaeltaranto:
Migrate to CSS `gap` internally.

With the browser support policy now enabling adoption of CSS [gap], Braid’s layout components are now able to lean into the platform directly for its declarative, parent-driven approach to white space management.

Previously to enable gap-like behaviour, layout components iterated over their children wrapping them in container elements that applied padding.
The trade off of this technique was increased number of DOM elements and the introduction of unwanted space if the child element was hidden or return `null`, requiring developers to hoist logic to the parent component.

[gap]: https://developer.mozilla.org/en-US/docs/Web/CSS/gap


packages/braid-design-system/src/lib/components/Tiles/Tiles.tsx
```diff
@@ -24,37 +25,33 @@ export const Tiles = ({
...
+
+  return (
     <Box
-      display="flex"
-      flexWrap="wrap"
-      className={negativeMargin('left', space)}
+      gap={space}
+      className={styles.tiles}
+      style={assignInlineVars({
+        ...(mobileColumns
```
askoufis, line 41: Are these values potentially `undefined`? If so, they don't need to be excluded from the object, `assignInlineVars` will remove them from the inline style, assuming we have [v2.1.0](https://github.com/vanilla-extract-css/vanilla-extract/blob/284098c02dcfb85026dab1dbe29563301095530a/packages/dynamic/CHANGELOG.md#210) or higher installed.

askoufis: Ah, [still on 2.0.3](https://github.com/seek-oss/braid-design-system/blob/0d694f136c4f0ab3c98196e81af93f7f15158855/pnpm-lock.yaml#L188-L190). Worth updating the dep to `^2.1.0`.

askoufis (approved):

PR #1606, opened 2024-10-02
Stack: Migrate to CSS gap

michaeltaranto:
Migrate to CSS `gap` internally.

With the browser support policy now enabling adoption of CSS [gap], Braid’s layout components are now able to lean into the platform directly for its declarative, parent-driven approach to white space management.

Previously to enable gap-like behaviour, layout components iterated over their children wrapping them in container elements that applied padding.
The trade off of this technique was increased number of DOM elements and the introduction of unwanted space if the child element was hidden or return `null`, requiring developers to hoist logic to the parent component.

[gap]: https://developer.mozilla.org/en-US/docs/Web/CSS/gap


.changeset/tasty-weeks-search.md
```diff
@@ -0,0 +1,13 @@
...
+
+---
+updated:
+  - Stack
+---
+
+**Stack:** Widen `component` support
+
+With `Stack` no longer adding intermidiary elements, the `component` prop can now accept a wider range of elements.
+The list is kept to a white list of relevant elements to the component that do not require other HTML attributes, keeping in mind that props are not blindly spread in Braid.
```
askoufis, line 13:
```suggestion
Valid options are kept to a white list of elements relevant to `Stack` that do not require other HTML attributes, keeping in mind that props are not blindly spread in Braid.
```

.changeset/brave-eagles-teach.md
```diff
@@ -0,0 +1,24 @@
...
+---
+
+---
+updated:
+  - Stack
+---
+
+**Stack:** Consumers need to implement `li` elements
+
+When setting `component` to `ul` or `ol` on a `Stack`, consumers need to ensure they implement `li` elements as children.
```
askoufis, line 12: "Implement" feels a bit imprecise here.
```suggestion
**Stack:** Consumers need to render `li` elements

When setting `component` to `ul` or `ol` on a `Stack`, consumers need to ensure they render children as `li` elements.
```

askoufis (approved):

PR #1608, opened 2024-10-02
Hidden: deprecate `screen` prop and hide content by default

felixhabib:
Deprecate `screen` prop.
To provide content to screen readers without rendering it to the screen, consumers should use `HiddenVisually` instead.

### MIGRATION GUIDE:

```diff
- <Hidden screen>
+ <Hidden>
```

Also: `Hidden` will now hide content if no hidden conditions are specified, i.e. if no `above`, `below`, or `print` props are provided.
Consumers should review usage of the component without these props to ensure `Hidden` behaves as expected.


michaeltaranto:
An additional change we wanted to make here was that `<Hidden />` with no props would actually hide the content — essentially do what the `screen` prop does now (as you have shown in the migration guide).

So internally updating to something like.

```diff
 const hiddenOnScreen = Boolean(screen);
+const hiddenAlways =  hiddenOnScreen || !(above && below && print);

-<Box display={hiddenOnScreen ? 'none' : ...} />
+<Box display={hiddenAlways ? 'none' : ...} />
```

Lets chat more about this one though.

packages/braid-design-system/src/lib/components/Hidden/Hidden.tsx
```diff
@@ -32,18 +33,25 @@ export const Hidden = ({
   data,
   ...restProps
 }: HiddenProps) => {
-  if (process.env.NODE_ENV === 'development' && screen) {
-    // eslint-disable-next-line no-console
-    console.warn(
-      `You used the "screen" prop on Hidden, but this probably doesn't do what you expect. If you're trying to provide content to screen readers without rendering it to the screen, use the <HiddenVisually> component instead. The "screen" prop is likely to be deprecated in a future release.`,
-    );
+  if (process.env.NODE_ENV === 'development') {
+    if (screen) {
```
michaeltaranto, line 37: Recommend using a `typeof` check here to capture any usage of the prop.

```suggestion
    if (typeof screen !== 'undefined') {
```

packages/braid-design-system/src/lib/components/Hidden/Hidden.tsx
```diff
@@ -32,18 +33,25 @@ export const Hidden = ({
...
+      // eslint-disable-next-line no-console
+      console.warn(
+        `You used the "screen" prop on Hidden, but this probably doesn't do what you expect.
+        If you're trying to provide content to screen readers without rendering it to the screen, use the <HiddenVisually> component instead.
+        The "screen" prop is deprecated and will be removed in a future release.`,
```
michaeltaranto, line 42: The message could probably do with some better guidance than what we had before.

Thoughts on something like this:

> The `screen` prop is deprecated and will be remove in a future release.
> For content designed to improve the screen reader experience, please use <HiddenVisually> instead.
> Alternatively, if the content should be hidden unconditionally, remove the `screen` prop.

.changeset/tasty-weeks-search.md
michaeltaranto, line 1: Not sure this is a valid diff, suggest reverting the file on your branch.

packages/braid-design-system/src/lib/components/Hidden/Hidden.tsx
```diff
@@ -32,18 +39,26 @@ export const Hidden = ({
...
+    }
   }
 
   const inText = Boolean(useContext(TextContext));
   const inHeading = Boolean(useContext(HeadingContext));
 
   const hiddenOnScreen = Boolean(screen);
   const hiddenOnPrint = Boolean(print);
+
+  const hiddenAlways = hiddenOnScreen || (!above && !below && !print);
```
michaeltaranto, line 60: Think we need a changeset to capture the breaking change here — hiding when no props are specified.

.changeset/honest-years-exercise.md
```diff
@@ -0,0 +1,13 @@
+---
+'braid-design-system': minor
```
michaeltaranto, line 2:
```suggestion
'braid-design-system': major
```

michaeltaranto (approved):

PR #1611, opened 2024-10-03
Columns, Inline: Only respect `reverse` with `collapseBelow`

michaeltaranto:
The `reverse` prop is only respected in combination with `collapseBelow`.
This ensures the content is reversed on the same row, but follows the document order when collapsed.

If content needs to be reversed without `collapseBelow` then it should be reversed in the document/source order directly.

### DIFF
--- a/packages/braid-design-system/src/lib/components/Columns/Columns.tsx
+++ b/packages/braid-design-system/src/lib/components/Columns/Columns.tsx
@@ -44,7 +44,7 @@ export const Columns = ({
 
   assert(
     !reverse || (reverse && collapseBelow),
-    'The `reverse` prop should only be applied in combination with the `collapseBelow` prop.\nIf you do not want to collapse responsively, it is recommended to reorder the order of the content directly.\n\nSee documentation for details: https://seek-oss.github.io/braid-design-system/components/Columns#reversing-the-column-order',
+    'The `reverse` prop should only be applied in combination with the `collapseBelow` prop.\nIf you do not want to collapse responsively, it is recommended to reorder the content directly.\n\nSee documentation for details: https://seek-oss.github.io/braid-design-system/components/Columns#reversing-the-column-order',
   );
 
   const normalizedSpace = normalizeResponsiveValue(space);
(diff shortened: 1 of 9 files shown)


packages/braid-design-system/src/lib/components/Columns/Columns.tsx
```diff
@@ -42,6 +42,11 @@ export const Columns = ({
       .join(', ')}]`,
   );
 
+  assert(
+    !reverse || (reverse && collapseBelow),
+    'The `reverse` prop should only be applied in combination with the `collapseBelow` prop.\nIf you do not want to collapse responsively, it is recommended to reorder the order of the content directly.\n\nSee documentation for details: https://seek-oss.github.io/braid-design-system/components/Columns#reversing-the-column-order',
```
askoufis, line 47:
```suggestion
    'The `reverse` prop should only be applied in combination with the `collapseBelow` prop.\nIf you do not want to collapse responsively, it is recommended to reorder the content directly.\n\nSee documentation for details: https://seek-oss.github.io/braid-design-system/components/Columns#reversing-the-column-order',
```

packages/braid-design-system/src/lib/components/Inline/Inline.tsx
```diff
@@ -41,6 +41,11 @@ export const Inline = ({
       .join(', ')}]`,
   );
 
+  assert(
+    !reverse || (reverse && collapseBelow),
+    'The `reverse` prop should only be applied in combination with the `collapseBelow` prop.\nIf you do not want to collapse responsively, it is recommended to reorder the order of the content directly.\n\nSee documentation for details: https://seek-oss.github.io/braid-design-system/components/Inline#reversing-the-order',
```
askoufis, line 46:
```suggestion
    'The `reverse` prop should only be applied in combination with the `collapseBelow` prop.\nIf you do not want to collapse responsively, it is recommended to reorder the content directly.\n\nSee documentation for details: https://seek-oss.github.io/braid-design-system/components/Inline#reversing-the-order',
```

askoufis (approved):

PR #1619, opened 2024-10-09
Inline, Tiles: Enable truncated items

michaeltaranto:
Identified an issue where truncated items (more explicitly, elements with `white-space: nowrap`) within an `Inline` or `Tiles` would not truncate correctly, and break the boundaries of the parent container.

The fix is to ensure children of these components apply the `min-width: 0` so flex and grid infer their width and correctly truncate.

There is no changeset as this is a refactor on the CSS gap updates already in the `next` branch

### DIFF
--- a/packages/braid-design-system/src/lib/components/Inline/Inline.screenshots.tsx
+++ b/packages/braid-design-system/src/lib/components/Inline/Inline.screenshots.tsx
@@ -343,8 +343,7 @@ export const screenshots: ComponentScreenshot = {
       ),
     },
     {
-      label:
-        'Test - truncation should be visible on all examples below',
+      label: 'Test - truncation should be visible on all examples below',
       Container,
       Example: () => (
         <Inline space="small">
--- a/packages/braid-design-system/src/lib/components/Inline/Inline.screenshots.tsx
+++ b/packages/braid-design-system/src/lib/components/Inline/Inline.screenshots.tsx
@@ -344,7 +344,7 @@ export const screenshots: ComponentScreenshot = {
     },
     {
       label:
-        'Test - truncation should work an (all examples should truncate below)',
+        'Test - truncation should be visible on all examples below',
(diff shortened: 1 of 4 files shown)


packages/braid-design-system/src/lib/components/Inline/Inline.screenshots.tsx
```diff
@@ -342,5 +342,40 @@ export const screenshots: ComponentScreenshot = {
         </Inline>
       ),
     },
+    {
+      label:
+        'Test - truncation should work an (all examples should truncate below)',
```
felixhabib, line 347:
```suggestion
        'Test - truncation should work (all examples should truncate below)',
```
think this is the expected label?

michaeltaranto, line 347: Its a pretty bad label, lets try again: 
```suggestion
        'Test - truncation should be visible on all examples below',
```

felixhabib, line 347:
```suggestion
        'Test - truncation should work (all examples should truncate below)',
```
think this is the expected label?

packages/braid-design-system/src/lib/components/Tiles/Tiles.screenshots.tsx
```diff
@@ -29,5 +29,39 @@ export const screenshots: ComponentScreenshot = {
         </Tiles>
       ),
     },
+    {
+      label:
+        'Test - truncation should work and showing two equal tiles with truncated text',
```
michaeltaranto, line 34:
```suggestion
        'Test - truncation should be visible on both tiles below, as well as both tiles being equally sized',
```

felixhabib (approved):
tiny lint error, but otherwise nice! 🙌

PR #1620, opened 2024-10-10
Remove sku peer dependency

michaeltaranto:
Braid no longer has a peer dependency on [sku], however there are build-time requirements that remain — which are now documented.
SEEK consumers should continue to use [sku] for building Braid-based web experiences.

[sku]: https://seek-oss.github.io/sku/


README.md
```diff
@@ -99,6 +99,43 @@ Start a local Storybook server:
...
+This requires use of a bundler plugin to collect the extracted styles (see Vanilla Extract&rsquo;s [integration] documentation), either injecting them into the document or a separate CSS stylesheet.
+
+At SEEK this is done via [sku] as part of the build process.
+
+[Vanilla Extract]: https://vanilla-extract.style/
+[integration]: https://vanilla-extract.style/documentation/getting-started
+
+### Assertions
+
+To ensure correct usage of it&rsquo;s components, Braid performs some precondition and invariant checking at runtime using the [assert] library.
```
askoufis, line 117:
```suggestion
To ensure correct usage of its components, Braid performs some precondition and invariant checking at runtime using the [assert] library.
```

askoufis (approved):

PR #1633, opened 2024-10-20
Disclosure: Add `size` support

michaeltaranto:
Introduce the `size` prop to the `Disclosure` component, providing the same options as the `Text` component.

**EXAMPLE USAGE:**
```jsx
<Disclosure
  size="small"
>
  ...
</Disclosure>
```

### DIFF
--- a/packages/braid-design-system/src/lib/components/Disclosure/Disclosure.tsx
+++ b/packages/braid-design-system/src/lib/components/Disclosure/Disclosure.tsx
@@ -86,8 +86,8 @@ export const Disclosure = ({
   const size = sizeProp ?? textContext?.size ?? 'standard';
   const defaultSpace = isInline
     ? /*
-       * If inline, only use `xxsmall` as the space between the trigger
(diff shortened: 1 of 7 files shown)


packages/braid-design-system/src/lib/components/Disclosure/Disclosure.tsx
```diff
@@ -63,6 +83,15 @@ export const Disclosure = ({
         }),
   });
 
+  const size = sizeProp ?? textContext?.size ?? 'standard';
+  const defaultSpace = isInline
+    ? /*
+       * If inline, only use `xxsmall` as the space between the trigger
+       * and the content will include the line height of the text
+       */
```
felixhabib, line 91: Just curious about the reason for this

felixhabib (approved):

felixhabib:
While here, is there a reason we have the `Stack` in the `Disclosure` playroom snippet but not in any docs examples?

PR #1637, opened 2024-10-22
useToast: Support dynamic dismiss timeout duration

felixhabib:
Previously, all toasts would automatically dismiss after 10 seconds.
This update adjusts the timeout duration for toasts based on their content length, ensuring users have enough time to read the message, while minimising the time content is obscured by the toast.

Think actionable toasts should get a longer timeout

Timeout length is based off this calculation referenced [here](https://sheribyrnehaber.medium.com/designing-toast-messages-for-accessibility-fb610ac364be).

> A good length of time to keep messages up is 5 seconds plus 1 extra second for every 120 words, rounding up. This is how fast the average American reads. That means the shortest default that should be used as a best practice is 6 seconds.


packages/braid-design-system/src/lib/components/useToast/useTimeout.ts
```diff
@@ -15,6 +15,7 @@ export const useTimeout = ({ onTimeout, duration }: UseTimeoutProps) => {
 
   useEffect(() => {
     if (activated) {
+      setActivated(true);
```
felixhabib, line 18: One method of having the toast work on the docs site in local dev. Was hitting an issue with strict mode

.changeset/great-houses-grin.md
```diff
@@ -0,0 +1,13 @@
...
+
+---
+updated:
+  - useToast
+---
+
+**useToast**: Support dynamic dismiss timeout duration
+
+Previously, all toasts would automatically dismiss after 10 seconds.
+This update adjusts the timeout duration for toasts based on their content length, ensuring users have enough time to read the message, while minimising the time content is obscured by the toast.
```
michaeltaranto, line 13: Nit picking:

```suggestion
**useToast**: Reduce auto dismiss timeout and dynamically increase it based on content length.

Reduce auto dismiss timeout to 5 seconds (previously 10 seconds) as a baseline.
The timeout will increase based on the content length, ensuring users have enough time to read the message, while minimising the time content is obscured by the toast.
```

packages/braid-design-system/src/lib/components/useToast/useToast.docs.tsx
```diff
@@ -428,9 +428,9 @@ const docs: ComponentDocs = {
       description: (
         <>
           <Text>
-            {`A toast message can be dismissed via the close button or automatically after ${
-              toastDuration / 1000
-            } seconds.`}
+            A toast message can be dismissed via the close button or
+            automatically after a timeout period depending on its content
+            length.
```
michaeltaranto, line 433: I think we should level these docs up to capture the algorithm being used. It'd be good to interpolate the time values as well. We cna pair on the phrasing if you like.

packages/braid-design-system/src/lib/components/useToast/useTimeout.ts
michaeltaranto: Might need a catch up on this one 😕

packages/braid-design-system/src/lib/components/useToast/useToast.docs.tsx
felixhabib: Workshopped the language to detail the algorithm used, but happy to pair on this. I'm now thinking it's probably easier to describe the minimum as 6 seconds, as that's the minimum time in practice. Updated the variables to reflect this, thought the calculation will give the same result

PR #1644, opened 2024-11-01
site: Migrate preview sites from surge to ghpages

michaeltaranto:
Migrate our preview site deployments off [surge.sh] over to GitHub pages.

Additionally updating the main site publishing script to use GitHub actions in the same way in favour of the custom script.

[surge.sh]: https://surge.sh/


site/src/App/Navigation/Navigation.tsx
```diff
@@ -112,6 +120,30 @@ export const Navigation = () => {
...
+            >
+              <Text icon={<IconRocket />}>
+                Preview:{' '}
+                <TextLink
+                  href={`https://github.com/seek-oss/braid-design-system/compare/master...${config.branchName}`}
+                  target="_blank"
+                  weight="weak"
+                >
+                  <Strong>{config.branchName}</Strong>
+                </TextLink>
```
askoufis, line 143: Thoughts on something more descriptive than `Preview`, e.g `Preview branch:` or `Branch preview:`?

Also, comparing against `master` won't always be correct, e.g. a PR against a feature branch. Probably not worth addressing right now but worth thinking about.

michaeltaranto: Yeah was a lazy mode to get to a diff (and that will show you the View Pull Request button if there is one). Can definitely be improved if we pass more information through in the future!

Happy to update the prefix, can chat about that for sure.

site/src/App/Navigation/Navigation.tsx
```diff
@@ -112,6 +120,30 @@ export const Navigation = () => {
...
+              bottom={0}
+              right={0}
+              margin="medium"
+              boxShadow="small"
+              borderRadius="large"
+              background="cautionLight"
+              padding="medium"
+            >
+              <Text icon={<IconRocket />}>
+                Preview:{' '}
```
michaeltaranto, line 136: Update prefix text

```suggestion
                Branch preview:{' '}
```

askoufis (approved):

PR #1648, opened 2024-11-06
site: Improve definition of preview path, fix double slash on root

michaeltaranto:
Centralising the definition of the preview path for deployments to a single env var. This makes it easier to update the path once centrally (e.g. the custom grad-connection site we maintain).

Also fixing a double slash on the production site link when on the root page.

### DIFF
--- a/site/src/App/Navigation/Navigation.tsx
+++ b/site/src/App/Navigation/Navigation.tsx
@@ -105,9 +105,7 @@ const PreviewBranchPanel = () => {
             )}
           >
             <MenuItemLink
-              href={`https://seek-oss.github.io/braid-design-system${pathname}${
-                !pathname.endsWith('/') ? '/' : ''
-              }${hash}`}
+              href={`https://seek-oss.github.io/braid-design-system${pathname}${hash}`}
               target="_blank"
               icon={<IconNewWindow />}
             >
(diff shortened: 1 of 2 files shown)


site/src/App/Navigation/Navigation.tsx
```diff
@@ -105,7 +105,9 @@ const PreviewBranchPanel = () => {
             )}
           >
             <MenuItemLink
-              href={`https://seek-oss.github.io/braid-design-system${pathname}/${hash}`}
+              href={`https://seek-oss.github.io/braid-design-system${pathname}${
+                !pathname.endsWith('/') ? '/' : ''
```
askoufis, line 109: `/` isn't necessary before the has right? Can't this just be removed altogether?

michaeltaranto: Yeah thats what i thought! I initially had no slash, but the hash wasnt appending to the link. Let me test is again, maybe my dev server was in a weird state.

site/src/App/Navigation/Navigation.tsx
```diff
@@ -105,7 +105,9 @@ const PreviewBranchPanel = () => {
...
+              }${hash}`}
```
michaeltaranto, line 110: Yep, works as expected, will update.

```suggestion
              href={`https://seek-oss.github.io/braid-design-system${pathname}${hash}`}
```

askoufis (approved):

PR #1649, opened 2024-11-07
Preserve preview sites on master release

michaeltaranto:
Whitelisting the `preview` (and `grad-connection`) folders to prevent the release of the docs site from cleaning the output directory and blowing them away.

Also opting into `single-commit` as an optimisation as we have no need to git history on the `gh-pages` branch.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,3 +50,8 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           folder: ./site/dist
+          # Preserve the preview deployment folder and custom grad-connection site
+          clean-exclude: |
+            preview/*
+            grad-connection/*
+          single-commit: true


.github/workflows/release.yml
```diff
@@ -50,3 +50,8 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           folder: ./site/dist
+          # Preserve the preview deployment folder and custom grad-connection site
+          clean-exclude: |
+            preview/*
+            grad-connection/*
```
michaeltaranto, line 56: Planning to host the `grad-connection` branch on a custom folder path (like what we had on surge.sh previously).

```
https://seek-oss.github.io/braid-design-system/grad-connection/
```

This will provide a persistent URL for the team.

askoufis (approved):

PR #1650, opened 2024-11-08
RFC: advise not to use `tone` on text within `Alert` and `Notice`

tadhglewis:


packages/braid-design-system/src/lib/components/Alert/Alert.docs.tsx
```diff
@@ -70,6 +70,9 @@ const docs: ComponentDocs = {
             This component has only been designed to use standard size text. Any
             other size of text will break the alignment with the icon.
           </Text>
+          <Text>
+            The tone should be set on the Alert component and not on the text as this will lead to inconsistent blending.
+          </Text>
```
tadhglewis, line 75: Open to feedback as to the exact wording :word:

PR #1658, opened 2024-11-25
MenuRenderer: Ensure menu is visible in container with overflow hidden

felixhabib:
**MenuRenderer**: Ensure menu is visible, even when its trigger element is inside a container with overflow hidden.


packages/braid-design-system/src/lib/components/MenuRenderer/MenuRenderer.tsx
```diff
@@ -220,6 +270,18 @@ export const MenuRenderer = ({
...
+  useEffect(() => {
+    const handleResize = () => {
+      dispatch({ type: WINDOW_RESIZE });
+    };
+
+    if (open) {
+      window.addEventListener('resize', handleResize);
+    } else {
+      window.removeEventListener('resize', handleResize);
+    }
```
michaeltaranto, line 282: The recommended way to remove global handlers is in the clean up function of the effect. The clean up function is run when the component unmounts, or before runs again on state change.

So something like:
```suggestion
    if (open) {
      window.addEventListener('resize', handleResize);
    }
    
    return () => window.removeEventListener('resize', handleResize);
```

packages/braid-design-system/src/lib/components/MenuRenderer/MenuRenderer.tsx
```diff
@@ -120,85 +146,109 @@ export const MenuRenderer = ({
...
+          triggerPosition: buttonRef && getPosition(buttonRef.current),
+        };
+      }
+      default:
+        return state;
+    }
+  }, initialState);
+
+  useEffect(() => {
+    dispatch({ type: CLIENT_ENVIRONMENT });
```
michaeltaranto, line 250: I would suggest moving this state from the reducer to a local boolean ref.

felixhabib: Have removed this state as part of updating the Menu to only render on `open`

michaeltaranto (approved):
Love this!

PR #1661, opened 2024-12-03
MenuRenderer, OverflowMenu: Provide improved scroll affordance

michaeltaranto:
Introduce scroll affordance to menus providing a visual cue that there are more items overflowing vertically.


packages/braid-design-system/src/lib/components/MenuRenderer/MenuRenderer.tsx
```diff
@@ -380,30 +382,32 @@ export function Menu({
...
         ]}
       >
-        <Box paddingY={styles.menuYPadding} className={styles.menuHeightLimit}>
-          {Children.map(children, (item, i) => {
-            if (isDivider(item)) {
-              dividerCount++;
-              return item;
-            }
+        <ScrollContainer direction="vertical" fadeSize="small">
+          <Box className={styles.menuHeightLimit}>
```
felixhabib, line 386: The max height of the menu has increased slightly, and it looks like it's because the padding is applied on an element above where the height limit is set. Just worth noting but not an issue

felixhabib (approved):
Impressive masking stuff 🙌

packages/braid-design-system/src/lib/components/MenuRenderer/MenuRenderer.tsx
michaeltaranto: Actually, because the padding is outside now I can remove the padding from the max height calc in the stylesheet. Good pick up.

PR #1664, opened 2024-12-09
MenuRenderer, OverflowMenu: Improve scroll fade effect in older browsers

felixhabib:


packages/braid-design-system/src/lib/components/private/ScrollContainer/ScrollContainer.tsx
```diff
@@ -8,16 +8,20 @@ import { throttle } from 'throttle-debounce';
...
 import * as styles from './ScrollContainer.css';
 
+const scrollOffset = 2; // 2 instead of 1 to account for rounding errors in some browsers
+
 const maskOverflow = (
   element: HTMLElement,
   direction: keyof typeof styles.direction,
 ) => {
-  const atTop = element.scrollTop === 0;
+  const atTop = element.scrollTop <= 0;
```
felixhabib, line 17: This was adjusted to account for momentum based scrolling that allows for overflow, such as in Safari and in some mobile environments

PR #1672, opened 2024-12-18
Columns: Ensure component occupies available height

michaeltaranto:
Enables `Columns` content to occupy the available height of the parent container.


jahredhope (approved):
Alternative Title: Make children as tall as their parents, they grow up so fast

PR #1674, opened 2024-12-19
Accordion, AccordionItem: improve default space values

felixhabib:
Improve default spacing values and simplify internal layout.

This change reduces the default spacing within `Accordion` and `AccordionItem` components at certain sizes to improve its visual balance, ensuring the content is better associated with the correct `AccordionItem`.

Within the `Accordion` component, the default space between `AccordionItem` components has been reduced for size `large` with dividers, and sizes `small` and `xsmall` without dividers.
Within the `AccordionItem` component, the space between the `label` and content has been reduced for sizes `large` and `small`.

### DIFF
--- a/.changeset/wise-guests-bow.md
+++ b/.changeset/wise-guests-bow.md
@@ -8,9 +8,9 @@ updated:
   - AccordionItem
 ---
 
-**Accordion, AccordionItem**: Improve default spacing values and simplify internal layout.
+**Accordion, AccordionItem**: Adjust spacing values for improved visual balance.
 
-This change reduces the default spacing within `Accordion` and `AccordionItem` components at certain sizes to improve its visual balance, ensuring the content is better associated with the correct `AccordionItem`.
+This change reduces the default spacing within `Accordion` and `AccordionItem` components at certain sizes, ensuring the content is better associated with the correct `AccordionItem`.
 
 Within the `Accordion` component, the default space between `AccordionItem` components has been reduced for size `large` with dividers, and sizes `small` and `xsmall` without dividers.
 Within the `AccordionItem` component, the space between the `label` and content has been reduced for sizes `large` and `small`.
--- a/.changeset/wise-guests-bow.md
+++ b/.changeset/wise-guests-bow.md
@@ -1,0 +1,16 @@
+---
+'braid-design-system': patch
+---
+
+---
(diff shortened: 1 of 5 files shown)


.changeset/wise-guests-bow.md
```diff
@@ -0,0 +1,16 @@
...
+'braid-design-system': patch
+---
+
+---
+updated:
+  - Accordion
+  - AccordionItem
+---
+
+**Accordion, AccordionItem**: Improve default spacing values and simplify internal layout.
```
michaeltaranto, line 11: Lets break this into two changesets — improve default spacing, and simplify internal layout.

packages/braid-design-system/src/lib/components/Accordion/AccordionContext.ts
```diff
@@ -1,10 +1,11 @@
 import { createContext } from 'react';
 import type { TextProps } from '../Text/Text';
 
+export type validSizeValues = TextProps['size'];
```
michaeltaranto, line 4: Can we make types title case please? Also if we are not whitelisting a subset of sizes, maybe just use `TextProps['size']` without needing this separate type def. Thoughts?

packages/braid-design-system/src/lib/components/Accordion/AccordionItem.tsx
```diff
@@ -124,7 +124,12 @@ export const AccordionItem = ({
   });
 
   return (
-    <Box {...buildDataAttributes({ data, validateRestProps: restProps })}>
+    <Box
+      {...buildDataAttributes({ data, validateRestProps: restProps })}
+      display="flex"
+      flexDirection="column"
+      gap={itemSpace}
+    >
```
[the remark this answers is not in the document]
felixhabib: Yes correct, I missed this when migrating the spacing around

packages/braid-design-system/src/lib/components/Accordion/AccordionContext.ts
felixhabib: I just split it out to make the Record a bit cleaner in `Accordion.tsx`. But thinking about it, it might be ok to just use `TextProps['size']` there

michaeltaranto (approved):

PR #1704, opened 2025-02-03
site: Add `data` prop docs

michaeltaranto:
The `data` prop has previously been documented only on the `Box` component docs, as it is the basis for all components, but that is not discoverable by consumers using the other components (and not all `Box` APIs are exposed for every component).

Adding explicit `data` prop docs to each component to increase awareness and discoverability.

#### Review Note
Also updated the icon generation script to sort the dependencies correctly, and not rely on the format step.

### DIFF
--- a/packages/braid-design-system/src/lib/components/Accordion/Accordion.docs.tsx
+++ b/packages/braid-design-system/src/lib/components/Accordion/Accordion.docs.tsx
@@ -220,6 +220,36 @@ const docs: ComponentDocs = {
           </>,
         ),
     },
+    {
+      label: 'Data attributes',
+      description: (
+        <>
+          <Text>
+            Braid components are very explicit about the properties they accept,
+            which makes providing arbitrary{' '}
+            <TextLink href="https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes">
+              data attributes
+            </TextLink>{' '}
+            not possible. Instead, all Braid components accept a{' '}
+            <Strong>data</Strong> prop, allowing a single collection of data
+            attributes to be provided.
+          </Text>
+        </>
+      ),
+      code: `
+        <Accordion
+          data={{ testid: 'accordion-1' }}
+          // => data-testid="accordion-1"
+        >
+          <AccordionItem
+            data={{ testid: 'accordion-item-1' }}
+            // => data-testid="accordion-item-1"
+          >
+            ...
+          </AccordionItem>
+        </Accordion>
+      `,
+    },
   ],
 };
 
--- a/packages/braid-design-system/src/lib/components/Accordion/Accordion.docs.tsx
+++ b/packages/braid-design-system/src/lib/components/Accordion/Accordion.docs.tsx
@@ -12,6 +12,7 @@ import {
   IconImage,
 } from '../';
 import { Placeholder } from '../../playroom/components';
+import { dataAttributeDocs } from '../private/dataAttribute.docs';
 
 import { validSpaceValues } from './Accordion';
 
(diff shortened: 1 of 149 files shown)


packages/braid-design-system/src/lib/components/Accordion/Accordion.docs.tsx
```diff
@@ -220,6 +220,36 @@ const docs: ComponentDocs = {
...
+      label: 'Data attributes',
+      description: (
+        <>
+          <Text>
+            Braid components are very explicit about the properties they accept,
+            which makes providing arbitrary{' '}
+            <TextLink href="https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes">
+              data attributes
+            </TextLink>{' '}
+            not possible. Instead, all Braid components accept a{' '}
```
jahredhope, line 233: It feels weird to read about 'all' components here. For the docs on individual components would it be better to just talk about this prop?

```suggestion
            not possible. Instead this component accept a{' '}
```

This is helpful for future flexibility (what if in the future one component doesn't allow data for some reason), and also (I think) reads better as it's not jumping to universally scoped information

packages/braid-design-system/src/lib/components/Accordion/Accordion.docs.tsx
```diff
@@ -220,6 +220,36 @@ const docs: ComponentDocs = {
           </>,
         ),
     },
+    {
```
jahredhope, line 223: Have you considered adding this as a util function to DRY it up?

```suggestion
    dataAttributeDocs({code: `
        <Accordion
          data={{ testid: 'accordion-1' }}
          // => data-testid="accordion-1"
        >
          <AccordionItem
            data={{ testid: 'accordion-item-1' }}
            // => data-testid="accordion-item-1"
          >
            ...
          </AccordionItem>
        </Accordion>
      `})
```

Or just the description section?

packages/braid-design-system/src/lib/components/Accordion/Accordion.docs.tsx
michaeltaranto: Yeah nice, how about:
```
Instead, a `data` prop can be provided as a single collection of data attributes.
```

jahredhope: I like it. The less text the better as people will be seeing it on every page so the easier it is to see and say "oh yeah, I've read the before" the better.

packages/braid-design-system/src/lib/components/Accordion/Accordion.docs.tsx
michaeltaranto: Done!

packages/braid-design-system/src/lib/components/private/dataAttribute.docs.tsx
```diff
@@ -0,0 +1,43 @@
...
+  supportsNativeSyntax,
+  ...restProps
+}) => ({
+  label: 'Data attributes',
+  description: (
+    <>
+      <Text>
+        Braid components are very explicit about the properties they accept,
+        which makes providing arbitrary{' '}
+        <TextLink href="https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes">
```
jahredhope, line 25: Is this an old URL? This URL currently 301 redirects to: https://developer.mozilla.org/en-US/docs/Learn_web_development/Howto/Solve_HTML_problems/Use_data_attributes

```suggestion
        <TextLink href="https://developer.mozilla.org/en-US/docs/Learn_web_development/Howto/Solve_HTML_problems/Use_data_attributes">
```

Screenshot 2025-02-04 at 3 34 05 pm

jahredhope (approved):

packages/braid-design-system/src/lib/components/private/dataAttribute.docs.tsx
michaeltaranto: As old as can be. Nice catch!

jahredhope (approved):

PR #1705, opened 2025-02-03
site: Add Tabs `reserveHitArea` docs

michaeltaranto:
Adding some more missing docs


packages/braid-design-system/src/lib/components/Tabs/Tabs.docs.tsx
```diff
@@ -376,6 +376,51 @@ const docs: ComponentDocs = {
...
+      description: (
+        <>
+          <Text>
+            By default, a <Strong>Tab</Strong> will only occupy the vertical
+            space from the top of the label to the active underline. This means
+            the hit area will bleed out into the space above.
+          </Text>
+          <Text>
+            The bleed can be disabled by setting the{' '}
+            <Strong>reserveHitArea</Strong> prop to <Strong>true</Strong>.
```
askoufis, line 390: I was a bit confused by this. It reads as if the hit area bleed is being disabled, but really it's the space bleed that's being disabled, which contradicts the name of the prop.

michaeltaranto: Reworked the example to better compliment the description.

askoufis (approved):

PR #1711, opened 2025-02-11
icons: Update and extend suite

michaeltaranto:
See changesets


.changeset/curvy-crabs-grow.md
```diff
@@ -0,0 +1,26 @@
...
+
+Add new icons to represent other list types.
+Sits alongside `IconChecklist` which already exists.
+
+```tsx
+<IconBulletList />
+<IconNumberedList />
+```
+
+As a result we are deprecated `IconList` in favour of the more specific list types. The `IconList` will be removed in a future major version.
```
askoufis, line 19:
```suggestion
As a result we have deprecated `IconList` in favour of the more specific list types. `IconList` will be removed in a future major version.
```

askoufis (approved):

PR #1712, opened 2025-02-11
ci: Make clean up based on git date

michaeltaranto:
Add cron job to clean up workflow and switch from folder modified date to git commit date.

Currently setting to 12 days but we can adjust if we are hitting the limit again.


.github/workflows/clean-up-previews.yml
```diff
@@ -9,21 +11,32 @@ jobs:
...
+          done
+          if [[ $(git status --porcelain --untracked-files=no | wc -l) -eq 0 ]]; then
+            echo "No directories to clean up"
+            exit 0
+          fi
           git config --local user.name 'GitHub Actions'
           git config --local user.email 'actions@github.com'
           git add ${{ env.PREVIEW_DIR }}
           git commit --quiet -m "Clean up old preview directories"
+          git pull --rebase
```
michaeltaranto, line 39: Have to pull to work around race condition of preview site deployments happening in the background.

askoufis (approved):

PR #1719, opened 2025-02-25
site: Use ScrollContainer for side navigation

michaeltaranto:
As part of the colour re-work the scroll shadow got out of sync with the background on the docs site. Realised we can adopt our `ScrollContainer` (still private i know) component and leverage the masking technique rather than try and guess the background colours.

### DIFF
--- a/site/src/App/Navigation/Navigation.tsx
+++ b/site/src/App/Navigation/Navigation.tsx
@@ -169,10 +169,7 @@ export const Navigation = () => {
           bottom={0}
           transition="fast"
           width="full"
-          display={{
-            mobile: isMenuOpen ? 'block' : 'none',
-            wide: 'block',
-          }}
+          pointerEvents={!isMenuOpen ? 'none' : undefined}
           zIndex="sticky"
           className={[
             styles.sideNavigationContainer,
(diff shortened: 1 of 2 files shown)


site/src/App/Navigation/Navigation.tsx
```diff
@@ -163,11 +164,10 @@ export const Navigation = () => {
...
 
       <RemoveScroll enabled={isMenuOpen} forwardProps>
-        <FixedContentBlock
-          overflow="auto"
+        <Box
+          position="fixed"
           bottom={0}
-          paddingX={gutterSize}
-          paddingBottom="xxlarge"
+          transition="fast"
```
felixhabib, line 170: Just testing, it seems like the opacity transition works without adding the prop?

felixhabib (approved):
Looks great!

site/src/App/Navigation/Navigation.tsx
michaeltaranto: Just found the page content was transitioning out of view, but the menu was just popping in due to the change in the `display` prop. I have updated this to transition the menu in.

PR #1721, opened 2025-02-26
site: Fix side nav in mobile, disable pointerEvents when menu is closed

felixhabib:
Fixes side nav being clickable when closed on mobile

### DIFF
--- a/site/src/App/Navigation/Navigation.css.ts
+++ b/site/src/App/Navigation/Navigation.css.ts
@@ -19,7 +19,7 @@ const fixedWidthAboveMobile = style(
   }),
 );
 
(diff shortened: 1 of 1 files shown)


site/src/App/Navigation/Navigation.css.ts
```diff
@@ -19,7 +19,7 @@ const fixedWidthAboveMobile = style(
   }),
 );
 
-const hideOnMobileWhenOpen = style({
```
felixhabib, line 22: These names were slightly confusing when fixing the issue

michaeltaranto (approved):

PR #1725, opened 2025-03-02
Fix Clear button always being visible in Playroom TextFields

jahredhope:
Edit: switched solution to option 1

## Problem

Currently when not setting onClear on a TextField the Clear button (Cross on the right side) isn't visible in applications and the documentation site.

**Example: Documentation Site - No Clear Button**

example-docs-site-no-clear-button

However, it is visible in Playroom, due to Playroom's automatic injection of state handling.

**Example: Playroom - Clear Button**

example-playroom-has-clear-button

This creates a discrepancy between what users see in docs examples and playrooms, and makes it impossible create playrooms where the Clear button is not visible.

## Solution

Avoid passing onClear through to the TextField in Playroom when it's falsy.

To make it easy for prototyping to add the clear field behaviour allow passing `true` instead of a no-op function.

```jsx
<TextField
  label="Label"
  stateName="field1"
  onClear={true}
/>
```

**Example: Playroom - With Change now has no Clear button**

When onClear is not provided:

```jsx
<TextField
  label="Label"
  stateName="field1"
/>
```

There would now be no clear button:

example-playroom-no-close-with-change

[Playroom](https://seek-oss.github.io/braid-design-system/preview/c409d5e747e9ad42bc9c2f84538ccd924ea2f565/playroom/#?code=N4Igxg9gJgpiBcIA8BlALgQzAawAQGcAHLGAXgB0QAbDAJwHMZKA%2BcgO11yQBUYAPNADEAljCpR2nTjQBGYiiAAyGOVUqSpENgGEAFhjaNSwfDDToMaGAApKVgQDNR4ygEoAvhs4A3DFQCuZMCM5phWtiD2aE5iEiAeXrhoWmSUYLTCaMJgfuocUgC2MPj4GEaU2hlZOVS4RSVlTCAaAPSsbBo8-ELOEvnSKvKUyqp5UrhaegZGJmYW4XbdMeIATG6e-bi%2BAUEh8zaLjr1r8RvjyWypIIQQ%2BJnC3k2J9aXlIAAKt-ePdcWvT-k2uxOrwBCJYolZEMlIM1M1NpN9IYgqZQpYDpElr0AMzrRLbQLGPZhDFRZZQXGnRIXK45fxZLRjQp-RoKbQYenCLS-BqMJlAjr5Lpg3qQ2EKEZiJmcRHTFFzEkRMm9AAseM2BN2CvRSqxsTVVM2NIUl3ptFy8PGL1ZlAAcjAzX4ef9%2Be0kC0LDhmCAADSRXQweoIADaIAwxDAvpAphg2AAUhAZPgQABdP0Ad2EUDQumT8GD2IAnAAGH0AdgAbAAOH0ARhWVeLKfcQA)

## Alternative Options

The downside with the above solution is that playrooms now use invalid code `onClear={true}` in prototypes that can't be copied into application code.

**Alternative 1 - Opt-out of Clear label:**  We could default to showing the Clear button (as we do currently) but allow passing `onClear={false}` to deliberately opt-out of the clear label.

```jsx
<TextField
  label="Label"
  stateName="field1"
  onClear={false}
/>
```

This makes the minimal change, doesn't effect existing playrooms but doesn't fix the disconnect between examples and playrooms.

**Alternative 2 - Authentic behaviour:** We could simply not allow passing a boolean in, requiring users to enter a dummy function `onClear={() => {}}` to activate the clear button.

```jsx
<TextField
  label="Label"
  stateName="field1"
  onClear={() => {}}
/>
```

This may be confusing to playroom users who aren't familiar with JavaScript as well as breaking existing playrooms.

### DIFF
--- a/packages/braid-design-system/src/lib/components/TextField/TextField.docs.tsx
+++ b/packages/braid-design-system/src/lib/components/TextField/TextField.docs.tsx
@@ -13,6 +13,7 @@ import {
   Heading,
   IconLanguage,
   TextDropdown,
+  Notice,
 } from '../';
 import { dataAttributeDocs } from '../private/dataAttribute.docs';
 
@@ -223,6 +224,18 @@ const docs: ComponentDocs = {
             />
           </>,
         ),
+    },
+    {
+      description: (
+        <Notice tone="info">
+          <Text>
+            When prototyping in Playroom, fields will automatically include a{' '}
+            <TextLink href="/components/IconClear">clear icon</TextLink> and
+            clearing functionality. To disable this behaviour, set{' '}
+            <Strong>onClear</Strong> to <Strong>false</Strong>.
+          </Text>
+        </Notice>
+      ),
     },
     {
       label: 'Inserting an icon',
--- a/packages/braid-design-system/src/lib/components/TextField/TextField.playroom.tsx
+++ b/packages/braid-design-system/src/lib/components/TextField/TextField.playroom.tsx
@@ -11,9 +11,10 @@ import {
 } from './TextField';
 
 type PlayroomTextFieldProps = StateProp &
-  Optional<TextFieldBaseProps, 'id' | 'value' | 'onChange'> &
+  Optional<Omit<TextFieldBaseProps, 'onClear'>, 'id' | 'value' | 'onChange'> &
   TextFieldLabelProps & {
     onChange?: (fakeEvent: { currentTarget: { value: string } }) => void;
+    onClear?: true | TextFieldBaseProps['onClear'];
   };
 
 export const TextField = ({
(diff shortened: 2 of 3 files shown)


packages/braid-design-system/src/lib/components/TextField/TextField.docs.tsx
```diff
@@ -224,6 +225,18 @@ const docs: ComponentDocs = {
...
+      description: (
+        <Notice tone="info">
+          <Text>
+            When prototyping in Playroom, fields will automatically include a{' '}
+            <TextLink href="/components/IconClear">clear icon</TextLink> and
+            clearing functionality. To disable this behaviour, set{' '}
+            <Strong>onClear</Strong> to <Strong>false</Strong>.
+          </Text>
+        </Notice>
+      ),
```
felixhabib, line 238: Putting the docs messaging below the example to reduce prominence. Happy for suggestions otherwise

jahredhope: Love it

felixhabib (approved):

PR #1727, opened 2025-03-04
docs: Fix `TooltipRenderer` formatting example

felixhabib:
The current example shows text formatted with `weight="strong"`, which is the same as the default `Text` weight in `Tooltips` in the flagship themes.

Changing this to large `Text`, but could also use regular `Text` or another example


michaeltaranto (approved):
Happy for this to go in. But more interested in the weight change for tooltips. I think this may have bumped to `strong` since we co-erced `medium` up to `strong` in support of system font fallbacks for Thai.

Maybe tooltips should be `regular`? Worth looking at at some point.

PR #1738, opened 2025-03-18
Popover: Add private component

felixhabib:
NB: This change results in a slight change to the `OverflowMenu` menu positioning due to current implementation of `Popover` API. It is now positioned based on the `OverflowMenu` `ButtonIcon` (not its bounding box with negative margins), in line with standard use of `MenuRenderer` and `ButtonIcon`.

### DIFF
--- a/packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.tsx
+++ b/packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.tsx
@@ -779,9 +779,8 @@ export const Autosuggest = forwardRef(function <Value>(
             triggerRef={fieldRef}
             open={isOpen}
             width="full"
-            placement="bottom"
             lockPlacement
-            offsetSpace="none"
+            offsetSpace="xxsmall"
             role={false}
           >
             {/* MenuRef gets forwarded down to UL by RemoveScroll by `forwardProps`. */}
@@ -797,7 +796,6 @@ export const Autosuggest = forwardRef(function <Value>(
                 borderRadius="standard"
                 boxShadow="medium"
                 width="full"
-                marginTop="xxsmall"
                 paddingY="xxsmall"
                 className={styles.menu}
                 {...a11y.menuProps}
--- a/packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.tsx
+++ b/packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.tsx
@@ -781,10 +781,15 @@ export const Autosuggest = forwardRef(function <Value>(
             width="full"
             lockPlacement
             offsetSpace="xxsmall"
+            modal={false}
             role={false}
           >
             {/* MenuRef gets forwarded down to UL by RemoveScroll by `forwardProps`. */}
-            <RemoveScroll ref={menuRef} forwardProps>
+            <RemoveScroll
+              noRelative // Allows portalled elements to be positioned correctly relative to the viewport size
+              ref={menuRef}
+              forwardProps
+            >
               <Box
                 textAlign="left"
                 component="ul"
--- a/packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.tsx
+++ b/packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.tsx
@@ -782,6 +782,7 @@ export const Autosuggest = forwardRef(function <Value>(
             placement="bottom"
             lockPlacement
             offsetSpace="none"
+            role={false}
           >
             {/* MenuRef gets forwarded down to UL by RemoveScroll by `forwardProps`. */}
             <RemoveScroll ref={menuRef} forwardProps>
--- a/packages/braid-design-system/src/lib/components/ButtonIcon/ButtonIcon.test.tsx
+++ b/packages/braid-design-system/src/lib/components/ButtonIcon/ButtonIcon.test.tsx
@@ -17,8 +17,8 @@ describe('ButtonIcon', () => {
         <BraidTestProvider>
           {/*
             Rendering `PrivateButtonIcon` so no tooltip is added.
-            Popper tooltip library does not like being rendered
-            to static markup, but not required for this test.
+            Tooltip does not like being rendered to static markup,
+            but not required for this test.
           */}
           <PrivateButtonIcon icon={<IconBookmark />} label="Bookmark" />
         </BraidTestProvider>,
--- a/packages/braid-design-system/src/lib/components/ButtonIcon/ButtonIcon.test.tsx
+++ b/packages/braid-design-system/src/lib/components/ButtonIcon/ButtonIcon.test.tsx
@@ -3,27 +3,21 @@ import 'html-validate/jest';
 
 import { render } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import { renderToStaticMarkup } from 'react-dom/server';
 
 import { ButtonIcon, IconBookmark, Text } from '..';
 import { BraidTestProvider } from '../../../entries/test';
 
-import { PrivateButtonIcon } from './ButtonIcon';
-
 describe('ButtonIcon', () => {
   it('should render valid html structure', () => {
-    expect(
-      renderToStaticMarkup(
-        <BraidTestProvider>
-          {/*
-            Rendering `PrivateButtonIcon` so no tooltip is added.
-            Tooltip does not like being rendered to static markup,
-            but not required for this test.
-          */}
-          <PrivateButtonIcon icon={<IconBookmark />} label="Bookmark" />
-        </BraidTestProvider>,
-      ),
-    ).toHTMLValidate({
+    const { getByLabelText } = render(
+      <BraidTestProvider>
+        <ButtonIcon id="bookmark" icon={<IconBookmark />} label="Bookmark" />
+      </BraidTestProvider>,
+    );
+
+    const button = getByLabelText('Bookmark');
+
+    expect(button).toHTMLValidate({
       extends: ['html-validate:recommended'],
     });
   });
--- a/packages/braid-design-system/src/lib/components/MenuRenderer/MenuRenderer.screenshots.tsx
+++ b/packages/braid-design-system/src/lib/components/MenuRenderer/MenuRenderer.screenshots.tsx
@@ -1,5 +1,5 @@
 import { calc } from '@vanilla-extract/css-utils';
-import { useRef } from 'react';
+import { type MutableRefObject, type ReactNode, useRef } from 'react';
 import type { ComponentScreenshot } from 'site/types';
 
 import {
@@ -15,7 +15,7 @@ import {
 } from '../';
 import { vars } from '../../../entries/css';
 import { Placeholder } from '../private/Placeholder/Placeholder';
-import { Popover } from '../private/Popover/Popover';
+import { Popover, type PopoverProps } from '../private/Popover/Popover';
 import { debugTouchableAttrForDataProp } from '../private/touchable/debugTouchable';
 
 import { Menu } from './MenuRenderer';
@@ -35,6 +35,41 @@ const defaultProps = {
 } as const;
 
 const triggerHeight = 44;
+
+const PopoverWrapper = ({
+  children,
+  popoverPlacement,
+}: {
+  children: (props: {
+    triggerRef: MutableRefObject<HTMLButtonElement | null>;
+  }) => ReactNode;
+  popoverPlacement: PopoverProps['placement'];
+}) => {
+  const triggerRef = useRef<HTMLButtonElement | null>(null);
+  const wrapperPadding = calc(vars.touchableSize).multiply(2.5).toString();
+
+  return (
+    <Box
+      display="flex"
+      style={
+        popoverPlacement === 'bottom'
+          ? { paddingBottom: wrapperPadding }
+          : { paddingTop: wrapperPadding }
+      }
+    >
+      <Box ref={triggerRef}>
+        <Placeholder label="Menu trigger" height={triggerHeight} />
+      </Box>
+      {children({ triggerRef })}
+    </Box>
+  );
+};
+
+const ScreenshotPopover = ({
+  ...props
+}: Omit<PopoverProps, 'open' | 'lockPlacement' | 'role'>) => (
+  <Popover open lockPlacement role={false} {...props} />
+);
 
 export const screenshots: ComponentScreenshot = {
   screenshotWidths: [320],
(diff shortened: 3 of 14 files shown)


packages/braid-design-system/package.json
```diff
@@ -202,7 +202,6 @@
     "polished": "^4.1.0",
     "react-focus-lock": "^2.9.7",
     "react-is": "^18.2.0",
-    "react-popper-tooltip": "^4.3.1",
```
askoufis, line 205: 🫡

packages/braid-design-system/src/lib/components/MenuRenderer/MenuRenderer.tsx
```diff
@@ -342,44 +279,30 @@ export const MenuRenderer = ({
...
-            position="fixed"
-            zIndex="modal"
-            top={0}
-            left={0}
-            className={styles.backdrop}
-          />
-        </>
-      ) : null}
+    <Box {...buildDataAttributes({ data, validateRestProps: restProps })}>
+      <Box>{trigger(triggerProps, { open })}</Box>
```
michaeltaranto, line 283: Is this wrapping `Box` necessary, or a result of refactoring?

felixhabib: Looks like it can go

packages/braid-design-system/src/lib/components/private/Popover/Popover.tsx
```diff
@@ -0,0 +1,342 @@
...
+    if (!showPopover) {
+      return null;
+    }
+
+    return (
+      <BraidPortal>
+        <Box
+          data-testid={
+            process.env.NODE_ENV !== 'production' ? 'backdrop' : undefined
+          }
```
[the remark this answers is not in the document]
felixhabib: It's in Autosuggest, so was just copying from there. I'd rather not have a testid at all here, but was finding it tricky to have something else for testing library to find the backdrop

packages/braid-design-system/src/lib/components/private/Field/Field.tsx
```diff
@@ -251,6 +250,7 @@ export const Field = ({
                 }),
                 touchableText.standard,
                 icon && !prefix ? styles.iconSpace : null,
+                showSecondaryIcon ? styles.secondaryIconSpace : undefined,
```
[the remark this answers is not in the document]
felixhabib: [a4bc037](https://github.com/seek-oss/braid-design-system/pull/1738/commits/a4bc037bd308bf68b33cfdb65d8f556c43042cbb) - not too sure what detail to provide here

.changeset/heavy-webs-buy.md
```diff
@@ -0,0 +1,15 @@
...
+updated:
+  - Autosuggest
+  - Dropdown
+  - MonthPicker
+  - PasswordField
+  - TextField
+  - Textarea
+---
+
+Simplify internal layout.
```
michaeltaranto, line 15: Updating the list, as not all `Field` usages make use of the `secondaryIcon` slot.

Also having a go at describing the change. Let me know what you think.

```suggestion
  - Autosuggest
  - PasswordField
  - TextField
---

Ensure the `input` element fills the visual field boundary.

Previously when a field action such as "clear" or "toggle password visibility" was present, the native `input` element would shrink to accomodate.
This change ensures the `input` does not change size, instead applying padding to prevent the field value from colliding with the action.
The native browser styles for features such as [`:autofill`] now visually fill the field.
```

[`:autofill`]: https://developer.mozilla.org/en-US/docs/Web/CSS/:autofill

packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
```diff
@@ -138,207 +106,107 @@ export const TooltipRenderer = ({
...
-            </TooltipContent>
-          </div>
-        </BraidPortal>
-      )}
+      <Box
+        tabIndex={-1}
+        onClick={() => isMobile() && setOpen(!open)}
+        onMouseEnter={() => !isMobile() && setOpen(true)}
+        onFocus={() => !isMobile() && setOpen(true)}
+        onMouseLeave={() => !isMobile() && setOpen(false)}
```
michaeltaranto, line 180: Rather than calling `isMobile` in the handlers, it would be better to conditionally bind the handlers.

The `isMobile` check is a user agent check too (not screen size), so it can be set once as a ref too, rather than call multiple times on every render.

packages/braid-design-system/src/lib/components/private/Popover/Popover.test.tsx
```diff
@@ -0,0 +1,240 @@
...
+import { Popover } from './Popover';
+
+function renderPopover({
+  open = false,
+  useManagedState = false,
+  useEnterFocusRef = false,
+}: {
+  open?: boolean;
+  useManagedState?: boolean;
+  useEnterFocusRef?: boolean;
```
michaeltaranto, line 19: I know its internal, but suggest renaming to avoid confusing with the Hook naming convention

packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
```diff
@@ -51,85 +51,53 @@ export const TooltipTextDefaultsProvider = ({
...
+    </TooltipTextDefaultsProvider>
   </Box>
 );
 
-const validPlacements = ['top', 'bottom'] as const;
-
-type Placement = (typeof validPlacements)[number];
+interface TriggerProps {
+  // Using any to support any HTML element type. Specific types do not work with all elements.
+  ref: Ref<any>;
```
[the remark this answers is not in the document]
felixhabib: I'm seeing type issues with `HTMLButtonElement` in the `TooltipRenderer` docs. I think I can use `HTMLElement` in more places thought, but not sure how to fix this one

.changeset/heavy-webs-buy.md
felixhabib: Nice!!

packages/braid-design-system/src/lib/components/private/Popover/Popover.tsx
```diff
@@ -0,0 +1,342 @@
...
+  role?: string;
+  align?: 'left' | 'right' | 'center';
+  width?: 'content' | 'full';
+  placement?: Placement;
+  lockPlacement?: boolean;
+  offsetSpace?: ResponsiveSpace;
+  open: boolean;
+  onClose?: () => void;
+  triggerRef: RefObject<any>;
+  enterFocusRef?: RefObject<any>;
```
michaeltaranto, line 36: Should be able to use the base `HTMLElement` for these:

```suggestion
  triggerRef: RefObject<HTMLElement>;
  enterFocusRef?: RefObject<HTMLElement>;
```

packages/braid-design-system/src/lib/components/private/Popover/Popover.tsx
```diff
@@ -0,0 +1,342 @@
...
+        if (enterFocusRef.current) {
+          enterFocusRef.current.focus();
+        } else {
+          // eslint-disable-next-line no-console
+          console.error(
+            dedent`
+                The enterFocusRef element could not be found.
+                Ensure it is being assigned to a child element of Popover, and is available on open.
+                `,
+          );
```
michaeltaranto, line 169: This will log errors in production. I would suggest wrapping in dev warning block

```ts
enterFocusRef.current?.focus();

if (process.env.NODE_ENV !== 'production') {
  if (!enterFocusRef.current) {
    console.error(
      dedent`
        The enterFocusRef element could not be found.
        Ensure it is being assigned to a child element of Popover, and is available on open.
      `,
    );
  }
}
```

packages/braid-design-system/src/lib/components/private/Popover/Popover.tsx
```diff
@@ -0,0 +1,342 @@
...
+    useEffect(() => {
+      if (!open) {
+        setTriggerPosition(undefined);
+        setHorizontalOffset(0);
+        return;
+      }
+
+      setTriggerPosition(getPosition(triggerRef.current));
+      // Without timeout, focus will not work on first render
+      // Needs to be 10ms to work in Safari - 0 for other browsers
```
michaeltaranto, line 154: I'd suggest colocating this second comment with the timeout duration further down.

packages/braid-design-system/src/lib/components/private/Popover/Popover.tsx
```diff
@@ -0,0 +1,342 @@
...
+          zIndex="modal"
+          top={0}
+          left={0}
+          className={styles.backdrop}
+        />
+        <Box
+          id={id}
+          ref={ref}
+          role={role || 'dialog'}
+          component="section"
```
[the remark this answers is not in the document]
felixhabib: Ah interesting, will have a look. There's also the `popover` element for when we have a more modern browser policy

packages/braid-design-system/src/lib/components/private/Popover/Popover.tsx
```diff
@@ -0,0 +1,344 @@
...
+
+import * as styles from './Popover.css';
+
+const animationDelayInMs = 250;
+
+type Placement = 'top' | 'bottom';
+
+export interface PopoverProps {
+  id?: string;
+  role?: string;
```
michaeltaranto, line 27: We can type this:

```suggestion
  role?: AllHTMLAttributes<HTMLElement>['role'];
```

packages/braid-design-system/src/lib/components/private/Popover/Popover.tsx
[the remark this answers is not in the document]
michaeltaranto: Sorry I said `dialog` because i saw the role above (is that the correct default? or should we require the `role` be passed?). Interested if there is a semantic element we can use anyway, just need to be mindful that it is reset correctly until we can leverage its features.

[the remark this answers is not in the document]
felixhabib: I don't think we can access the `popover` element from `BoxProps` with the React version we're on now, so think it might be better to add a todo and leave this one for now.

Doing a bit of research and looking at other Popovers in the wild, it’s most common to use `div` and occasionally I’m seeing `section`. `section` seems reasonable based on the [mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/section), but I don’t think it’s adding much value.

With role, it feels like `dialog` is fitting and is commonly used, so I like this as the default. I can only really see `dialog`, `tooltip` or `menu` being used from what I’ve found. If we’re going with `dialog` we might need to open up `aria-labelledby` and `aria-describedby` as well. Thoughts?

packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.tsx
```diff
@@ -768,107 +770,110 @@ export const Autosuggest = forwardRef(function <Value>(
...
               </Box>
             )}
           </Field>
+          <Popover
+            triggerRef={fieldRef}
+            open={isOpen}
+            width="full"
+            placement="bottom"
+            lockPlacement
+            offsetSpace="none"
```
michaeltaranto, line 784: We can remove this as it is the default value

.changeset/better-animals-deny.md
michaeltaranto, line 1: Seems a merge issue has introduced some old changesets.

packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.tsx
```diff
@@ -768,107 +770,110 @@ export const Autosuggest = forwardRef(function <Value>(
...
+                component="ul"
+                background={
+                  !hasSuggestions && noSuggestionsMessage
+                    ? { lightMode: 'neutralSoft', darkMode: 'neutral' }
+                    : 'surface'
+                }
+                borderRadius="standard"
+                boxShadow="medium"
+                width="full"
+                marginTop="xxsmall"
```
michaeltaranto, line 800: Maybe this should be moved to `offsetSpace` above. Have a play as i think it would be like for like

packages/braid-design-system/src/lib/components/ButtonIcon/ButtonIcon.test.tsx
```diff
@@ -17,8 +17,8 @@ describe('ButtonIcon', () => {
         <BraidTestProvider>
           {/*
             Rendering `PrivateButtonIcon` so no tooltip is added.
-            Popper tooltip library does not like being rendered
-            to static markup, but not required for this test.
+            Tooltip does not like being rendered to static markup,
+            but not required for this test.
```
michaeltaranto, line 21: Given the new tests below are using `ButtonIcon`, lets try updating this as well. If it still doesnt work itd be worth updating the comment to reflect the constraint of using `html-validate/jest` for this test.

packages/braid-design-system/src/lib/components/MenuRenderer/MenuRenderer.screenshots.tsx
```diff
@@ -34,13 +36,41 @@ const defaultProps = {
...
+      {children({ triggerRef })}
+    </Box>
+  );
 };
 
+const ScreenshotPopover = ({
+  ...props
+}: Omit<PopoverProps, 'open' | 'lockPlacement' | 'role'>) => (
+  <Popover open lockPlacement role={false} {...props} />
+);
```
michaeltaranto, line 72: Lets move this from a component abstraction to a default props object that can be shared between examples below.

packages/braid-design-system/src/lib/components/private/Popover/Popover.screenshots.tsx
```diff
@@ -0,0 +1,371 @@
...
+      {children({ triggerRef })}
+    </Box>
+  );
+};
+
+const ScreenshotPopover = ({
+  ...props
+}: Omit<PopoverProps, 'open' | 'lockPlacement' | 'role'>) => (
+  <Popover open lockPlacement role={false} {...props} />
+);
```
michaeltaranto, line 51: Same a MenuRenderer screenshots, lets refactor this to a default props object and use Popover directly in the examples

packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
```diff
@@ -138,207 +105,115 @@ export const TooltipRenderer = ({
...
 
-  const isStatic = useContext(StaticTooltipContext);
-  const [controlledVisible, setControlledVisible] = useState(false);
-  const [opacity, setOpacity] = useState<0 | 100>(0);
   const { grid, space } = useSpace();
-  const triggerBoundingBoxRef =
-    useRef<ReturnType<typeof normaliseRect>>(defaultRect);
-  const tooltipBoundingRectRef =
-    useRef<ReturnType<typeof normaliseRect>>(defaultRect);
+  const edgeOffsetAsPx = space[edgeOffset] * grid;
```
michaeltaranto, line 116: Rename to `edgeOffsetInPx` for clarity

packages/braid-design-system/src/lib/components/private/Popover/Popover.tsx
```diff
@@ -0,0 +1,360 @@
...
+          }}
+          pointerEvents={!onClose ? 'none' : undefined}
+          position="fixed"
+          zIndex="modal"
+          top={0}
+          left={0}
+          className={styles.backdrop}
+        />
+
+        {/* Todo - use `popover` property when available */}
```
michaeltaranto, line 319: I would prefer this is moved to our Browser Support "desired features list" reminding us to return here, rather than documenting at the call site.

packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
```diff
@@ -138,207 +105,115 @@ export const TooltipRenderer = ({
...
-      };
-
-      const handleScroll = () => {
-        setControlledVisible(false);
-      };
+    const handleKeyDown = ({ key }: KeyboardEvent) => {
+      if (key === 'Escape') {
+        setOpen(false);
+      }
+    };
```
michaeltaranto, line 126: Can we refactor this to use `onClose` from `Popover`?

askoufis (commented):
Had this old review comment lying around.

PR #1744, opened 2025-03-28
Add `ref` to Table components

tadhglewis:
TODO

- [x] Test snapshot

note: [`forwardRef` is deprecated](https://react.dev/blog/2024/12/05/react-19#ref-as-a-prop). We could migrate now or wait for the codemod?

### DIFF
--- a/.changeset/rare-chairs-occur.md
+++ b/.changeset/rare-chairs-occur.md
@@ -2,4 +2,15 @@
 'braid-design-system': minor
 ---
 
-Add `ref` to table components
+---
+updated:
+  - Table
+  - TableBody
+  - TableCell
+  - TableFooter
+  - TableHeader
+  - TableHeaderCell
+  - TableRow
+---
+
+Add `ref` support to table components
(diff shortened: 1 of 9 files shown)


tadhglewis:
Ah this doesn't work as intended. For `@hello-pangea/dnd`, we need to pass in

`{...provided.draggableProps}`

```
interface DraggableProvidedDraggableProps {
    style?: DraggableStyle;
    'data-rfd-draggable-context-id': ContextId;
    'data-rfd-draggable-id': DraggableId;
    onTransitionEnd?: TransitionEventHandler;
}
```
Braid discards these props though as it only supports explicit `data` attribute 🤔

felixhabib:
Thanks Tadhg, if you're still interested in getting refs anyway, it probably makes sense to add support for all Table elements and not just TableRow and TableBody

tadhglewis:
@felixhabib added ref forwarding to all table components

michaeltaranto:
>note: [forwardRef is deprecated](https://react.dev/blog/2024/12/05/react-19#ref-as-a-prop). We could migrate now or wait for the codemod?

Having only recently dropped React 17, our baseline is to support React 18 so we can't use React 19 APIs internally just yet.

michaeltaranto (approved):

PR #1743, opened 2025-03-28
AccordionItem, Disclosure: Change `id` prop from required to optional

felixhabib:

PR #1748, opened 2025-04-02
Autosuggest, TextDropdown, TextField: Change `id` prop from required to optional

felixhabib:
Grouping these three components together because their docs files use each other. Easier to remove all ids from these files at once.

Also - forgot to remove `ids` from `Disclosure.gallery.tsx` in #1743. Adding it here

### DIFF
--- a/packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.tsx
+++ b/packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.tsx
@@ -14,6 +14,7 @@ import {
   useCallback,
   useEffect,
   forwardRef,
+  useId,
 } from 'react';
 import { RemoveScroll } from 'react-remove-scroll';
 
@@ -277,8 +278,9 @@ type HighlightOptions = 'matching' | 'remaining';
 
 export type AutosuggestBaseProps<Value> = Omit<
   FieldBaseProps,
-  'value' | 'autoComplete' | 'prefix'
+  'value' | 'autoComplete' | 'prefix' | 'id'
 > & {
+  id?: FieldBaseProps['id'];
   value: AutosuggestValue<Value>;
   suggestions:
     | Suggestions<Value>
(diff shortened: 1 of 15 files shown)


packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.tsx
```diff
@@ -377,6 +379,9 @@ export const Autosuggest = forwardRef(function <Value>(
     }
   }
 
+  const fallbackId = useId();
```
jahredhope, line 382: 👍 Nice work

jahredhope (commented):
I approve of this PR but it should probably wait for other's approval

packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.tsx
```diff
@@ -739,13 +744,13 @@ export const Autosuggest = forwardRef(function <Value>(
...
             componentName="Autosuggest"
-            id={id}
+            id={resolvedId}
             value={value.text}
             prefix={undefined}
             secondaryIcon={
               onClear ? (
                 <ClearField
-                  id={`${id}-clearfield`}
+                  id={`${resolvedId}-clearfield`}
```
michaeltaranto, line 753: We should note there is a refactor here when we update `ButtonIcon` to make `id` optional. There are a few layers we can update

packages/braid-design-system/src/lib/components/TextDropdown/TextDropdown.docs.tsx
```diff
@@ -108,7 +105,6 @@ const docs: ComponentDocs = {
       code: `
         <TextDropdown
           data={{ testid: 'text-dropdown-1' }}
-          // => data-testid="text-dropdown-1"
```
michaeltaranto, line 111: Was this deliberate?

packages/braid-design-system/src/lib/components/Autosuggest/Autosuggest.tsx
felixhabib: Yep yep, will add it to the card

packages/braid-design-system/src/lib/components/TextDropdown/TextDropdown.docs.tsx
felixhabib: Not deliberate, my bad

michaeltaranto (approved):

PR #1753, opened 2025-04-10
Support null in some more places

AaronMoat:
As a consumer, I found it surprising I couldn't do `<Checkbox badge={something ? <Badge ... /> : null}`; only undefined was accepted. Typically react node props accept `ReactNode` which has `null` built-in, but the strongly-typed arguments like this don't.

### DIFF
--- a/.changeset/nasty-rooms-stay.md
+++ b/.changeset/nasty-rooms-stay.md
@@ -2,9 +2,27 @@
 'braid-design-system': patch
 ---
 
-Support `null` for components' optional props, in addition to `undefined`
+---
+updated:
+  - AccordionItem
+  - Button
+  - ButtonLink
+  - Checkbox
+  - Heading
+  - MenuItem
+  - MenuItemLink
+  - RadioItem
+  - Tab
+  - Text
+  - TextLink
+  - TextLinkButton
+  - useToast
+---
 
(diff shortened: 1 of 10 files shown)


.changeset/nasty-rooms-stay.md
```diff
@@ -0,0 +1,13 @@
+---
+'braid-design-system': patch
+---
+
+Support `null` for components' optional props, in addition to `undefined`
+
+Previously, optional props could only be explicitly omitted with  `undefined`. This change allows you to alternatively pass `null`.
```
michaeltaranto, line 7:
```suggestion
---
updated:
  - AccordionItem
  - Button
  - ButtonLink
  - Checkbox
  - Heading
  - MenuItem
  - MenuItemLink
  - RadioItem
  - Tab
  - Text
  - TextLink
  - TextLinkButton
  - useToast
---

Support `null` for `badge` and `icon` slots, in addition to `undefined`.

Previously, `badge` and `icon` props could only be explicitly omitted with `undefined`.
This change allows passing `null`.
```

felixhabib: Ahh true, it is all `badge` and `icon` slots

felixhabib (approved):

PR #1755, opened 2025-04-21
Dialog, Drawer: Change `id` prop from required to optional

felixhabib:


packages/braid-design-system/src/lib/components/Dialog/Dialog.tsx
```diff
@@ -13,22 +15,42 @@ const modalStyle = {
...
-export const Dialog = ({ width = defaultWidth, ...restProps }: DialogProps) => (
-  <Modal width={width} {...restProps} {...modalStyle} />
-);
+export const Dialog = ({
+  id,
+  width = defaultWidth,
+  ...restProps
+}: DialogProps) => {
+  const fallbackId = useId();
+  const resolvedId = id || fallbackId;
```
michaeltaranto, line 29: Did you consider handling this once down in `ModalContent` and making that `id` optional?

That would make `id` just a passthrough for all these higher level `Dialog`, `DialogContent`, `Drawer` and `DrawerContent` components.

packages/braid-design-system/src/lib/components/private/Modal/ModalContent.tsx
```diff
@@ -99,12 +100,15 @@ export const ModalContent = ({
...
+  const fallbackId = useId();
+  const resolvedId = id || fallbackId;
+
   const defaultModalRef = useRef<HTMLElement>(null);
   const modalRef = modalRefProp || defaultModalRef;
   const defaultHeadingRef = useRef<HTMLElement>(null);
   const headingRef = headingRefProp || defaultHeadingRef;
 
-  const descriptionId = `${id}_desc`;
+  const descriptionId = `${resolvedId}_desc`;
```
michaeltaranto, line 111: As we did on the other PR, i'd suggest removing the custom suffixes for `descriptionId` and the close button, in favour of their own generated ids

michaeltaranto (approved):

PR #1756, opened 2025-04-21
Dropdown: Change `id` prop from required to optional

felixhabib:


packages/braid-design-system/src/lib/components/private/Field/Field.tsx
```diff
@@ -44,7 +44,7 @@ export type FieldLabelVariant =
     };
 
 export interface FieldBaseProps {
-  id: NonNullable<FormElementProps['id']>;
+  id?: NonNullable<FormElementProps['id']>;
```
michaeltaranto, line 47:
```suggestion
  id?: FormElementProps['id'];
```

packages/braid-design-system/src/lib/components/PasswordField/PasswordField.tsx
```diff
@@ -76,11 +76,13 @@ export const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
       [visible, onVisibilityToggle, inputRef],
     );
 
+    // Todo - remove once `ButtonIcon` id prop is optional
+    const id = useId();
```
michaeltaranto, line 80: I would suggest making this more specific, e.g. `visibilityToggleId`, and remove the templating below. The templating was only to make it unique off a single input value.
```suggestion
    const visibilityToggleId = useId();
```

packages/braid-design-system/src/lib/components/TextField/TextField.tsx
```diff
@@ -94,11 +89,13 @@ export const TextField = forwardRef<HTMLInputElement, TextFieldProps>(
         value.length > 0,
     );
 
+    // Todo - remove once `ButtonIcon` id prop is optional
+    const id = useId();
```
michaeltaranto, line 93: [As above](https://github.com/seek-oss/braid-design-system/pull/1756/files#r2053171947)

packages/braid-design-system/src/lib/components/private/Field/Field.tsx
```diff
@@ -149,8 +152,8 @@ export const Field = ({
     validateTabIndex(tabIndex);
   }
 
-  const messageId = `${id}-message`;
-  const descriptionId = description ? `${id}-description` : undefined;
+  const messageId = `${resolvedId}-message`;
+  const descriptionId = description ? `${resolvedId}-description` : undefined;
```
michaeltaranto, line 156: Similar to above, rather than template these off the single id, we can create unique ids for each.

packages/braid-design-system/src/lib/components/PasswordField/PasswordField.tsx
```diff
@@ -91,7 +93,7 @@ export const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
         secondaryIcon={
           disabled ? null : (
             <FieldButtonIcon
-              id={`${id}-toggle`}
+              id={`${visibilityToggleId}-toggle`}
```
michaeltaranto, line 96:
```suggestion
              id={visibilityToggleId}
```

packages/braid-design-system/src/lib/components/TextField/TextField.tsx
```diff
@@ -111,7 +108,7 @@ export const TextField = forwardRef<HTMLInputElement, TextFieldProps>(
         secondaryIcon={
           onClear ? (
             <ClearField
-              id={`${resolvedId}-clear`}
+              id={`${clearFieldButtonId}-clear`}
```
michaeltaranto, line 111:
```suggestion
              id={clearFieldButtonId}
```

packages/braid-design-system/src/lib/components/PasswordField/PasswordField.tsx
felixhabib: oh yikes 😅

PR #1758, opened 2025-04-22
OverflowMenu: Remove playroom wrapper component

felixhabib:
`id` for `OverflowMenu` is already optional


packages/braid-design-system/src/lib/components/OverflowMenu/OverflowMenu.playroom.tsx
michaeltaranto, line 1: If there is no reason for this override component anymore, i'd suggest removing the file and use the original component.

michaeltaranto (approved):

PR #1762, opened 2025-04-22
FieldLabel: Remove `useFallbackId` call for playroom

felixhabib:


packages/braid-design-system/src/lib/components/FieldLabel/FieldLabel.playroom.tsx
```diff
@@ -14,14 +13,7 @@ export const FieldLabel = ({
...
+  const fallbackFor = useId();
 
-  return (
-    <BraidFieldLabel
-      id={id ?? fallbackId}
-      htmlFor={htmlFor ?? fallbackFor}
-      {...restProps}
-    />
-  );
+  return <BraidFieldLabel htmlFor={htmlFor ?? fallbackFor} {...restProps} />;
```
michaeltaranto, line 18: Hmm, im not sure this makes sense. Why would we generate an id for the `for` attribute? 🤔 Doesnt really make sense.

I think this is only present to support reduced verbosity in the snippets by making `htmlFor` optional.

In that case this should probably be:
```tsx
type PlayroomFieldLabelProps = Optional<FieldLabelProps, 'htmlFor'>;

export const FieldLabel = ({
  htmlFor,
  ...restProps
}: PlayroomFieldLabelProps) => (
  <BraidFieldLabel htmlFor={htmlFor ?? false} {...restProps} />
);
```

felixhabib: Ahh yes, it was just to allow it to be optional, but this is much better!

michaeltaranto (approved):

PR #1763, opened 2025-04-22
MonthPicker, RadioGroup: Change `id` prop from required to optional

felixhabib:


packages/braid-design-system/src/lib/components/MonthPicker/MonthPicker.tsx
```diff
@@ -160,8 +165,8 @@ const MonthPicker = ({
 
   const monthRef = createRef<HTMLSelectElement>();
   const yearRef = createRef<HTMLSelectElement>();
-  const monthId = `${id}-month`;
-  const yearId = `${id}-year`;
+  const monthId = useId();
+  const yearId = useId();
```
michaeltaranto, line 169: Given these are not customisable and get passed straight to the `Dropdown` they probably arent necessary (as `Dropdown` will generate them).

But this does mean if you pass an `id` prop, it will be on the native input on mobile, or on the fieldset otherwise (not the dropdowns).

michaeltaranto (approved):

PR #1764, opened 2025-04-22
ButtonIcon: Change `id` prop from required to optional

felixhabib:


michaeltaranto (approved):
Great work 👏  Such a good clean up

PR #1766, opened 2025-04-23
Replace `useFallbackId` custom hook

felixhabib:
The custom hook for Playroom is no longer being used, so deleting this. Also removing `uuid` dependency that is no longer being used

Adding a new `useFallbackId` hook to consolidate logic for managing fallback ids in Braid components

### DIFF
--- a/packages/braid-design-system/package.json
+++ b/packages/braid-design-system/package.json
@@ -205,8 +205,7 @@
     "react-popper-tooltip": "^4.3.1",
     "react-remove-scroll": "^2.5.3",
     "throttle-debounce": "^5.0.2",
-    "utility-types": "^3.10.0",
-    "uuid": "^8.3.2"
+    "utility-types": "^3.10.0"
   },
   "devDependencies": {
     "@babel/core": "^7.26.0",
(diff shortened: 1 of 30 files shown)


michaeltaranto (approved):

packages/braid-design-system/package.json
```diff
@@ -205,8 +205,7 @@
     "react-popper-tooltip": "^4.3.1",
     "react-remove-scroll": "^2.5.3",
     "throttle-debounce": "^5.0.2",
-    "utility-types": "^3.10.0",
-    "uuid": "^8.3.2"
```
askoufis, line 209: Removing the `uuid` dep deserves a changeset IMO.

felixhabib: Can it be included in the changeset for the ids change since it's related? Or you reckon it gets its own?

askoufis: Oh, if that's all going to be in a single changeset then yeah worth just waiting until that release is ready 👍 .

askoufis (approved):

PR #1767, opened 2025-04-23
Remove remaining `id` usage

felixhabib:
* Remove id from `ExampleProps`
* Remove all use of `id` in Examples
* Remove logic to remove `id` in Playroom code

### DIFF
--- a/packages/braid-design-system/src/lib/components/Pagination/Pagination.docs.tsx
+++ b/packages/braid-design-system/src/lib/components/Pagination/Pagination.docs.tsx
@@ -43,7 +43,7 @@ const docs: ComponentDocs = {
         <TextLink href="https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/#x4-2-general-principles-of-landmark-design">
           General Principles of Landmark Design
         </TextLink>{' '}
-        it is neccessary for it to have an <Strong>aria-label</Strong>.
+        it is necessary for it to have an <Strong>aria-label</Strong>.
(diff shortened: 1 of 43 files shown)


packages/braid-design-system/src/lib/components/Pagination/Pagination.docs.tsx
```diff
@@ -43,7 +43,7 @@ const docs: ComponentDocs = {
         <TextLink href="https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/#x4-2-general-principles-of-landmark-design">
           General Principles of Landmark Design
         </TextLink>{' '}
-        it is neccessary for it to have an <Strong>aria-label</Strong>.
+        it is necessary for it to have an <Strong>aria-label</Strong>.
```
felixhabib, line 46: Not related but caught during cleanup

packages/braid-design-system/src/lib/components/useToast/useToast.docs.tsx
michaeltaranto, line 1: Think we might be able to avoid these ids being created here.

`Toast` requires an `id` internally but is using a [counter] that it increments. We could leverage `useId` there to create a unique `id` when it gets added to the toast queue.

Thoughts?

[counter]: https://github.com/seek-oss/braid-design-system/blob/master/packages/braid-design-system/src/lib/components/useToast/ToastContext.tsx#L172

askoufis (approved):

PR #1769, opened 2025-04-28
Remove all remaining `id` usage

felixhabib:
Some `id`s were missed in #1767. This change:

* Removes remaining `id` usage in `Example`
* Removes `id` in tests except where necessary
* Removes some missed `id` uses in snippets, Gallery and docs
* Removes `id` on `ButtonIcon` wrapper elements which are no longer necessary
* Updates `FieldMessage.playroom.tsx` to `useFallbackId`
* Change `id` prop from required to optional on `InlineField`

### DIFF
--- a/packages/braid-design-system/src/lib/components/Checkbox/Checkbox.tsx
+++ b/packages/braid-design-system/src/lib/components/Checkbox/Checkbox.tsx
@@ -1,6 +1,5 @@
 import { forwardRef } from 'react';
 
-import { useFallbackId } from '../../hooks/useFallbackId';
 import {
   type InlineFieldProps,
   InlineField,
@@ -17,9 +16,7 @@ export interface CheckboxProps
 }
 
 export const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
-  ({ id, checked, tabIndex, ...restProps }, ref) => {
-    const resolvedId = useFallbackId(id);
-
+  ({ checked, tabIndex, ...restProps }, ref) => {
     const calculatedChecked = Array.isArray(checked)
       ? resolveCheckedGroup(checked)
       : checked;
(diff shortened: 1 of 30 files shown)


packages/braid-design-system/src/lib/components/RadioGroup/RadioItem.tsx
```diff
@@ -47,10 +47,12 @@ export const RadioItem = forwardRef<HTMLInputElement, RadioItemProps>(
       radioItemContext === 0 && !Boolean(radioGroupContext.value);
     const tababble = checked || isFirstRadioWithNoCheckedValueInGroup;
 
+    const id = useId();
```
michaeltaranto, line 50: We can remove this in favour of `InlineField` doing this internally right?

packages/braid-design-system/src/lib/components/Checkbox/Checkbox.tsx
```diff
@@ -17,17 +16,14 @@ export interface CheckboxProps
 }
 
 export const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
-  ({ id, checked, tabIndex, ...restProps }, ref) => {
-    const resolvedId = useFallbackId(id);
-
+  ({ checked, tabIndex, ...restProps }, ref) => {
```
michaeltaranto, line 19: If `InlineFIeld` no longer requires an `id` the props can probably be cleaned up too now.

```diff
export interface CheckboxProps
+  extends Omit<InlineFieldProps, 'checked'> {
-  extends Omit<InlineFieldProps, 'id' | 'checked'> {
-  id?: InlineFieldProps['id'];
  checked: CheckboxChecked | boolean[];
}
```

michaeltaranto (approved):

PR #1770, opened 2025-04-28
Run `prettier` inside `source.macro` to trim down docs bundle

askoufis:
Since `id` has been phased out of code snippets, code to [remove `id` props from snippets](https://github.com/seek-oss/braid-design-system/pull/1767/files#diff-6628fe60035e75c5ffdce29c5d3147b718ae027f7e077135d60cbadb55680e62L52) was also recently removed. This makes it possible to do all code snippet formatting at build time (via the source macro) instead of at runtime.

Doing so lets us remove `prettier/standalone` from the docs site bundle, reducing its size from ~929kb to ~645kb. This is a reduction of **284kb** or **~31%**.

Another benefit of doing this is we can remove the async state updates that were made necessary by the prettier v3 update as we can utilize `@prettier/sync` for formatting since it is now performed in `node` rather than in the browser (which is incompatible with `@prettier/sync`).

The minor tradeoff is that we need to use `dedent` in a few more places, specifically wherever a plain string is used as a code snippet.

I've also marked this as a breaking change, just in case, as this _will_ affect the `code` string returned by the macro.

### DIFF
--- a/.changeset/better-animals-deny.md
+++ b/.changeset/better-animals-deny.md
@@ -8,13 +8,12 @@ Run `prettier` on macro code output
 
 The code string returned by the macro is now formatted with `prettier` by default. The resulting code string may be different relative to previous versions of the macro, so please ensure it aligns with your expectations.
 
-Prettier formatting can be disabled via the `formatWithPrettier` configuration option available on the macro:
+Prettier formatting can be disabled via the `formatWithPrettier` configuration option2️⃣:
 
 ```js
-import { source } from '@braid-design-system/source.macro';
-
-const code = source(
-  <Button>Click me</Button>,
-  { formatWithPrettier: false },
-);
+const babelTransformOptions = {
+  plugins: [
+    [require.resolve('babel-plugin-macros'), { source: { formatWithPrettier: false } }],
+  ],
+}
 ```
--- a/.changeset/better-animals-deny.md
+++ b/.changeset/better-animals-deny.md
@@ -1,5 +1,5 @@
 ---
-'@braid-design-system/source.macro': minor
+'@braid-design-system/source.macro': major
 ---
 
 Run `prettier` on macro code output
--- a/.changeset/better-animals-deny.md
+++ b/.changeset/better-animals-deny.md
@@ -1,9 +1,20 @@
 ---
-'@braid-design-system/source.macro': major
+'@braid-design-system/source.macro': minor
 ---
 
 Run `prettier` on macro code output
 
 **BREAKING CHANGE**:
 
-The code string returned by the macro is now formatted with `prettier`. The resulting code string may be different relative to previous versions of the macro, so please ensure it aligns with your expectations.
+The code string returned by the macro is now formatted with `prettier` by default. The resulting code string may be different relative to previous versions of the macro, so please ensure it aligns with your expectations.
+
+Prettier formatting can be disabled via the `formatWithPrettier` configuration option available on the macro:
+
+```js
+import { source } from '@braid-design-system/source.macro';
+
+const code = source(
+  <Button>Click me</Button>,
+  { formatWithPrettier: false },
+);
+```
--- a/.changeset/better-animals-deny.md
+++ b/.changeset/better-animals-deny.md
@@ -1,0 +1,9 @@
+---
+'@braid-design-system/source.macro': major
+---
+
+Run `prettier` on macro code output
+
+**BREAKING CHANGE**:
+
+The code string returned by the macro is now formatted with `prettier`. The resulting code string may be different relative to previous versions of the macro, so please ensure it aligns with your expectations.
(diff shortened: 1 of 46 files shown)


packages/source.macro/source.macro.test.ts
```diff
@@ -60,10 +60,52 @@ pluginTester({
         import source from './source.macro';
 
         const result = source({
-          hello: () => 'world',
-          some: 'value',
+          hello: () => ({ world: '!', foo: 'bar'}),
```
askoufis, line 63: For some reason the previous code here was unable to be parsed by `prettier` running inside the macro. I modified the test case a bit as a workaround.

askoufis:
Made a few more changes:
- When the `Code` component receives a `string` child, it now conditionally dedents that string. This means docs that consume `Code` don't need to dedent `string` snippets themselves.
- A lot of the example docs code snippets didn't use the `source` macro. This means they end up formatted using [`react-element-to-jsx-string`](https://www.npmjs.com/package/react-element-to-jsx-string). This does a good-enough job in most cases, but it _always_ puts children on a separate line, which vertically stretches out code snippets with lots of nesting. To mitigate this, I made 2 changes: use the `source` macro for all code snippets in example docs, and set `maxInlineAttributesLineLength: 60` so inline attributes can on the same line together.

packages/braid-design-system/src/lib/components/useToast/useToast.docs.tsx
michaeltaranto, line 1: Can we remove `dedent` from here too now?

askoufis: Woops, missed those. https://github.com/seek-oss/braid-design-system/pull/1770/commits/920d9a5d91464b53f1872d9391373022a5946169

packages/source.macro/source.macro.cjs
michaeltaranto, line 1: Should formatting be able to be controlled via config? I think for all our use cases we would want it formatted, but wonder whether we should be able to opt out of it?

What do you think?

askoufis: Good idea! Though this would be tedious since you can only deal with AST nodes inside the macro, but turns out `path.evaluate()` exists, which is very handy for this use case.

https://github.com/seek-oss/braid-design-system/pull/1770/commits/dde51ee949611bb63d2c3aee29e65ce080b9d951

.changeset/better-animals-deny.md
```diff
@@ -0,0 +1,20 @@
+---
+'@braid-design-system/source.macro': minor
```
askoufis, line 2: Downgraded this to `minor` as we're still 0-versioned.

packages/source.macro/__snapshots__/source.macro.test.ts.snap
```diff
@@ -103,13 +103,214 @@ exports[`source.macro 5. code only: 5. code only 1`] = `
...
+      
+
+      ↓ ↓ ↓ ↓ ↓ ↓
+
+const result = (
+  <Code>
+    {({ setDefaultState, getState: getStateUntyped }) => {
+      const getState = getStateUntyped as (key: string) => MockJob[];
+      return {
+        code: "<>\\n          {setDefaultState('jobs', [\\n  {\\n    id: 1,\\n    title: 'Lead Designer',\\n    location: 'Melbourne'\\n  },\\n  {\\n    id: 2,\\n    title: 'Senior Developer',\\n    location: 'Sydney'\\n  },\\n  {\\n    id: 3,\\n    title: 'Product Manager',\\n    location: 'Canberra'\\n  }]\\n  )}\\n\\n          <Stack space=\\"medium\\">\\n            {getState('jobs').map((job) =>\\n    <Card key={job.id}>\\n                <Stack space=\\"small\\">\\n                  <Text weight=\\"strong\\">{job.title}</Text>\\n                  <Text>{job.location}</Text>\\n                </Stack>\\n              </Card>\\n    )}\\n          </Stack>\\n        </>",
```
askoufis, line 258: Note the different quoting and whitespace relative to the formatted snapshot above.

packages/source.macro/source.macro.cjs
askoufis: Hoisted the config to the plugin level as discussed. https://github.com/seek-oss/braid-design-system/pull/1770/commits/8667eb7f88784328a1a3453e5add53b4a3d39045

packages/source.macro/__snapshots__/source.macro.test.ts.snap
```diff
@@ -103,13 +103,194 @@ exports[`source.macro 5. code only: 5. code only 1`] = `
...
+              <Card key={job.id}>
+                <Stack space="small">
+                  <Text weight="strong">{job.title}</Text>
+                  <Text>{job.location}</Text>
+                </Stack>
+              </Card>
+            ))}
+          </Stack>
+        </>,
+        { formatWithPrettier: false });
```
michaeltaranto, line 246: This is a hangover from the inline config right? Should we have a test that demos the formatting is disabled?

askoufis: Oh, yes it is.

askoufis: https://github.com/seek-oss/braid-design-system/pull/1770/commits/8b74ece2332034c5fc2f7c74e5f7ff59e806f99b

michaeltaranto (approved):

PR #1771, opened 2025-04-29
Alert, Badge, ScollContainer: Prevent vertical content clipping

felixhabib:
Issue can be seen on these components when using `Box` with display `flex` and the default `alignItems` value (stretch)

* [Pure example](https://seek-oss.github.io/braid-design-system/playroom/#?code=N4Igxg9gJgpiBcIA8AhCAPABFAlgZwAcAbAQwE8BeAHRADMiZ0aA%2BKgO001Q232POp0GTEJnqMAIjgBOMMABccENoMhEArgFs2Ldp05IAgg2nzM85TFXScisCSK6O%2BgwBVG81s5ecAomxxNTAA3GCJbTBwCPC1I6XVZTAgwezkSM3kYTQIIaSTaWhwwHBJMGHVsCCJczBIoPR9ORhhpYvl0pTYAOkxfMxIAoNkCWQALGDZYGzNadQBzErMYMwC8HFrwgEd1KMwYtiXBhsbt-EwwdSICUuGxiamIxh6AYWU8OUz5BMxbmHHJloRNjKMrhJalOrHHynDY4baldT9TTLSrVRLXGzpb44NjFXBQdQHTDAvDyeJQPaEpY4dR4TTQLpQgwAenc6E8UKQzOMLQ53i4aCw4xwc1G8kEsyIjlEzK8LMFXi5CvYIAANCB5ONkXgEABtEDvGAAawAUhAAEY6gC66oA7jgoJqdfBdQBGAAMACYACxWgC%2BQA)
* [More realistic example](https://seek-oss.github.io/braid-design-system/playroom/#?code=N4Igxg9gJgpiBcIA8AhCAPABFAlgZwAcAbAQwE8BeAHRADMiZ0aA%2BKgO00yQGUAXEsAGtMhATGohSAJwDmMFu06ckAQQZTemXhDbiaYKTl44wJIgo5LlAFUa9Wlq5wCibHAFtMANxhEjmHAI8AFdPHClgqRhMCDBTGFNNXhh3AggpGNpaExwSTBhg7AgidMwSKEUnTkYYKTAjEmMdADpMZ00SN08ogiiACxg2WENNWmCZXM0YTTc8HDK-AEdgwJFgtimuyqrl-EwwYKICPJ7%2BweH-RlaAYR08BOTeSMxTmAGh2v82HXy-KbzytsnLsFjhlnlgh13NMiiUophjoZGs8cGx6rgoOsZhA8LwIlA1ht8jhgnh3NBmkDlAB6WzoexApDUtS1BmOLi3IihNh4ETHMB6SQkWTyEAOKpITnc8VVLgABVIAr6xWGmAGOBkfV4FGAADYAAwAX0wpAARr4JFL3GwaJhqTKrEyrWwHUpJcVpVS3QqxMqiKr1ZrtXqjSaSOaiJaPdbbfavU7oy7GdTnXgZUy%2BAJBA4mWh0A4QAAaEC8AbQvAIADaIHuMEEACkIKaKwBdYsAdxwUFLFfglYAjPqAEwAFhbhqAA)

Screenshot tests would require a bit too many components as compared to the existing tests, so thought I'd just rely on a comment in the css. But happy to add these if necessary

### DIFF
--- a/packages/braid-design-system/src/lib/components/private/ScrollContainer/ScrollContainer.css.ts
+++ b/packages/braid-design-system/src/lib/components/private/ScrollContainer/ScrollContainer.css.ts
@@ -43,6 +43,8 @@ export const direction = styleVariants({
   horizontal: {
     overflowX: 'auto',
     overflowY: 'hidden',
+    // Ensures content doesn't clip when inside nested flex containers with stretched siblings
+    minHeight: 'fit-content',
   },
   vertical: {
     overflowX: 'hidden',
--- a/.changeset/wild-lights-hug.md
+++ b/.changeset/wild-lights-hug.md
@@ -1,0 +1,10 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated:
+  - Alert
+---
+
+**Alert**: Ensure `Alert` content is not vertically clipped when used inside nested flex containers with full height elements.
--- a/.changeset/wild-lights-hug.md
+++ b/.changeset/wild-lights-hug.md
@@ -5,6 +5,10 @@
 ---
 updated:
   - Alert
+  - Badge
+  - MenuRenderer
+  - OverflowMenu
+  - Table
 ---
 
-**Alert**: Ensure `Alert` content is not vertically clipped when used inside nested flex containers with full height elements.
+Ensure content is not clipped when used inside nested flex containers with stretched sibling elements
--- a/packages/braid-design-system/src/lib/components/Alert/Alert.css.ts
+++ b/packages/braid-design-system/src/lib/components/Alert/Alert.css.ts
@@ -2,6 +2,11 @@ import { style } from '@vanilla-extract/css';
 
 import { virtualTouchable } from '../private/touchable/virtualTouchable.css';
 
+export const alert = style({
+  // Ensures content doesn't clip when inside nested flex containers with stretched siblings
+  minHeight: 'fit-content',
+});
+
 export const closeButton = style([{}, virtualTouchable]);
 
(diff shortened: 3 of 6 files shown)


packages/braid-design-system/src/lib/components/private/ScrollContainer/ScrollContainer.css.ts
```diff
@@ -43,6 +43,8 @@ export const direction = styleVariants({
   horizontal: {
     overflowX: 'auto',
     overflowY: 'hidden',
+    // Ensures content doesn't clip when inside nested flex containers with stretched siblings
+    minHeight: 'fit-content',
```
michaeltaranto, line 47: Should this only be applied for the `horizontal` variant? What happens with `all`?

felixhabib: I'm avoiding `all` because it feels like a min width/height should only be applied to the direction that doesn't scroll. You would expect the content to exceed the available space and become scrollable, rather than the whole container grow to accommodate content size. Just for a testing example, if I set the `minHeight: 'fit-content'` on the component list on the docs site, the container grows past the screen edge and looses scroll.

I'm struggling to replicate the issue for the `vertical`, I would think it's best to add only if I can find a problem. With the other components, you would expect `Alert` and `Badge` to be able to shrink horizontally to accommodate for other content, but not vertically. So keeping the fix only vertically for these.

michaeltaranto (approved):

PR #1782, opened 2025-05-23
useToast: Add Toast stack

felixhabib:
Keen for feedback, and to get this into the next release if approved now that we have a cleaner slate


packages/braid-design-system/src/lib/components/useToast/Toast.tsx
```diff
@@ -139,60 +157,67 @@ const Toast = forwardRef<HTMLDivElement, ToastProps>(
...
+          textAlign="left"
+          background={{ lightMode: 'surfaceDark', darkMode: 'surface' }}
+          boxShadow={{
+            lightMode: 'borderNeutral',
+            darkMode: 'borderNeutralLight',
+          }}
+          borderRadius={borderRadius}
+          paddingY="medium"
+          paddingX="gutter"
+          marginTop={toastGap}
```
michaeltaranto, line 177: Can we use `gap` on the parent flex container to make this space parent controlled?

packages/braid-design-system/src/lib/components/useToast/Toaster.tsx
```diff
@@ -22,25 +26,55 @@ export const Toaster = ({ toasts, removeToast }: ToasterProps) => {
...
+        zIndex="notification"
+      >
+        <Box
+          width="full"
+          maxWidth={toastWidth}
+          pointerEvents={toasts.length === 0 ? 'none' : undefined}
+          marginX="gutter"
+          paddingBottom="xsmall"
+          onMouseEnter={() => !isMobile() && setExpanded(true)}
+          onMouseLeave={() => !isMobile() && setExpanded(false)}
```
michaeltaranto: I suggest storing the mobile check once in a variable then conditionally attaching the handlers. We are already calling `isMobile` on render above, so storing it and reusing it has no cost.
```tsx
onMouseEnter={!isMobile ? () => setExpanded(true) : undefined}
onMouseLeave={!isMobile ? () => setExpanded(false) : undefined}
```

(Maybe even store the functions, i.e. `expandToast = () => setExpanded(true)` for reuse between handlers here)

packages/braid-design-system/src/lib/components/useToast/useFlipList.ts
```diff
@@ -71,41 +97,95 @@ export const useFlipList = () => {
     }> = [];
 
     Array.from(refs.entries()).forEach(([toastKey, element]) => {
-      if (element) {
-        const prevTop = positions.get(toastKey);
-        const { top, height } = element.getBoundingClientRect();
-
-        if (typeof prevTop === 'number' && prevTop !== top) {
-          // Move animation
+      if (element && toastStates.get(toastKey) !== 'exiting') {
```
michaeltaranto: This might read better as a `switch` statement for the `toastStates`

packages/braid-design-system/src/lib/components/useToast/Toast.tsx
felixhabib: I think not unfortunately. I don't think you can animate between `gap` values which is why I had to use margin here, sadly. Can jam to revisit

packages/braid-design-system/src/lib/components/private/px.ts
michaeltaranto, line 1: This is probably be better suited to `src/lib/utils` than under `components/private`.

packages/braid-design-system/src/lib/components/useToast/Toaster.tsx
```diff
@@ -22,25 +27,55 @@ export const Toaster = ({ toasts, removeToast }: ToasterProps) => {
...
+          width="full"
+          maxWidth={toastWidth}
+          pointerEvents={toasts.length === 0 ? 'none' : undefined}
+          marginX="gutter"
+          paddingBottom="xsmall"
+          onMouseEnter={() => !isMobileDevice && setExpanded(true)}
+          onMouseLeave={() => !isMobileDevice && setExpanded(false)}
+          onClick={() => setExpanded(!expanded)}
+          onFocus={() => setExpanded(true)}
+          onBlur={() => setExpanded(false)}
```
michaeltaranto: Did you consider storing the functions for reuse, e.g.

```ts
const expandHandler = () => setExpanded(true);
const collapseHandler = () => setExpanded(false);
```

Also, the `isMobileDevice` check should conditionally bind the handler rather than always binding and checking inside the handler.
```tsx
  onMouseEnter={!isMobileDevice ? expandHandler : undefined}
  onMouseLeave={!isMobileDevice ? collapseHandler : undefined}
  onFocus={expandHandler}
  onBlur={collapseHandler}
  onClick={() => setExpanded(!expanded)}
```

packages/braid-design-system/src/lib/components/useToast/Toaster.tsx
```diff
@@ -22,25 +27,55 @@ export const Toaster = ({ toasts, removeToast }: ToasterProps) => {
...
+        bottom={0}
+        width="full"
+        zIndex="notification"
+      >
+        <Box
+          width="full"
+          maxWidth={toastWidth}
+          pointerEvents={toasts.length === 0 ? 'none' : undefined}
+          marginX="gutter"
+          paddingBottom="xsmall"
```
michaeltaranto: I believe it is these properties resulting in the screenshot diff (even though they dont result in a visual).

The `pointerEvents` change looks like it was working around this permanent space too. Happy to pair on this

felixhabib: Yes I thought so too, trying to only render when there are toasts

PR #1801, opened 2025-06-18
Seperate codemod

williamlark:
The braid-design-system/codemod package is bundled in with braid at the moment. This script contributes 14 MB of the 16.5 MB size of braid and is only used as a helper for migrating Braid version. Splitting this package out will reduce the download size of braid-design-system and allow us to iterate on the package independently. 

~I've marked this as a **major release** since it removes functionality.~


felixhabib:
Probably don't wanna trigger a major release for the braid package specifically if it can be avoided? The codemod package is fine

packages/codemod/package.json
```diff
@@ -10,6 +9,12 @@
...
   },
+  "bin": {
+    "braid-upgrade": "dist/index.js"
+  },
+  "files": [
+    "dist"
+  ],
   "author": "SEEK",
   "license": "MIT",
   "dependencies": {
```
askoufis: These should all be moved to `devDependencies`. This didn't matter before since the package was private, but now that it's not we don't want consumers to install all these deps since they're already bundled with the CLI.

askoufis:
> If it needs to be major, that's ok. But we should probably discuss when to release it. I'd prefer that current work isn't held behind a major release for our customers. Probably the sku group and others can give better feedback on if we can get away with minor

IMO this is one of those "technically major" but "practically minor" releases. The codemod is barely public API anyway, it's only "documented" inside the changelog, and when the time comes for consumers to actually run the codemod, we'll refer to the new package and it'll be fine.

Additionally, there haven't been any codemods since v31.11, so as long as a consumer doesn't skip straight to the latest version when coming from 30 or 30.10 or something, then they should be fine.

Ultimately I _think_ there's very little risk with this change.

packages/codemod/package.json
williamlark: Good catch. Fixed up

williamlark:
> > If it needs to be major, that's ok. But we should probably discuss when to release it. I'd prefer that current work isn't held behind a major release for our customers. Probably the sku group and others can give better feedback on if we can get away with minor
> 
> IMO this is one of those "technically major" but "practically minor" releases. The codemod is barely public API anyway, it's only "documented" inside the changelog, and when the time comes for consumers to actually run the codemod, we'll refer to the new package and it'll be fine.
> 
> Additionally, there haven't been any codemods since v31.11, so as long as a consumer doesn't skip straight to the latest version when coming from 30 or 30.10 or something, then they should be fine.
> 
> Ultimately I _think_ there's very little risk with this change.

Agree. Sounds like were happy to do a minor for this one. I've updated the changeset with a note mentioning the minor bump and changed the changeset release type for braid.

packages/codemod/README.md
```diff
@@ -0,0 +1,11 @@
+# @braid-design-system/codemod
+
+Automates migrations Braid migrations.
```
michaeltaranto:
```suggestion
Automates migrations between versions of Braid Design System.
```

packages/codemod/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Usage
+
+You must provide a target Braid version and a glob to your projects source files. For example:
+
+```bash
+pnpm dlx @braid-design-system/codemod v31.11 "**/*.{ts,tsx}
```
michaeltaranto:
```suggestion
pnpm dlx @braid-design-system/codemod v31.11 "**/*.{ts,tsx}"
```

packages/codemod/package.json
```diff
@@ -10,16 +9,25 @@
     "format:eslint": "eslint --cache --fix .",
     "format:prettier": "prettier --cache --write ."
   },
+  "bin": {
+    "braid-upgrade": "dist/index.js"
+  },
```
michaeltaranto, line 14: Now that we are removing the `bin` file from the other directory, how are we ensuring the built file has a `#!` env is set to `node`? i.e. `#!/usr/bin/env node`

williamlark: The index file includes it explicitly and it's included when built with esbuild. As long as we have it in the index.ts it will be added to the top of the output file. I tested it out to double check this was the case.

askoufis: IIRC pointing a module's `bin` to a built file causes issues when installing the package in a monorepo. We had this issue in `sku` when we moved to TypeScript. If you run `pnpm install` without running a `build` first, `pnpm` will try to find a non-existent `dist/index.js` file.

Having a separate `bin.js` that points to the built file fixes this. [Example in sku](https://github.com/seek-oss/sku/tree/d76bd1c2a6eb3e1e5ba9be3714f0fa00e4656c0a/packages/sku/bin). However, this _probably_ isn't an issue for the codemod since it's not installed anywhere in the monorepo.

.changeset/bumpy-ducks-push.md
```diff
@@ -0,0 +1,16 @@
+---
+'braid-design-system': minor
+'@braid-design-system/codemod': major
+---
+
+@braid-design-system/codemod has been moved to its own package.
+
+**Note: This is technically a breaking change for local migrations, but we expect minimal impact so are releasing this as a minor version change.**
+
+The `braid-upgrade` command is now no longer apart of the core Braid Design System. Instead, `braid-upgrade` can be run via the `@braid-design-system/codemod` package.
```
askoufis:
```suggestion
The `braid-upgrade` command is now no longer part of the `braid-design-system` package. Instead, `braid-upgrade` can be run via the `@braid-design-system/codemod` package.
```

PR #1808, opened 2025-06-25
Adding Vitest and bumping ink

williamlark:
This change bumps ink to v5, forcing codemod to become `type: module` since ink v4+ is now pure esm.

For easier testing of modules, vitest has also been added to the codebase and the codemod tests now rely on it. The aim is that vitest will replace Jest as the test running in this codebase, but the other migrations will happen in a seperate PR. 

To migrate a test to use vitest, simply change the test name. `*.test.*` > `*.vitest.*`.


packages/codemod/package.json
```diff
@@ -34,11 +35,10 @@
     "dedent": "^1.5.3",
     "esbuild": "^0.21.5",
     "fast-glob": "^3.3.2",
-    "ink": "^3.2.0",
+    "ink": "^5.2.0",
     "prettier": "^3.4.1",
     "react": "^18.3.1",
     "recast": "^0.21.1",
-    "workerpool": "^6.2.1",
-    "yoga-layout-prebuilt": "^1.10.0"
```
williamlark, line 42: `yoga-layout-prebuilt` was a dependancy in inks prior version. I assume we added it as a devdep here so that the patch is applied, since it's not used explicitly in our codebase. ink v4.1.0 moved to a WASM build of Yoga, so I've removed `yoga-layout-prebuilt` from devdeps as its no longer required.

vitest.config.ts
williamlark, line 1: We will need to add the react and vanilla-extract vite plugin here once we starting testing the braid ui code. I've left them out for now since they aren't needed yet.

jest.config.js
```diff
@@ -26,6 +26,7 @@ module.exports = {
   testPathIgnorePatterns: [
     `<rootDir>${slash}(.*${slash})?(dist|node_modules)${slash}`,
     `<rootDir>/integration`, // these have a separate Jest config
+    `\\.vitest\\.*$`,
```
michaeltaranto, line 29: Rather than scoping by filename and then ending up with a every file having `vitest` instead of `test`, what about scoping by package/directory? I.e. jest ignores the "codemod" package directory.

That way the file naming will end up `*.test.ts` at the end.

package.json
```diff
@@ -3,8 +3,10 @@
   "private": true,
   "version": "1.0.0",
   "scripts": {
-    "test": "pnpm run generate && jest",
+    "test": "pnpm run generate && vitest run && jest",
+    "test:jest": "pnpm run generate && jest",
     "test:integration": "jest --config integration/jest.config.js",
+    "test:vitest": "pnpm run generate && vitest run",
```
michaeltaranto, line 9: Im not sure it makes sense to have separate commands for the different frameworks. I understand having them while setting up the tooling, but I dont think there is value in committing these.

vitest.config.ts
```diff
@@ -0,0 +1,9 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    globals: true,
+    environment: 'jsdom',
+    include: ['**/*.vitest.(js|ts|tsx)'],
```
michaeltaranto, line 7: As above, I would prefer to stick with the `*.test.ts` convention, and scope to codemod package to start.

jest.config.js
williamlark: That was my initial approach, but scoping by file name has helped me test migrating files individually. They won't stay like this. Once we've moved off of jest I will change them all back to `.test.`, but its been useful separating them this way to support migrating things over.

package.json
williamlark: Yea these are also temporary while I'm transitioning things to vitest. I don't want to be running the full suite each time so these scripts are just so I can quickly test the two test runnings.

packages/codemod/package.json
askoufis: Should clean up the patch file too since you're deleting the dependency.

williamlark: Ah yes, good catch. Was meant to do this 👍

williamlark:
Draft PR for the rest of the vitest migration is here. A couple small lint errors I need to fix, but all the tests pass. All the temporary code mentioned in the comments has been removed in [this branch/PR](https://github.com/seek-oss/braid-design-system/pull/1809).

jest.config.js
askoufis: Honestly, given that #1809 isn't too big, maybe we merge it into this PR and do the full vitest migration in 1 go? Then we avoid the file rename and config shenanigans.

askoufis: Or even merge this branch into a temporary `vitest` branch, then fast follow with merging `#1809` into that branch, and then merge _that_ into `master`. I'm not a fan of feature branches, but if this one is short-lived then I think it's worth it.

michaeltaranto: Agree. Im happy with the one branch instead of temporary file renames/commands.

williamlark: The other PR is open. Take a look at it [here](https://github.com/seek-oss/braid-design-system/pull/1809) and I'll close this one.

PR #1809, opened 2025-06-25
Full vitest migration

williamlark:
Fully migrating to vitest as the test running within the Braid codebase. While this change doesn't add any speed improvements out-of-the-box (CI test speed has stayed the same), it does help in supporting esm within our code, and allows us to more easily transition packages to type module (like we have with braid codemod).

Also see #1808 (closed), This change bumps ink to v5, forcing codemod to become type: module since ink v4+ is now pure esm.

Notes:
* `html-validate` has a patch applied that fixes vitest ts types. They Have a PR raised to fix it already, so it should be fixed in the next release.

#### Comparison 

**vitest** (new)

CleanShot 2025-06-26 at 16 06 56

**jest** (old)

CleanShot 2025-06-26 at 16 08 26


packages/braid-design-system/src/lib/components/Tabs/Tabs.test.tsx
```diff
@@ -16,13 +14,6 @@ import {
 } from '..';
 import { BraidTestProvider } from '../../../entries/test';
 
-// We have the test environment set to `jsdom`, but the actual code runs in Node.js.
-// That means we will get the error that "useLayoutEffect does nothing on the server".
-// To avoid the error we need to manually switch to `useEffect`.
-jest.mock('../../hooks/useIsomorphicLayoutEffect', () => ({
-  useIsomorphicLayoutEffect: jest.requireActual('react').useEffect,
-}));
-
```
williamlark, line 25: This seems to work fine now within the tests so I've removed the code. Please let me know if there was a unique case that triggered this that I can recreate and I can keep it in.

.depcheckrc.yaml
```diff
@@ -1,11 +1,13 @@
 ignores:
   - '@crackle/cli'
   - '@testing-library/dom'
-  - '@types/jest'
```
DanDroryAu, line 4: Out with the old. In with the new!

package.json
askoufis, line 1: I think there's opportunity to clean up some unused deps from here. Specifically packages used in `jest/babelTransform.js` and the top-level `jest.config.js`, both of which have been deleted.

packages/braid-design-system/vitest.config.ts
```diff
@@ -0,0 +1,13 @@
+import { vanillaExtractPlugin } from '@vanilla-extract/vite-plugin';
+import react from '@vitejs/plugin-react';
+import { defineProject } from 'vitest/config';
+
+export default defineProject({
+  test: {
+    globals: true,
+    environment: 'jsdom',
+    setupFiles: './vitest-setup.ts',
+    css: true,
```
askoufis, line 10: Is `css: true` necessary for Vanilla Extract styles?

askoufis (commented):
Both shocked and relieved that this was a relatively uneventful migration. Well done!

packages/braid-design-system/vitest.config.ts
williamlark: Yep. Without it it's not able to tell when a component is hidden via css, like the "next" button when on the last page in the pagination component.

package.json
williamlark: Absolutely. I've gone ahead a deleted a bunch of them and building + testing still works. Let me know if you see anything I missed

michaeltaranto (approved):
Love this! Great work mate

PR #1810, opened 2025-06-26
Migrate focus rings to use `outline` and `focus-visible`, deprecate `outline` vars prop

felixhabib:


packages/braid-design-system/src/lib/components/Stepper/Stepper.css.ts
```diff
@@ -166,20 +171,15 @@ export const progressUnfilled = style({
 
 export const indicatorContainer = style({
   width: stepIndicatorSize,
+  outline: resetOutline,
+  transition: `${outlineTransition}, ${vars.transition.fast}`,
```
felixhabib: Needing to add both `outlineTransition` and `fast` transition here so one doesn't override the other. Maybe `fast` should get the `outlineTransition` as well? Worth investigating

packages/braid-design-system/src/lib/components/private/Overlay/Overlay.tsx
```diff
@@ -42,10 +40,7 @@ export const Overlay = ({
     transition={transition}
     inset={0}
     opacity={!visible ? 0 : undefined}
-    className={[
-      onlyVisibleForKeyboardNavigation ? hideFocusRingsClassName : null,
-      className,
-    ]}
+    className={[className]}
```
michaeltaranto:
```suggestion
    className={className}
```

packages/braid-design-system/src/lib/components/private/Modal/ModalContent.tsx
```diff
@@ -63,20 +62,12 @@ const ModalContentHeader = forwardRef<HTMLElement, ModalContentHeaderProps>(
       <Heading level={headingLevel} align={center ? 'center' : undefined}>
         <Box
           ref={ref}
-          component="span"
           tabIndex={-1}
-          outline="none"
+          component="span"
           position="relative"
-          className={styles.headingRoot}
+          className={[styles.headingRoot, styles.headingFocus]}
```
michaeltaranto: Given the class is only returning an atom, you could apply it directly on `Box` with the other style properties
```suggestion
          borderRadius="small"
          className={styles.headingRoot}
```

felixhabib: It's hard to see why there's a `borderRadius`, so I at least liked that we can name the class. I'll take your suggestion, but add a comment instead

packages/braid-design-system/src/lib/components/Tabs/TabPanel.tsx
```diff
@@ -52,18 +51,10 @@ export const TabPanel = ({
...
       position="relative"
-      outline="none"
       className={styles.tabPanel}
       {...buildDataAttributes({ data, validateRestProps: restProps })}
     >
       {isSelected || renderInactive ? children : undefined}
-      <Overlay
-        zIndex={1}
-        boxShadow="outlineFocus"
-        borderRadius="large"
```
michaeltaranto, line 63: This `borderRadius` needs to be transferred to the parent panel element above. TabPanels are looking pointy 😄

packages/braid-design-system/src/lib/components/Tabs/Tabs.css.ts
```diff
@@ -50,12 +50,7 @@ export const marginAuto = style({
 });
 
 export const tabFocusRing = style({
-  margin: vars.borderWidth.large,
-  selectors: {
-    [`${tab}:focus &`]: {
-      opacity: 1,
-    },
-  },
+  outlineOffset: calc.negate(vars.borderWidth.large),
```
michaeltaranto: This should use the `focusRingSize` to prevent clipping.
```suggestion
  outlineOffset: calc.negate(vars.focusRingSize),
```

packages/braid-design-system/src/lib/components/TextDropdown/TextDropdown.tsx
```diff
@@ -96,8 +95,7 @@ export function TextDropdown<Value>({
         position="absolute"
         inset={0}
         width="full"
-        opacity={0}
-        className={styles.select}
+        className={[styles.select, styles.focusOverlay]}
```
michaeltaranto: Lets inline the borderRadius with a comment like the other update.

packages/braid-design-system/src/lib/components/Toggle/Toggle.css.ts
```diff
@@ -116,6 +121,8 @@ export const slider = styleVariants(sizes, (size) => {
   return {
     height: vars.inlineFieldSize[size],
     width: vars.inlineFieldSize[size],
+    outline: resetOutline,
+    transition: `${vars.transition.fast}, ${outlineTransition}`,
```
michaeltaranto:
```suggestion
    transition: [
      vars.transition.fast,
      outlineTransition
    ].join(', '),
```

packages/braid-design-system/src/lib/components/Stepper/Stepper.css.ts
michaeltaranto: Lets use the array joining here too

packages/braid-design-system/src/lib/css/reset/reset.css.ts
```diff
@@ -11,10 +15,10 @@ export const base = style({
...
   verticalAlign: 'baseline',
   WebkitTapHighlightColor: 'transparent',
-  selectors: {
-    [`[${hideFocusRingsDataAttribute}] &`]: {
-      outline: 'none',
-    },
+  outline: resetOutline,
+  transition: outlineTransition,
+  ':focus-visible': {
+    outlineColor: focusOutlineColor,
```
michaeltaranto: Can we transition `outline` rather than just `outlineColor`. Only reason to do that would be whether we can create a `var` that represents the full `outline` to both provide to consumers and use ourselves, or whether the theme should have an object that represents the outlines parts.

packages/braid-design-system/src/lib/css/atoms/atomicProperties.ts
```diff
@@ -38,6 +38,13 @@ const boxShadow = {
   borderPositiveLight: `inset 0 0 0 ${vars.borderWidth.standard} ${vars.borderColor.positiveLight}`,
   borderPromote: `inset 0 0 0 ${vars.borderWidth.standard} ${vars.borderColor.promote}`,
   borderPromoteLight: `inset 0 0 0 ${vars.borderWidth.standard} ${vars.borderColor.promoteLight}`,
+  // Note, this does not actually deprecate the value.
```
michaeltaranto: Check the value is deprecated when moving through the list of boxshadows in the IDE, if so you can remove this comment.

felixhabib: It does not show 😔

packages/docs-ui/src/components/LinkableHeading/LinkableHeading.css.ts
```diff
@@ -1,9 +1,16 @@
...
+import { vars } from 'braid-design-system/css';
 
-export const linkableHeading = style({});
+export const linkableHeading = style({
+  display: 'inline-block',
+  maxWidth: 'fit-content',
+  scrollMarginBlockStart: vars.space.small,
+  borderRadius: vars.borderRadius.small,
+  outlineOffset: vars.space.xsmall,
+});
```
michaeltaranto: Given the root was a `flex` container previously, the `display` property should probably move to `block`. With the combination of `maxWidth: 'fit-content'` this should give us the desired result.

Also composing with `atoms` for common styles. 

```suggestion
export const linkableHeading = style([
  atoms({
    display: 'block',,
    borderRadius: 'small',
  }),
  {
    maxWidth: 'fit-content',
    scrollMarginBlockStart: vars.space.small,
    outlineOffset: vars.space.xsmall,
  }
]);
```

packages/braid-design-system/src/lib/css/atoms/atoms.ts
```diff
@@ -16,6 +18,39 @@ export const atoms = ({ reset, ...rest }: Atoms) => {
...
+      console.warn(
+        dedent`
+          \`boxShadow='outlineFocus'\` is deprecated and will be removed in a future release.
+    
+          Braid manages most focus outlines automatically.
+          For a custom focus outlines, consider \`outline='focus'\` or \`outlineStyle\`.
+
+          See documentation:
+          1. https://seek-oss.github.io/braid-design-system/css/vars
+          2. https://seek-oss.github.io/braid-design-system/css/outlineStyle
```
michaeltaranto:
```suggestion
          Braid no longer uses the `box-shadow` for focus outlines.
          Instead, consider \`outline='focus'\` or the \`outlineStyle\` css utility.
```

packages/braid-design-system/src/lib/css/atoms/atoms.ts
```diff
@@ -16,6 +18,39 @@ export const atoms = ({ reset, ...rest }: Atoms) => {
...
+      console.warn(
+        dedent`
+          \`outline='none'\` is deprecated and will be removed in a future release.
+    
+          Braid manages most focus outlines automatically.
+          For a custom focus outlines, consider \`outline='focus'\` or \`outlineStyle\`.
+
+          See documentation:
+          1. https://seek-oss.github.io/braid-design-system/css/vars
+          2. https://seek-oss.github.io/braid-design-system/css/outlineStyle
```
michaeltaranto:
```suggestion
          Braid manages the focus outlines of interactive elements as part of its
          scoped CSS reset.
          For a custom focus outlines on non-interactive elements, consider
          \`outline='focus'\` or \`outlineStyle\` for more complex scenarios.

          See documentation:
          1. https://seek-oss.github.io/braid-design-system/components/Box
          2. https://seek-oss.github.io/braid-design-system/css/outlineStyle
```

PR #1822, opened 2025-07-01
Add support for React 19.x

williamlark:
The Braid API has been tested to verify React 19 support, and as such the peer dependancies have been updated. No React 19 APIs are being used yet so React 18 is still supported and no breaking changes are expected when updating, hence the release as a patch.

Notable changes:
* `react-router-dom` > `react-router`
* `isFragment.ts` was pulled out from `react-is` to support both react 18 and 19.
* `Markdown.tsx` has had a large refactor to work with `react-markdown@10` (updated from v5).


site/package.json
```diff
@@ -27,9 +27,8 @@
     "@types/lodash.chunk": "^4.2.9",
     "@types/lodash.memoize": "^4.1.9",
     "@types/lodash.partition": "^4.6.9",
-    "@types/react": "^18.3.18",
+    "@types/react": "^19.1.8",
     "@types/react-dom": "^18.3.0",
```
askoufis: `@types/react-dom` needs to be updated too.

packages/braid-design-system/package.json
```diff
@@ -217,10 +217,9 @@
     "@types/babel__core": "^7.20.1",
     "@types/lodash.curry": "^4.1.9",
     "@types/node": "^18.19.68",
-    "@types/react": "^18.3.18",
+    "@types/react": "^19.1.8",
     "@types/react-dom": "^18.3.0",
```
askoufis: Ditto.

packages/braid-design-system/src/lib/utils/isFragment.ts
michaeltaranto, line 1: Happy with this direction. I would suggest extending the comment to include when we drop support for React 18 we should move back to `react-is`.

site/src/App/Markdown/Markdown.tsx
```diff
@@ -38,69 +34,79 @@ const paragraph = ({ children }: any) => (
...
   },
-  strong: Strong,
-  emphasis: Strong,
-  inlineCode: InlineCode,
-  code: Code,
-  link: TextLink,
-  linkReference: TextLink,
-  blockquote: ({ children }: any) => (
+  strong: ({ children }) => <Strong>{children}</Strong>,
+  em: ({ children }) => <Strong>{children}</Strong>,
```
williamlark, line 92: I can't do `strong: Strong` here since it causes type errors, so I'm forced to explicitly deconstruct and pass through the children

packages/docs-ui/package.json
```diff
@@ -36,12 +36,12 @@
...
-    "@types/react": "^18.3.18",
+    "@types/react": "^19.1.8",
     "braid-design-system": "workspace:*",
-    "react": "^18.3.1"
+    "react": "^19.1.0"
   },
   "peerDependencies": {
     "braid-design-system": "^33.0.0",
-    "react": "^18.2.0"
+    "react": "^19.1.0"
```
michaeltaranto: Any reason to drop 18 from the peer deps here? If not we should keep it.
```suggestion
    "react": "^18 || ^19"
```

packages/generate-component-docs/src/__snapshots__/contract.test.ts.snap
```diff
@@ -1562,187 +1583,12 @@ exports[`BoxRenderer 1`] = `
...
-        | "ul"
-        | "use"
-        | "var"
-        | "video"
-        | "view"
-        | "wbr"
-        | "webview"
+        | number
+        | string
+        | symbol
```
michaeltaranto: This one surprised me. The `component` prop is mapped to `ElementType` which seems to [still be available]. Can we look into this.

[still be available]: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/4e2eae6a902102df3b0635a2f59fce81b405603a/types/react/index.d.ts#L4051

packages/generate-component-docs/src/generate.ts
```diff
@@ -242,7 +244,7 @@ function extractTypeInfo(file: string, options: CompilerOptions) {
         checker.typeToString(unionItem),
       );
 
-      if (isEqual(types.slice(0, 7), unionAliases.ReactNode)) {
+      if (isEqual(types.slice(0, 9), unionAliases.ReactNode)) {
```
michaeltaranto: Might be nice to use the length of the array for this slice to make their relationship explicit.
```suggestion
      if (
        isEqual(
          types.slice(0, unionAliases.ReactNode.length),
          unionAliases.ReactNode,
        )
      ) {
```

site/src/App/Markdown/Markdown.tsx
michaeltaranto: Ah because we mark `children` as `required`. Np.

.changeset/small-lions-fold.md
```diff
@@ -0,0 +1,7 @@
+---
+'braid-design-system': patch
+'@braid-design-system/codemod': patch
```
askoufis: The codemod doesn't have a peer dep on `react`. It still needs a release, but only because of dependency updates.

packages/braid-design-system/src/lib/utils/isFragment.ts
```diff
@@ -0,0 +1,26 @@
+import type { ReactElement } from 'react';
+
+const REACT_ELEMENT_TYPE = Symbol.for('react.transitional.element');
+const LEGACY_REACT_ELEMENT_TYPE = Symbol.for('react.element');
+const REACT_FRAGMENT_TYPE = Symbol.for('react.fragment');
+
+/**
+ * A subset of the logic from `react-is` that supports both React 18 and 19.
+ * We only use this to check for React fragments so only that code has been used.
+ * @see https://github.com/facebook/react/blob/main/packages/react-is/src/ReactIs.js#L40
```
askoufis: Worth linking to the specific commit this code was taken from: https://github.com/facebook/react/blob/97cdd5d3c33eda77be4f96a43f72d6916d3badbb/packages/react-is/src/ReactIs.js#L40. FYI you can press `y` on any github page to update the URL to the exact commit.

site/src/App/Markdown/Markdown.tsx
```diff
@@ -38,69 +34,79 @@ const paragraph = ({ children }: any) => (
...
-        ) : (
-          <Heading level={resolvedLevel === '5' ? '4' : resolvedLevel}>
-            {children}
-          </Heading>
-        )}
+const renderers: Components = {
+  h1: ({ children }) => (
+    <Box>
+      <Heading level="2">{children}</Heading>
+    </Box>
```
askoufis: Can this `Box` be removed?

site/src/App/Markdown/Markdown.tsx
```diff
@@ -38,69 +34,79 @@ const paragraph = ({ children }: any) => (
...
     </TextContext.Provider>
   ),
-  listItem: ({ children }: { children: any }) => {
+  li: ({ children }) => {
     const childList = Children.toArray(children);
 
     // @ts-expect-error
-    if (childList[0]?.type?.isParagraph) {
-      return <Stack space="medium">{children}</Stack>;
+    if (childList[1]?.type?.isParagraph) {
```
askoufis: What's the reason for the change from 0 -> 1 here?

site/src/App/Markdown/Markdown.tsx
```diff
@@ -38,69 +34,79 @@ const paragraph = ({ children }: any) => (
...
+  code: ({ children, className, ...rest }) => {
+    // This property is added by the `checkInlineCode` rehype plugin.
+    if ('data-inline' in rest && rest['data-inline']) {
+      return <InlineCode>{children as string}</InlineCode>;
+    }
+    const match = /language-(\w+)/.exec(className ?? '');
+    return <Code lang={match?.[1] ?? null} value={children as string} />;
+  },
+  a: ({ children, href }) => (
+    <TextLink href={href ?? '#!'}>{children}</TextLink>
```
askoufis, line 103: Is `#!` a special href of some kind? I've never seen it anywhere. Would typically use just `#` for links that intentionally don't go anywhere.
```suggestion
    <TextLink href={href ?? '#!'}>{children}</TextLink>
```

site/src/App/Markdown/Markdown.tsx
williamlark: The output is slightly different and adds an extra element before each paragraph when in a list now. I tried to get it to remove the extra element but it didn't seem doable.

site/src/App/Markdown/Markdown.tsx
williamlark: `#` goes to the top of the page, `#!` acts as an nonexistent anchor and doesn't go anywhere (doesn't scroll to page top)

packages/docs-ui/package.json
williamlark: Nope, missed this. Thanks!

packages/generate-component-docs/src/__snapshots__/contract.test.ts.snap
williamlark: Found the problem. Had to update the `JSX` reference to `React.JSX`. The issue with Braids TS linting meant this didn't get picked up automatically.

.changeset/small-lions-fold.md
williamlark: Yea I did pick this up but forgot I could just create a seperate release file for it. I've done so now.

packages/braid-design-system/src/lib/utils/isFragment.ts
williamlark: Good to know! Updated

site/src/App/Markdown/Markdown.tsx
```diff
@@ -112,18 +114,51 @@ const renderers = {
...
+  };
+
+/**
+ * Removes newlines in lists.
+ * This is used to prevent newlines in list items from being rendered as empty list items.
+ */
+const removeNewlineInLists = () =>
+  function addProp(tree: any) {
+    visit(tree, 'text', (node, index, parent) => {
+      const newLine = '\n';
```
askoufis: `newLine` can be hoisted to a higher scope so it's not re-initialized on each node visit.
```suggestion
    const newLine = '\n';
    visit(tree, 'text', (node, index, parent) => {
```

williamlark: I don't think this is really needed since js will reference the same string in memory because of string interning, it's may just be a little extra garbage collection which won't make any noticeable performance or memory improvements.

Regardless, I've actioned it anyway. My gift to you 🎁

PR #1825, opened 2025-07-03
Warn instead of throwing when an unsupported syntax highlighter language is used

askoufis:
As a followup to #1813, passing `undefined` to `SyntaxHighlighter` is valid, so there's no need to throw an error for unsupported languages. This means if a unsupported language is used it just ends up unhighlighted.


williamlark (dismissed):

packages/braid-design-system/package.json
```diff
@@ -216,7 +216,7 @@
     "@types/autosuggest-highlight": "^3.1.1",
     "@types/babel__core": "^7.20.1",
     "@types/lodash.curry": "^4.1.9",
-    "@types/node": "^18.19.68",
+    "@types/node": "^22.16.0",
```
askoufis, line 219: Had to update the node types so I could use the `glob` and `styleText` APIs. Also it's out of date relative to the configured node version anyway.

site/tsconfig.json
```diff
@@ -1,4 +1,7 @@
 {
   "extends": "../tsconfig.json",
-  "include": ["./src"]
+  "include": ["./src", "./scripts"],
+  "compilerOptions": {
+    "allowImportingTsExtensions": true
+  }
```
askoufis, line 6: `allowImportingTsExtensions` makes sense in the `site` because we use a bundler.

tsconfig.json
```diff
@@ -11,7 +11,7 @@
     "strict": true,
     "jsx": "preserve",
     "lib": ["dom", "ES2022"],
-    "target": "es5",
+    "target": "ES2022",
```
askoufis, line 14: This was necessary to make `tsc` not complain about using a top-level `await` in the new script. It doesn't affect any runtime behaviour since we're not using `tsc` for transpilation.

site/scripts/lintChangesets.ts
```diff
@@ -0,0 +1,66 @@
...
+import { glob, readFile } from 'fs/promises';
+import { resolve } from 'path';
+import { styleText } from 'util';
+
+import { supportedLanguages } from '../src/App/Code/supportedLanguages.ts';
+
+const braidChangesetRegex =
+  /^---\n'braid-design-system': (?:patch|minor|major)\n---$/m;
+
+const markdownCodeBlockLanguageRegex = /``` *(?<language>[a-z]*)\n/g;
```
michaeltaranto: What are your thoughts on tweaking this to only find code blocks with a language?

```suggestion
const markdownCodeBlockLanguageRegex = /```(?<language>[a-z]+)\n/g;
```

site/scripts/lintChangesets.ts
```diff
@@ -0,0 +1,66 @@
...
+    backtickCounter++;
+    if (backtickCounter % 2 === 0) {
+      continue;
+    }
+
+    // Skip code blocks that have no language
+    const language = match.groups?.language;
+    if (!language) {
+      continue;
+    }
```
michaeltaranto: If we update the regex above (and allow no language as we do here), these checks are no longer required, as we just check the matching languages exist in the `supportedLanguages`. Thoughts?

askoufis: Yeah good shout. I was originally capturing all backticks regardless, but since we don't care about blocks with no language this makes sense.

site/scripts/lintChangesets.ts
askoufis: [Adjusted the regexp](https://github.com/seek-oss/braid-design-system/pull/1825/commits/1c1668448dd3c70eda17a4d8ffac87b53f95bd1a#diff-942fec9598cd6ef00e760e5a69fb82b1e42139f63d5f92a3e281eefc4fd2bba8R11). I've kept the ` *` after the backticks because technically that's still valid: you can have any number of spaces before the language of a codeblock.

michaeltaranto (approved):

PR #1826, opened 2025-07-03
docs: Simplify `DocNavigation` component

felixhabib:
Did this as part of debugging but feel like it's worth keeping. Splitting out for ease of PR review, but it should have no functional impact.

1. This component requires a unique layout solution, so I think the code should be a bit more explicit for ease of reading (i.e. comments)
2. Tried to consolidate some `Box` components

### DIFF
--- a/site/src/App/DocNavigation/DocNavigation.tsx
+++ b/site/src/App/DocNavigation/DocNavigation.tsx
@@ -76,7 +76,7 @@ export const DocNavigationItem = ({
   ) : undefined;
 
   return (
-    <Box component="li">
+    <Box component="li" position="relative" paddingX={navItemPaddingX}>
       <Link
         href={href}
         onMouseEnter={() => setHovered(true)}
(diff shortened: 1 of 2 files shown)


site/src/App/DocNavigation/DocNavigation.tsx
```diff
@@ -76,60 +76,57 @@ export const DocNavigationItem = ({
   ) : undefined;
 
   return (
-    <Box component="li">
+    <Box component="li" position="relative" paddingX={navItemPaddingX}>
```
michaeltaranto: Hoisting this padding outside of the link creates dead space between each navigation item which I dont think is desirable.

What do you think?

felixhabib: Yep agreed, will add the improvement after #1810  is merged, as this file will change slightly

site/src/App/DocNavigation/DocNavigation.tsx
```diff
@@ -84,52 +84,56 @@ export const DocNavigationItem = ({
...
-              <Text size="standard" weight="strong">
-                {children}
-              </Text>
-              {badgeElement}
-            </Box>
+        <Box
+          position="relative"
+          display="flex"
+          alignItems="center"
+          paddingX={navItemPaddingX}
```
michaeltaranto: Did you consider applying these styles to the link as a composition with `docNavLink` styles?

site/src/App/DocNavigation/DocNavigation.tsx
```diff
@@ -84,52 +84,56 @@ export const DocNavigationItem = ({
...
+        >
+          {/* Active, strong title */}
+          <Box
```
michaeltaranto: While we are here let's improve the markup. We should really be using `span` elements inside of `a` elements.
```suggestion
          <Box
            component="span"
```

site/src/App/DocNavigation/DocNavigation.tsx
```diff
@@ -84,52 +84,56 @@ export const DocNavigationItem = ({
...
-              display="flex"
-              alignItems="center"
               position="absolute"
-              top={0}
-              paddingY={navItemPaddingY}
-              opacity={active ? 0 : undefined}
-              {...(index === 0
-                ? { left: 0 }
-                : { className: styles.centerHorizontally })}
+              width="full"
```
michaeltaranto: As above.
```suggestion
              component="span"
              width="full"
```

site/src/App/DocNavigation/DocNavigation.tsx
```diff
@@ -84,52 +84,56 @@ export const DocNavigationItem = ({
...
+            <Text size="standard" weight="strong">
+              {children}
+            </Text>
+            {badgeElement}
+          </Box>
+
+          {/* Inactive, weak title */}
+          <Box
+            aria-hidden
+            display="flex"
```
michaeltaranto:
```suggestion
            component="span"
            display="flex"
```

site/src/App/DocNavigation/DocNavigation.tsx
michaeltaranto: Also if the tab content is a flex container we can use `gap` and [remove the `Box` wrapping the badge](https://github.com/seek-oss/braid-design-system/pull/1826/files#diff-72b6a80e83ec2b3c686c1c68b08ebd4760e2f4284c6bf1c642041597c6bc7078R75)

felixhabib: Yep yep. The goal here was to improve readability of the component. I'd think if we want further optimisations, we should do these separately. But happy to go through these now

site/src/App/DocNavigation/DocNavigation.tsx
```diff
@@ -71,8 +68,9 @@ export const DocNavigationItem = ({
   const index = useContext(DocNavigationItemIndexContext);
   const [hovered, setHovered] = useState(false);
 
+  const badgeSpacing = 'xsmall';
   const badgeElement = badge ? (
-    <Box paddingLeft="xsmall">{cloneElement(badge, { bleedY: true })}</Box>
+    <>{cloneElement(badge, { bleedY: true })}</>
```
michaeltaranto:
```suggestion
    cloneElement(badge, { bleedY: true })
```

michaeltaranto (approved):

PR #1830, opened 2025-07-08
repo: Add .nvmrc file

felixhabib:
If you use `mise` currently, lint is failing because it can’t get the correct node version. If we are going with `mise`, this seems necessary


askoufis:
> 👍 (not actually approving as I'm not familiar enough with CI)

CI is still using the volta config in `package.json`. If we want to align on `mise`, we'll need to make sure everyone is using it, and then kill the volta config and update CI to point to `.nvmrc`.

askoufis (approved):

PR #1832, opened 2025-07-08
Table: Ensure `alignY` applied consistently across browsers

michaeltaranto:
Fixes an issue where setting the `alignY` prop to `top` would not apply the `vertical-align` CSS property — instead falling through to our CSS reset which sets `vertical-align: baseline` (rendering inconsistently across browsers).

### DIFF
--- a/.changeset/long-ways-peel.md
+++ b/.changeset/long-ways-peel.md
@@ -1,0 +1,12 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated:
+  - Table
+---
+
+**Table:** Ensure `alignY` applied consistently across browsers
+
+Fixes an issue where setting the `alignY` prop to `top` would not apply the `vertical-align` CSS property — instead falling through to our CSS reset which sets `vertical-align: baseline` (rendering inconsistently across browsers).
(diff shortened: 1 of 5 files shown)


.changeset/long-ways-peel.md
```diff
@@ -0,0 +1,12 @@
+---
+'braid-design-system': patch
+---
+
+---
+updated:
+  - Table
+---
+
+**Table:** Ensure `alignY` applied consistently across browsers
```
felixhabib:
```suggestion
**Table:** Ensure `alignY` prop is applied consistently across browsers
```

packages/braid-design-system/src/lib/components/Table/TableCell.tsx
```diff
@@ -113,7 +113,7 @@ const Cell = forwardRef<HTMLTableCellElement, CellProps & BaseCellProps>(
           [styles.softWidth]: softWidth,
           [styles.minWidth]: typeof minWidth !== 'undefined',
           [styles.maxWidth]: hasMaxWidth,
-          [styles.alignYCenter]: tableContext.alignY === 'center',
+          [styles.alignY[tableContext.alignY]]: Boolean(tableContext.alignY),
```
felixhabib: Do we want to always apply this style? Since we have a default for `alignY`, it's always being set?

felixhabib (approved):

packages/braid-design-system/src/lib/components/Table/TableCell.tsx
michaeltaranto: Yes, it should always be set via context. We default it to `center` in the `Table` component. Are you angling at have a value of `true` here?

michaeltaranto:
```suggestion
          [styles.alignY[tableContext.alignY]]: true,
```

PR #1839, opened 2025-07-15
chromatic: Add diff threshold

felixhabib:
Attempt to prevent diffs with antialiasing as per [docs](https://www.chromatic.com/docs/threshold/).

Default is `.063` but bumping this to `0.1` to see if there's an improvement. Issue is predominantly with `TooltipRenderer` text. Keen to try this out.


felixhabib:
@askoufis personally, I don't think so. If the threshold is high enough to allow false negatives at all it shouldn't be applied to `TooltipRenderer` even. I'm not keen to see more antialiasing diffs and have to add this to more components in the future

PR #1846, opened 2025-07-22
Fix screenshot story const name casing

felixhabib:
In my opinion, these names need a bit of work. But fixing the worst issue first


michaeltaranto (approved):
Step in the right direction. It'll be good to define a convention for this naming and align to it in the future.

PR #1848, opened 2025-07-23
useToast: Fix layout, ensure X padding on mobile

felixhabib:
No changeset because `Toast` update hasn't released yet.


packages/braid-design-system/src/lib/components/useToast/Toaster.css.ts
```diff
@@ -1,8 +1,17 @@
 import { style } from '@vanilla-extract/css';
+import { calc } from '@vanilla-extract/css-utils';
+
+import { gutters } from '../PageBlock/PageBlock';
```
michaeltaranto: We should ideally move the gutters out to a separate file in `PageBlock`, as currently this will import the whole component and its deps (I understand we already use this in `ModalContent` internally).

michaeltaranto (approved):

PR #1864, opened 2025-08-29
ButtonIcon: Better match `Tooltip` text size with `size` prop

felixhabib:


.changeset/true-cars-buy.md
```diff
@@ -0,0 +1,13 @@
+---
+'braid-design-system': minor
```
michaeltaranto: I would consider this a `patch` against the existing `small` feature we have.
```suggestion
'braid-design-system': patch
```

PR #1876, opened 2025-09-25
docs: Update Alert component docs

cherylpaulsen:
Adding design best practice advice. Added new sections: description, choosing a tone, when to use and content guidelines. Also updated the rendering order in DocDetails.tsx so that description appears first on the page and alternatives appears last.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -45,7 +45,7 @@ importers:
         version: 22.18.6
       '@types/react':
         specifier: ^19.1.8
-        version: 19.1.13
+        version: 19.1.15
       '@types/sanitize-html':
         specifier: ^2.13.0
         version: 2.16.0
@@ -54,10 +54,10 @@ importers:
         version: 1.18.8
       '@vanilla-extract/vite-plugin':
         specifier: ^5.0.7
-        version: 5.1.1(@types/node@22.18.6)(babel-plugin-macros@3.1.0)(jiti@2.6.0)(terser@5.44.0)(tsx@4.20.5)(vite@7.1.7(@types/node@22.18.6)(jiti@2.6.0)(terser@5.44.0)(tsx@4.20.5)(yaml@2.8.1))(yaml@2.8.1)
+        version: 5.1.1(@types/node@22.18.6)(babel-plugin-macros@3.1.0)(jiti@2.6.0)(terser@5.44.0)(tsx@4.20.6)(vite@7.1.7(@types/node@22.18.6)(jiti@2.6.0)(terser@5.44.0)(tsx@4.20.6)(yaml@2.8.1))(yaml@2.8.1)
       '@vitejs/plugin-react':
         specifier: ^4.6.0
-        version: 4.7.0(vite@7.1.7(@types/node@22.18.6)(jiti@2.6.0)(terser@5.44.0)(tsx@4.20.5)(yaml@2.8.1))
+        version: 4.7.0(vite@7.1.7(@types/node@22.18.6)(jiti@2.6.0)(terser@5.44.0)(tsx@4.20.6)(yaml@2.8.1))
       concurrently:
         specifier: ^7.6.0
         version: 7.6.0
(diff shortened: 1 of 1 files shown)


packages/braid-design-system/src/lib/components/Alert/Alert.docs.tsx
```diff
@@ -7,13 +7,19 @@ import { dataAttributeDocs } from '../private/dataAttribute.docs';
...
+    </Text>
+  ),
   Example: () =>
     source(
       <Stack space="medium">
         <Alert tone="promote">
           <Text>This is a promoted message.</Text>
         </Alert>
-        <Alert tone="info">
+        <Alert tone="info" data={{ testid: 'alert-1' }}>
```
felixhabib:
```suggestion
        <Alert tone="info">
```
Think we don't need the `testid` here

packages/braid-design-system/src/lib/components/Alert/Alert.docs.tsx
```diff
@@ -169,6 +165,71 @@ const docs: ComponentDocs = {
...
+    },
+    {
+      label: 'When to use',
+      description: (
+        <Stack space="xxlarge">
+          <Stack space="large">
+            <Text>Use an Alert if your message:</Text>
+            <List space="large">
+              <Text>
+                relates to an important condition, status or system change;
```
felixhabib: Maybe we should standardise the punctuation used at the end of each bullet point, like semi-colons or commas everywhere?

packages/braid-design-system/src/lib/components/Alert/Alert.docs.tsx
```diff
@@ -43,29 +49,48 @@ const docs: ComponentDocs = {
...
+      label: 'Formatting your message',
       description: (
         <Stack space="large">
           <Text>
             An Alert can contain layout components such as{' '}
             <TextLink href="/components/Stack">Stack</TextLink> and{' '}
             <TextLink href="/components/Inline">Inline</TextLink>, as well as
             typographic components such as{' '}
-            <TextLink href="/components/Text">Text</TextLink>,{' '}
+            <TextLink href="/components/Text">Text</TextLink>,
```
felixhabib:
```suggestion
            <TextLink href="/components/Text">Text</TextLink>,{' '}
```
Might need to add back the space here?

felixhabib (commented):
First PR, nice one! ⭐

packages/braid-design-system/src/lib/components/Alert/Alert.docs.tsx
cherylpaulsen: I've updated so that all bullet points end with a full stop @felixhabib

felixhabib: Hmm I think if we go for the full stop, maybe we need a capital letter at the start of each line? Why the full stop over the semi-colon out of curiosity?

cherylpaulsen: Truthfully I don't really know what the right answer is. I searched for official rules and found [this guidance](https://www.stylemanual.gov.au/structuring-content/lists).

It suggests:
- For lists that are complete sentences: Start each item with a capital letter and end it with a full stop.
- For lists that are fragments / incomplete sentences: Start each item with lowercase and add a full stop at the last list item only.

I think this list would be considered a fragment with a lead-in, which would suggest lowercase with a full stop on the last item only. Does that feel right? I can also check in with Content Design but it might take a while for them to get back to me.

@felixhabib

felixhabib: Ooo great article! I love the style of start with lowercase with no comma at the end, looks very clean and balanced.

It also says to avoid using 'and' and 'or' in the list items which I think is sensible. Maybe let's do that?

cherylpaulsen: Cool! I think it says to avoid using 'and' and 'or' **after** list items. So within should be okay. Example from the doc below.
Screenshot 2025-09-29 at 4 52 56 pm

felixhabib: Yep yep that's what I meant 😅

cherylpaulsen: Sweet! I also checked in with Content Design and [they confirmed](https://myseek.atlassian.net/wiki/spaces/AWS/pages/2123956542/Punctuation+Writing+conventions#Punctuation-in-bulleted-lists) the guidance we found is correct. I've updated and pushed my latest changes 😎

felixhabib (approved):

PR #1879, opened 2025-09-30
docs: Update Notice component docs

cherylpaulsen:
Adding design best practice advice to include:
- Component description
- Choosing a tone section
- When to use section
- Content guidelines section
- Updated alternatives

Some of the advice for `Notice` is the same as `Alert` so don't stress if you get deja vu while reading 😊

### DIFF
--- a/packages/braid-design-system/src/lib/components/Notice/Notice.docs.tsx
+++ b/packages/braid-design-system/src/lib/components/Notice/Notice.docs.tsx
@@ -70,7 +70,7 @@ const docs: ComponentDocs = {
           <Text>
             Use the <Strong>tone</Strong> property to help communicate the
             meaning behind the Notice. Notices support{' '}
-            <Strong>promote, info, positive, caution</Strong> and{' '}
+            <Strong>promote, info, positive</Strong> and{' '}
             <Strong>critcal</Strong> tones.
           </Text>
           <Text>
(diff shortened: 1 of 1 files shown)


packages/braid-design-system/src/lib/components/Notice/Notice.docs.tsx
```diff
@@ -47,16 +53,35 @@ const docs: ComponentDocs = {
...
+        <Stack space="large">
+          <Text>
+            Use the <Strong>tone</Strong> property to help communicate the
+            meaning behind the Notice. Notices support{' '}
+            <Strong>promote, info, positive</Strong> and{' '}
+            <Strong>critcal</Strong> tones.
+          </Text>
+          <Text>
+            Read more about{' '}
+            <TextLink href="/foundations/tones">Braid tones</TextLink>.
```
felixhabib:
```suggestion
            <TextLink href="/foundations/tones">Tones</TextLink>.
```
Possibly can drop 'Braid' here to keep in line with the other links?

packages/braid-design-system/src/lib/components/Notice/Notice.docs.tsx
felixhabib:
```suggestion
            <TextLink href="/components/Text">Text</TextLink>,{' '}
```
Not your code but I think we need a space here?

packages/braid-design-system/src/lib/components/Notice/Notice.docs.tsx
cherylpaulsen: Good call. I'll update in Alert.docs.tsx as well 👍

felixhabib (approved):

PR #1880, opened 2025-09-30
Migrate to trusted publishing

samchungy:
In order to create a snapshot, just manually run the release workflow on any branch.

We're removing all of our npm tokens and moving to trusted publishing


michaeltaranto (approved):

packages/generate-component-docs/package.json
michaeltaranto, line 1: This package is private, but cant hurt to add it anyway.

PR #1890, opened 2025-10-29
Docs: Accordion updates

cherylpaulsen:
Updates include:
- Added description
- Updated appearance section
- Updated badge section
- Updated icons section
- Updated alternatives section
- Added content guidelines section
- Added when to use section

### DIFF
--- a/packages/braid-design-system/src/lib/components/Accordion/Accordion.docs.tsx
+++ b/packages/braid-design-system/src/lib/components/Accordion/Accordion.docs.tsx
@@ -213,6 +213,44 @@ const docs: ComponentDocs = {
           </>,
         ),
     },
+    {
+      label: 'When to use',
+      description: (
+        <Stack space="xxlarge">
+          <Stack space="large">
+            <Text>Use an Accordion:</Text>
+            <List space="large">
+              <Text>
+                to shorten pages and reduce scrolling when content isn&rsquo;t
+                crucial to read in full (e.g. FYIs)
+              </Text>
+              <Text>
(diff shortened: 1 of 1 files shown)


packages/braid-design-system/src/lib/components/Accordion/Accordion.docs.tsx
```diff
@@ -205,6 +213,44 @@ const docs: ComponentDocs = {
           </>,
         ),
     },
+    {
+      label: 'When to use',
+      description: (
+        <Stack space="xxlarge">
```
felixhabib, line 219: I might've missed this in the other PRs but not sure we need to have the XXL space between these paragraphs? Do we think `xlarge` is ok to be a bit more consistent with the other text?

cherylpaulsen: We’re using `xxlarge` in the other pages. I also just doubled checked DocDetails.tsx and we use `xxlarge` between sections, and `large` within sections. So if we were going to reduce the spacing here, it would probably need to go to `large` to keep the rhythm, but that feels a bit tight to me. Thoughts?

felixhabib: Ah yep yep, I'm leaning towards `xxlarge` between and `xlarge` within sections. Could even make these consts so we don't have to manually write this every time, so might have a go quickly and go back through the other pages. I'm personally leaning towards tighter within sections but open for disagreement

PR #1903, opened 2025-11-12
Docs: Dialog component updates

cherylpaulsen:
Update Dialog docs:
- Add description
- Add 'When to use' section
- Update section headings for 'Width', 'Illustrations' and 'Tailoring the close behaviour'
- Update and rename 'General best practice' section
- Reorder sections

Update Disclosure docs:
- Fix broken link

Changed 'Customise' language to 'Tailor' across these pages:
- Icons
- Pagination
- Drawer
- Dialog

### DIFF
--- a/packages/braid-design-system/src/lib/components/Dialog/Dialog.docs.tsx
+++ b/packages/braid-design-system/src/lib/components/Dialog/Dialog.docs.tsx
@@ -284,8 +284,8 @@ const docs: ComponentDocs = {
           </Text>
           <Text tone="promote" id="translations">
             <IconLanguage title="Translation hint" titleId="translations" /> The{' '}
-            <Strong>aria-label</Strong> for the close button can be customised
-            by providing a <Strong>closeLabel</Strong> prop.
+            <Strong>aria-label</Strong> for the close button can be tailored by
+            providing a <Strong>closeLabel</Strong> prop.
           </Text>
         </>
       ),


packages/braid-design-system/src/lib/components/Dialog/Dialog.docs.tsx
```diff
@@ -292,7 +274,7 @@ const docs: ComponentDocs = {
       },
     },
     {
-      label: 'Customising the close behaviour',
+      label: 'Tailoring the close behaviour',
```
felixhabib, line 277: Curious why we're switching from 'Customising' to 'Tailoring' here

felixhabib (approved):

packages/braid-design-system/src/lib/components/Dialog/Dialog.docs.tsx
cherylpaulsen: Just conscious of the term 'Custom' and how we tend to use it to indicate breaking away from system components

felixhabib: Ahh, yep makes sense. I do see 'customising' as different to 'custom' but not too fussed either way. The issue is that we use this as a common heading across several components such as Pagination and Drawer so we should keep them consistent imo. Probably worth changing that language all at once

PR #1910, opened 2025-11-13
apac: remove theme, Card: remove `rounded` prop

felixhabib:
👋👋🥹 final goodbye to `apac`

Making a pr to branch `v34` to keep master clean. Screenshot tests should be run on pr `v34` -> `master`

This also means we can remove the `legacy` flag on themes

### DIFF
--- a/packages/braid-design-system/src/lib/components/Card/Card.gallery.tsx
+++ b/packages/braid-design-system/src/lib/components/Card/Card.gallery.tsx
@@ -40,30 +40,5 @@ export const galleryItems: GalleryComponent = {
           </Tiles>,
         ),
     },
-    {
-      label: 'Rounded corners',
-      Example: () =>
-        source(
-          <Tiles space="large" columns={{ mobile: 1, tablet: 2 }}>
-            <Stack space="small">
-              <Text size="xsmall" tone="secondary">
-                DEFAULT
-              </Text>
-              <Card>
-                <Placeholder height={100} />
-              </Card>
-            </Stack>
-
-            <Stack space="small">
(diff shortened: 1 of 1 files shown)


packages/braid-design-system/src/lib/components/Card/Card.gallery.tsx
```diff
@@ -58,7 +58,7 @@ export const galleryItems: GalleryComponent = {
               <Text size="xsmall" tone="secondary">
                 ROUNDED
               </Text>
-              <Card rounded>
+              <Card>
```
michaeltaranto: This whole gallery item should be removed

packages/braid-design-system/src/lib/components/MenuItem/useMenuItem.tsx
```diff
@@ -180,8 +179,8 @@ function MenuItemChildren({
   isCheckbox = false,
 }: MenuItemChildrenProps) {
   const menuRendererContext = useContext(MenuRendererContext);
-  const badgeSpace = useBraidTheme().legacy ? 'small' : badgeSlotSpace;
-  const iconSpace = useBraidTheme().legacy ? 'small' : iconSlotSpace;
+  const badgeSpace = badgeSlotSpace;
+  const iconSpace = iconSlotSpace;
```
michaeltaranto: These can be inlined now i'd suggest

felixhabib: Ahh yep!

michaeltaranto (approved):

PR #1911, opened 2025-11-13
Toggle: Remove `bleedY` prop

felixhabib:
More braid `v34` prep following #1910

### DIFF
--- a/.changeset/spicy-clouds-sin.md
+++ b/.changeset/spicy-clouds-sin.md
@@ -1,0 +1,20 @@
+---
+'braid-design-system': major
+---
+
+---
+updated:
+  - Toggle
+---
+
+**Toggle:** Remove `bleedY` prop
+
+Remove the deprecated `bleedY` prop from the `Toggle` component.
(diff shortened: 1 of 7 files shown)


integration/__snapshots__/build-output.test.ts.snap
```diff
@@ -15,10 +12,6 @@ exports[`build > Vanilla Extract files imported as side-effects 1`] = `
 
 ### dist/lib/themes/wireframe/index.mjs:
 1 │ import "./wireframeTheme.css.mjs";
-
-### dist/reset.mjs:
-1 │ import "./lib/css/reset/reset.css.mjs";
-2 │ import "./lib/css/atoms/sprinkles.css.mjs";
```
askoufis, line 21: This doesn't look correct.

felixhabib: No it doesn't, when I run test I am getting it 🫤

michaeltaranto (approved):

PR #1913, opened 2025-11-14
Replace `assert` with `tiny-invariant`

askoufis:
We only use the truthy assertion API from `assert`, which `tiny-invariant` implements in a fraction (0) of [the dependencies of `assert`](https://npmgraph.js.org/?q=assert).

[Sku 12.6.0](https://github.com/seek-oss/sku/releases/tag/sku%4012.6.0) is necessary to strip these assertions, but that's quite an old version so there's not much risk in making it a requirement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,12 +115,12 @@ At SEEK this is done via [sku] as part of the build process.
 
 To ensure correct usage of its components, Braid performs some precondition and invariant checking at runtime using the [tiny-invariant] library.
 
-To prevent these checks from being included in production builds and distrupting the end user experience, it is recommended that `assert` calls are stripped at build time using the [unassert](https://www.npmjs.com/package/unassert) library.
+To prevent these checks from being included in production builds and disrupting the end user experience, it is recommended that `assert` calls are stripped at build time using the [unassert] library.
 
 At SEEK this is done by [sku] as part of the build process via [Babel] with the [babel-plugin-unassert] plugin.
 The minimum `sku` version that supports stripping assertions from the `tiny-invariant` library is [12.6.0].
 
-[assert]: https://www.npmjs.com/package/tiny-invariant
+[tiny-invariant]: https://www.npmjs.com/package/tiny-invariant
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -113,16 +113,18 @@ At SEEK this is done via [sku] as part of the build process.
...
+To ensure correct usage of its components, Braid performs some precondition and invariant checking at runtime using the [tiny-invariant] library.
 
 To prevent these checks from being included in production builds and distrupting the end user experience, it is recommended that `assert` calls are stripped at build time using the [unassert](https://www.npmjs.com/package/unassert) library.
 
-At SEEK this is done via [sku] as part of the build process via [Babel] with the [babel-plugin-unassert] plugin.
+At SEEK this is done by [sku] as part of the build process via [Babel] with the [babel-plugin-unassert] plugin.
+The minimum `sku` version that supports stripping assertions from the `tiny-invariant` library is [12.6.0].
 
-[assert]: https://www.npmjs.com/package/assert
+[assert]: https://www.npmjs.com/package/tiny-invariant
```
michaeltaranto: The link was renamed, so the reference need updating.

```suggestion
[tiny-invariant]: https://www.npmjs.com/package/tiny-invariant
```

README.md
```diff
@@ -113,16 +113,18 @@ At SEEK this is done via [sku] as part of the build process.
 
 ### Assertions
 
-To ensure correct usage of its components, Braid performs some precondition and invariant checking at runtime using the [assert] library.
+To ensure correct usage of its components, Braid performs some precondition and invariant checking at runtime using the [tiny-invariant] library.
 
 To prevent these checks from being included in production builds and distrupting the end user experience, it is recommended that `assert` calls are stripped at build time using the [unassert](https://www.npmjs.com/package/unassert) library.
```
michaeltaranto: Noticed we can use the reference here:

(and typo)

```suggestion
To prevent these checks from being included in production builds and disrupting the end user experience, it is recommended that `assert` calls are stripped at build time using the [unassert] library.
```

PR #1921, opened 2025-11-20
Docs: Update Stepper docs

cherylpaulsen:
Merging best practice guidance. Updates include:
- Add description
- Reorder content
- Update 'Linear mode' section copy
- Update 'Linear mode' example to include secondary copy
- Update 'Non-linear mode' section copy
- Update 'Non-linear mode' example to include secondary copy and navigation buttons
- Add new section 'Providing user controls' with example
- Add 'When to use' section
- Fix spelling 'Forth' to 'Fourth' in multiple locations

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -27,7 +27,7 @@ importers:
         version: 0.5.2
       '@crackle/cli':
         specifier: ^0.16.0
-        version: 0.16.2(@types/node@22.19.1)(babel-plugin-macros@3.1.0)(jiti@2.6.1)(magicast@0.3.5)(terser@5.44.1)(typescript@5.5.4)(webpack@5.102.1(@swc/core@1.15.1)(esbuild@0.25.12))(yaml@2.8.1)
+        version: 0.16.2(@types/node@22.19.1)(babel-plugin-macros@3.1.0)(jiti@2.6.1)(magicast@0.3.5)(terser@5.44.1)(typescript@5.5.4)(webpack@5.102.1(@swc/core@1.15.2)(esbuild@0.25.12))(yaml@2.8.1)
       '@eslint/compat':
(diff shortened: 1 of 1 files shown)


packages/braid-design-system/src/lib/components/Stepper/Stepper.docs.tsx
```diff
@@ -71,20 +83,73 @@ const docs: ComponentDocs = {
   ),
   alternatives: [],
   additional: [
+    {
+      label: 'Alignment',
```
felixhabib, line 87: While we're looking at the docs, I think this section could use a rework too. I honestly can't understand the difference in Alignment modes based on this description or the example

packages/braid-design-system/src/lib/components/Stepper/Stepper.docs.tsx
```diff
@@ -258,52 +367,171 @@ const docs: ComponentDocs = {
...
+            <Strong>solid</Strong>, <Strong>ghost</Strong>,{' '}
+            <Strong>soft</Strong> and <Strong>transparent</Strong>.
           </Text>
+          <Notice>
+            <Text>Resize your browser window to see responsive behaviour.</Text>
+          </Notice>
         </>
       ),
-      Example: () =>
+      Example: ({ setDefaultState, setState, getState }) =>
```
felixhabib, line 629: This example is visually pretty great, but the code is quite long, and uses duplication, which we like to avoid. Maybe let's discuss this one on Slack?

felixhabib (changes requested):
Let's workshop the new example code a little more. Otherwise, looks great!

cherylpaulsen:
Hey @felixhabib, I've updated based on our slack chat:
- Updated the 'Providing user controls' section example
- Removed colons from example captions

PR #1926, opened 2025-12-02
Popover: Ensure `TooltipRenderer` is correctly positioned on Android devices

felixhabib:
Some minor tweaks to docs page as well to ensure ease of maintenance

### DIFF
--- a/packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
+++ b/packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
@@ -115,8 +115,8 @@ export const TooltipContent = ({
           className={styles.arrow[placement ?? 'top']}
           style={assignInlineVars({
             [styles.arrowX]:
-              clampedArrowX !== undefined ? `${clampedArrowX}px` : 'unset',
-            [styles.arrowY]: arrowY !== undefined ? `${arrowY}px` : 'unset',
+              clampedArrowX !== undefined ? `${clampedArrowX}px` : undefined,
+            [styles.arrowY]: arrowY !== undefined ? `${arrowY}px` : undefined,
           })}
         />
       </TooltipTextDefaultsProvider>


packages/braid-design-system/src/lib/components/private/Popover/Popover.tsx
```diff
@@ -88,29 +115,57 @@ const PopoverContent = forwardRef<HTMLElement, PopoverProps>(
...
+      placement: floatingUiRequestedPosition,
+      middleware,
+      whileElementsMounted: autoUpdate,
+    });
 
-    const transitionThresholdInPx =
-      useSpace().space[styles.transitionThreshold];
+    useEffect(() => {
+      refs.setReference(triggerRef.current);
+    }, [triggerRef, refs]);
```
michaeltaranto, line 162: Can Hooks that call `refs.setReference` and `refs.setFloating` use [the `elements` option] instead?

```ts
useFloating({
  ...,
  elements: {
    reference: triggerRef.current,
    floating: ref.current,
  }
})
```

[the `elements` option]: https://floating-ui.com/docs/usefloating#elements

packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
```diff
@@ -56,38 +55,72 @@ export const TooltipTextDefaultsProvider = ({
...
+
+  const [tooltipWidth, setTooltipWidth] = useState(0);
+  const tooltipContainerRef = useRef<HTMLElement | null>(null);
+
+  const handleTooltipContainerRef = (element: HTMLElement | null) => {
+    tooltipContainerRef.current = element;
+    if (element) {
+      const rect = element.getBoundingClientRect();
+      setTooltipWidth(rect.width);
+    }
```
michaeltaranto: This state setting side effect from a ref function doesnt feel like the right approach to take.

Feel like this sort of DOM querying that `useLayoutEffect` was designed for. So rather than refactor all of `clampedArrowX` like we spoke about previously, i'd suggest looking at moving the setting of tooltip width into a layout effect.

packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
```diff
@@ -56,38 +55,72 @@ export const TooltipTextDefaultsProvider = ({
...
+        </Box>
+        <Box
+          ref={arrowRef}
+          position="fixed"
+          background="neutral"
+          className={styles.arrow[placement ?? 'top']}
+          style={assignInlineVars({
+            [styles.arrowX]:
+              clampedArrowX !== undefined ? `${clampedArrowX}px` : 'unset',
+            [styles.arrowY]: arrowY !== undefined ? `${arrowY}px` : 'unset',
```
michaeltaranto: Rather than setting the value to `unset`, is there a reason you are setting the var at all?

```suggestion
              clampedArrowX !== undefined ? `${clampedArrowX}px` : undefined,
            [styles.arrowY]: arrowY !== undefined ? `${arrowY}px` : undefined,
```

packages/braid-design-system/src/lib/components/private/Popover/Popover.tsx
felixhabib: I'm not sure that this works here because the floating element is not in the DOM until useFloating runs, meaning the ref is null. Trying this it's not working. Not sure if there's further work required

PR #1928, opened 2025-12-02
Popover: Fix Autosuggest full width when resizing window, and MenuRenderer offset space

felixhabib:
Following from #1926 noticing these two issues with the migration to `floating-ui/react-dom` after testing.

Each commit can be reviewed individually. Not adding a changeset as this should bring the behaviour back to master.


packages/braid-design-system/src/lib/components/private/Popover/Popover.tsx
```diff
@@ -142,6 +140,14 @@ const PopoverContent = forwardRef<HTMLElement, PopoverProps>(
         shift({
           crossAxis: align === 'center',
         }),
+      width === 'full' &&
```
michaeltaranto: Maybe ternary this with the above condition

michaeltaranto (approved):

PR #1932, opened 2025-12-05
ButtonIcon, TooltipRenderer: Close tooltips after click on desktop

felixhabib:
Been around for a bit, but it's annoying me now. 

For things like `Dialog` you will see the close button tooltip stay after clicking it during the close animation, which is pretty janky. Once the interaction is done, the tooltip no longer serves a purpose and should be hidden imo.

### DIFF
--- a/packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
+++ b/packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
@@ -212,6 +212,7 @@ export const TooltipRenderer = ({
         triggerRef.current.addEventListener('focus', openTooltip);
         triggerRef.current.addEventListener('mouseenter', openTooltip);
         triggerRef.current.addEventListener('mouseleave', closeTooltip);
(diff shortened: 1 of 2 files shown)


packages/braid-design-system/src/lib/components/TooltipRenderer/TooltipRenderer.tsx
```diff
@@ -212,6 +212,7 @@ export const TooltipRenderer = ({
         triggerRef.current.addEventListener('focus', openTooltip);
         triggerRef.current.addEventListener('mouseenter', openTooltip);
         triggerRef.current.addEventListener('mouseleave', closeTooltip);
+        triggerRef.current.addEventListener('click', closeTooltip);
       }
```
michaeltaranto: Adding this here makes the surrounding logic strange.

Suggest a minor tidy up to make `blur` and `click` unconditional, and now `focus`, `mousseleave` and `mouseenter` are if `!isMobileDevice`.

michaeltaranto (approved):

PR #2023, opened 2026-03-25
docs: new section updates

cherylpaulsen:
Primary changes are:
- Applied the new `docSections` to Accordion, Badge and Button docs and removed the old `additional` array
- Renamed `Usage` section to `Best practices`
- Moved `Accessibility` and `Alternatives` to sit under `Best practices`
- Added a hacky style override for `Heading level=2` to land on the look and feel

Cursor-change-summary


site/src/App/HeadingOverride.css.ts
SEEKilian, line 1: Let's delete this file for now and revert back to the existing h2 styles. 

I'll address the styling of these headings as part of a followup with the 'Jump to' feature by creating a new CategoryHeading component given it's used in that menu as well

site/src/App/DocNavigation/DocDetails.tsx
```diff
@@ -54,15 +54,20 @@ const hasContent = (example: {
...
 
+  const hasBestPractices = Boolean(
+    docs?.docSections?.bestPractices?.some(hasContent),
+  );
+
   /*
     Build the ToC sections
-      - Accessibility
-      - Each docSection and their children
+      - Accessibility (top-level when no bestPractices section, otherwise nested within it)
```
SEEKilian: With this change we're reordering where accessibility sits within the docs. Previously we were treating accessibility as high-importance by consistently elevating it to the second item on the page. 

I feel like we should retain that pattern, but open to discussion. Thoughts @michaeltaranto ?

packages/braid-design-system/src/lib/components/Button/Button.docs.tsx
```diff
@@ -169,239 +169,266 @@ const docs: ComponentDocs = {
...
+              </Button>
+            </Actions>,
+          ),
+      },
+      {
+        description: (
+          <Stack space="large">
+            <Heading level="4">Critical</Heading>
+            <Text>
+              For destructive actions like “Delete” you can set the
```
SEEKilian: These quotations should be replace with [HTML entities](https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references#List_of_character_entity_references_in_HTML)

Also present on the following line is an `&rsquo;` where an `&apos;` should be used.

packages/braid-design-system/src/lib/components/Button/Button.docs.tsx
```diff
@@ -416,755 +443,737 @@ const docs: ComponentDocs = {
...
+                horizontally to facilitate better alignment with surrounding
+                text.
+              </Text>
+            </Notice>
+            <Text>
+              For example, we can align a button to a{' '}
+              <TextLink href="/components/Heading">Heading</TextLink> element
+              using an <TextLink href="/components/Inline">Inline</TextLink>,
+              even though the button is actually taller than the heading. If we
+              didn’t use the <Strong>bleed</Strong> prop in this case, the
```
SEEKilian: This `’` should be replace with an `&apos;`

site/src/App/DocNavigation/DocDetails.tsx
cherylpaulsen: Happy to revert if we think that's best. Just worth noting, if we do 'Accessibility' will be the only section not assigned to a docSection. Maybe it's okay tho given it would be at the top of the page before any other H2

cherylpaulsen: I've reverted the change so 'Accessibility' renders at the top again 😎 @SEEKilian

SEEKilian (approved):

PR #2029, opened 2026-04-07
Add 'Jump to' functionality to Braid Docs

SEEKilian:
Adds a Modal that allows for quickly jumping to certain Navigation items. Opens with CTRL/CMD+K, as well as an icon below the theme switcher.

**`JumpToModal` component** — A new site-level modal (Leverages Braid's `Dialog`) that lets users search across Foundations, Components, CSS, and Logic by name. Supports arrow-key navigation through results, `Enter` to navigate, `Shift+Enter` to jump directly to a component's props page (if present), and auto-focuses the search input on open.

**`SearchResults` & `getSearchItems`** — Supporting modules for the modal: `getSearchItems` builds a flat list of all searchable items from the existing navigation helpers/routes, and `SearchResults` renders them grouped by category.

**`KeyboardShortcut` & `KeyboardIcon` components** — New `docs-ui` components that render styled key cap badges (e.g. `⌘`, `K`) alongside a label, used to surface shortcut hints in the UI.

**`HeaderNavigation` updated** — Added an `onSearchClick` prop and a new search trigger button in the header, showing the platform-appropriate shortcut hint (`⌘K` / `Ctrl+K`) using `KeyboardShortcut`.

**`Navigation` updated** — Wires up `isSearchOpen` state, renders `JumpToModal`, and passes `onSearchClick` down to both the fixed and sticky `Header` instances.

**`useSearchHotkey` hook** — Listens for `⌘K` (macOS) or `Ctrl+K` (other) globally and calls `onOpen` to trigger the modal, with platform detection via a small utility (`isMac`).

**`CategoryHeading` component** — For rendering uppercased category labels within the search results list - will be leveraged within `docDetails` in a follow-up branch


packages/docs-ui/src/components/HeaderNavigation/HeaderNavigation.tsx
```diff
@@ -41,6 +55,31 @@ export const HeaderNavigation = ({
...
+              navigator.platform === 'iPod'
+                ? '⌘'
+                : 'Ctrl',
+              'K',
+            ]}
+            shortcutLabel={<IconSearch />}
+          />
+        </Box>
+      </Bleed>
+    </Stack>
```
michaeltaranto: Based one where this landed we can remove the wrapping `Stack` and the `Box` around the `themeToggle`.

packages/docs-ui/src/components/HeaderNavigation/HeaderNavigation.tsx
```diff
@@ -41,6 +55,31 @@ export const HeaderNavigation = ({
...
+      <Bleed horizontal="xxsmall" bottom="xxsmall">
+        <Box
+          component="button"
+          padding="xxsmall"
+          paddingRight="xsmall"
+          borderRadius="standard"
+          className={searchButton}
+          onClick={onSearchClick}
+        >
+          <KeyboardShortcut
```
michaeltaranto, line 68: A subtle one, but can you ensure all elements rendered in here use `span` elements. Given we wrap it in a `button` ideally we dont use `div` elements inside.

packages/docs-ui/src/components/HeaderNavigation/HeaderNavigation.css.ts
```diff
@@ -0,0 +1,8 @@
+import { style } from '@vanilla-extract/css';
+import { vars } from 'braid-design-system/css';
+
+export const searchButton = style({
+  ':hover': {
+    background: vars.backgroundColor.neutralSoft,
```
michaeltaranto: When we change the background colour outside of React Context we need to manage the colour mode change manually. So here need to use the `colorModeStyle` utility and specify the right token for dark mode.

You can test it on the docs site by focusing the window and typing `braiddark` 🥸

site/tsconfig.json
```diff
@@ -1,6 +1,11 @@
 {
   "extends": "../tsconfig.json",
-  "include": ["./playroom.config.mts", "./src", "./scripts"],
+  "include": [
+    "./playroom.config.mts",
+    "./src",
+    "./scripts",
+    "../packages/docs-ui/src/components/KeyboardShortcut"
```
michaeltaranto: This should not be necessary. Let me know if this is a deliberate work around for something we need to look at.

packages/docs-ui/src/index.ts
michaeltaranto, line 1: Need to add a changeset for this new component

packages/docs-ui/src/components/HeaderNavigation/HeaderNavigation.tsx
michaeltaranto, line 1: Should add a changeset for this, as its new API for the docs-ui component

site/src/App/CategoryHeading/CategoryHeading.tsx
```diff
@@ -0,0 +1,13 @@
...
+export const CategoryHeading = ({
+  children,
+}: {
+  children: React.ReactNode;
+}) => (
+  <Box style={{ textTransform: 'uppercase' }} component="li">
+    <Text size="xsmall" weight="medium" component="h2">
+      {children}
+    </Text>
+  </Box>
```
michaeltaranto: I dont think these elements are correct for all usages. I would suggest flipping the Text and Box around and exposing the `component` prop on the outer Text component, so the element can be set from the call site.
```suggestion
  <Text component={component} size="xsmall" weight="medium">
    <Box component="span" style={{ textTransform: 'uppercase' }}>
      {children}
    </Box>
  </Text>
```

site/src/App/CategoryHeading/CategoryHeading.tsx
```diff
@@ -0,0 +1,13 @@
+import { Box, Text } from 'braid-src/index';
```
michaeltaranto: This import path is not right, we might have some setup issues to work through.

site/src/App/JumpToModal/getSearchItems.ts
michaeltaranto, line 1: I believe `getSearchItems` can be a static array rather behind a function call. That will mean it also wont need to be memoised as its value never changes.

site/src/App/JumpToModal/JumpToModal.tsx
```diff
@@ -0,0 +1,177 @@
...
+      open={isOpen}
+      onClose={onClose}
+      width="medium"
+      title="Jump to a page"
+      closeLabel="Close search"
+    >
+      <TextField
+        icon={<IconSearch />}
+        ref={inputRef}
+        label=""
```
michaeltaranto: We should use `aria-label` here instead of a blank `label`

site/src/App/JumpToModal/SearchResults.tsx
```diff
@@ -0,0 +1,118 @@
...
+  groupedResults,
+  flatResults,
+  selectedIndex,
+  onSelectIndex,
+  onNavigate,
+}: SearchResultsProps) => {
+  if (flatResults.length === 0 && searchQuery.trim()) {
+    return (
+      <Box
+        background="neutralSoft"
```
michaeltaranto: Favour using `neutralLight` over `neutralSoft`, where soft primarily serves the button use case. This will get you dark mode support out the box too.

Maybe a subtle radius on this one too?

site/src/App/JumpToModal/SearchResults.tsx
```diff
@@ -0,0 +1,118 @@
...
+        justifyContent="center"
+        alignItems="center"
+        height="full"
+      >
+        <Text tone="secondary" size="xsmall" align="center">
+          Matching pages will appear here.
+        </Text>
+      </Box>
+    );
+  }
```
michaeltaranto: Maybe combine with the previous if block or use a component for this placeholder state to keep them consistent.

site/src/App/JumpToModal/SearchResults.tsx
```diff
@@ -0,0 +1,118 @@
...
+          const items = groupedResults[category];
+          if (items.length === 0) {
+            return null;
+          }
+
+          return (
+            <Stack key={category} space="xxsmall" component="ul">
+              <Box marginBottom="xsmall">
+                <CategoryHeading>{category}</CategoryHeading>
+              </Box>
```
michaeltaranto: The use of margin means the stack model is probably not right. This is also highlighted by the separation of the `ul` and `li` elements. I suggest something like this:

```tsx
<Stack space="xsmall">
  <CategoryHeading>...</CategoryHeading>

  <Stack space="xxsmall" component"ul">
    {items.map(() => (
      <Bleed component="li">
        <Spread space="small">
          <ButtonLink>...</ButtonLink>
          <KeyboardShortcut />
        </Spread>
      </Bleed>
    ))}
  </Stack>
</Stack>
```

packages/docs-ui/src/components/HeaderNavigation/HeaderNavigation.tsx
SEEKilian: Yep good pickup

site/src/App/CategoryHeading/CategoryHeading.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export const CategoryHeading = ({
+  children,
+  component,
+}: {
+  children: React.ReactNode;
+  component?: ComponentProps<typeof Text>['component'];
+}) => (
+  <Box style={{ textTransform: 'uppercase' }}>
+    <Text size="xsmall" weight="medium" component={component}>
```
michaeltaranto, line 12: A consumer setting `component` would likely be more interested in controlling the container element. Suggest applying to the `Box`:
```suggestion
  component?: ComponentProps<typeof Box>['component'];
}) => (
  <Box component={component} style={{ textTransform: 'uppercase' }}>
    <Text size="xsmall" weight="medium">
```

site/src/App/DocNavigation/DocNavigation.tsx
```diff
@@ -213,7 +226,23 @@ export const DocNavigation = () => {
...
     <Stack space={['xlarge', 'xxlarge']}>
       <Stack space={['large', 'xlarge']}>
-        <Heading level="1">{docsName}</Heading>
+        <Inline space="small" alignY="center">
+          {iconBrowseSearch !== undefined ? (
+            <ButtonIcon
+              variant="transparent"
+              size="large"
+              label="Back to browse"
+              icon={<IconChevron direction="left" />}
```
michaeltaranto, line 235: Recommend using `IconArrow` for this in favour of Chevron

site/src/App/JumpToModal/SearchResults.tsx
```diff
@@ -0,0 +1,105 @@
...
+          {searchQuery.trim()
+            ? 'No results found.'
+            : 'Matching pages will appear here.'}
+        </Text>
+      </Box>
+    );
+  }
+
+  return (
+    <Stack space="large" test-id="search-results">
```
michaeltaranto, line 53:
```suggestion
    <Stack space="large">
```

michaeltaranto (approved):
A couple of minor improvements, but im looking forward to this one landing. Nice work.

PR #2034, opened 2026-04-23
Docs category headings

SEEKilian:
Migrated `DocDetails` and `SideNavigation` to use new `<CategoryHeading>` component.

This kicked off a minor cascade of changes in support, primary of which is the creation of a new `<TitleLink>` component - a hash-linked anchor with an optional copyable prop that enables copying the URL to clipboard.

Other changes are minor refactors, elevation of `<CategoryHeading>` into `docs-ui`, etc.


.changeset/legal-pigs-tan.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+---
+
+**CategoryHeading:** Added new component for rendering category-style navigation headings.
+
+**EXAMPLE USAGE:**
+```jsx
+<CategoryHeading component="h2">Layout</CategoryHeading>
+```
```
michaeltaranto: Extract this to separate changeset please.

packages/docs-ui/src/components/CategoryHeading/CategoryHeading.tsx
```diff
@@ -0,0 +1,18 @@
...
+type CategoryHeadingProps = {
+  children: ReactNode;
+  component?: ComponentProps<typeof Text>['component'];
+};
+
+export const CategoryHeading = ({
+  children,
+  component,
+}: CategoryHeadingProps) => (
+  <Box component={component} display="flex" gap="xsmall" alignItems="center">
```
michaeltaranto: Is this outer Box still required? If not, we can move the `component` back to the `Text` node and remove the Box wrapper

packages/docs-ui/src/components/TitleLink/TitleLink.css.ts
```diff
@@ -0,0 +1,24 @@
...
+  {
+    maxWidth: 'fit-content',
+    scrollMarginBlockStart: vars.space.small,
+    outlineOffset: vars.space.xsmall,
+  },
+]);
+
+export const hashLink = style({
+  selectors: {
+    [`${titleLink}:hover &, ${titleLink}:focus-visible &`]: {
```
michaeltaranto: What are your thoughts on showing the tick while copying? So if the user moves away we still show the tick for the two seconds.

packages/docs-ui/src/components/TitleLink/TitleLink.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const { copying, onCopyClick } = useCopy();
+
+  const handleClick: MouseEventHandler<HTMLAnchorElement> = (event) => {
+    if (event.metaKey) {
+      return;
+    }
+    event.preventDefault();
+    const anchor = event.currentTarget as HTMLAnchorElement;
+    const url = `${window.location.origin}${window.location.pathname}${anchor.hash}`;
+    onCopyClick(url);
```
michaeltaranto: Pretty sure we can just grab the `href` from the element:
```suggestion
    const anchor = event.currentTarget as HTMLAnchorElement;
    onCopyClick(anchor.href);
```

site/src/App/DocNavigation/DocSection.tsx
```diff
@@ -14,33 +14,46 @@ export const DocSection = ({
...
-          background={section.background}
-          showCodeByDefault={
-            section.showCodeByDefault || section.Example === undefined
+  <Stack space="small">
+    {section.deprecated ? (
+      <Bleed left="xsmall">
+        <Badge
+          tone="caution"
+          id={section.label ? `deprecated-${section.label}` : undefined}
+          aria-hidden="true"
```
michaeltaranto: I dont believe the Badge support `aria-hidden`.  So would either need to add an inner span or add the feature on Badge which im not against.

site/src/App/DocNavigation/DocSection.tsx
```diff
@@ -14,33 +14,46 @@ export const DocSection = ({
...
+        >
+          Deprecated
+        </Badge>
+      </Bleed>
+    ) : undefined}
+    <Stack space={headingSpacing}>
+      {section.label ? (
+        <Heading
+          level="3"
+          aria-describedby={
```
michaeltaranto: As above, i dont believe this is passed through (TS doesnt validate aria- so no errors, but also it doesnt apply).

We could do with a lap of whitelisting and opening some aria props i think.

packages/docs-ui/src/components/TitleLink/TitleLink.css.ts
SEEKilian: Yeah I like that

site/src/App/DocNavigation/DocSection.tsx
SEEKilian: This is a little hairy with a programatic aria-describedby as part of TitleLink. I'll push up a solution and you can give me your thoughts

.changeset/red-drinks-float.md
```diff
@@ -0,0 +1,16 @@
...
+  - vars
+---
+
+**vars:** Exposed `vars.transition`. Transition CSS variables are available in stylesheets and runtime styles.
+
+**EXAMPLE USAGE:**
+```ts
+export const myStyle = style({
+  transition: vars.transition.fast,
+});
```
michaeltaranto: Missed the closing code block, and maybe just include the Braid import to emphasise our vars here:

````suggestion
```ts
import { vars } from 'braid-design-system/css':

export const myStyle = style({
  transition: vars.transition.fast,
});
```
````

File: site/src/App/DocNavigation/DocSection.tsx [hunk 8]
```diff
@@ -13,34 +19,42 @@ export const DocSection = ({
...
+          <Badge tone="caution">Deprecated</Badge>
+        </Bleed>
+      ) : undefined}
+      <Stack space={headingSpacing}>
+        {section.label ? (
+          <Heading level="3">
+            <TitleLink label={section.label} copyable>
+              {section.label}
+              {deprecated ? (
+                <HiddenVisually>, deprecated</HiddenVisually>
```
michaeltaranto, line 37: Just updating semantics inside the link:

```suggestion
                <HiddenVisually component="span">, deprecated</HiddenVisually>
```

packages/braid-design-system/src/lib/css/vars.docs.tsx
```diff
@@ -49,16 +49,22 @@ const Row = ({
...
 );
 
-const ContentWidthValue = ({ var: varName }: { var: string }) => {
-  const [size, setSize] = useState(0);
+const CssVarValue = ({
+  var: varName,
+  property,
+}: {
+  var: string;
+  property: 'width' | 'transition';
```
michaeltaranto: Should we type this as `CSSStyleDeclaration` given we are using it to access computed style props and provide an inline style object. Just trying to avoid the `any`

michaeltaranto (approved):

PR #2037, opened 2026-05-01
site: Migrate sku configs to TypeScript

michaeltaranto:
Improve the type safety of our sku config and route configurations for the site.


SEEKilian (approved):

site/package.json
```diff
@@ -21,10 +21,10 @@
   "license": "MIT",
   "dependencies": {
     "@babel/core": "^7.26.0",
-    "@babel/preset-typescript": "^7.26.0",
```
askoufis, line 24: This should still be a dep right? The script still uses it.

michaeltaranto: Correct, depcheck was not happy as it didnt find it. Looking at the build atm.

PR #2047, opened 2026-05-15
Remove `braid-src` alias usage from `braid-design-system` package

askoufis:
## Why

I'm starting to look into migrating braid's docs site to use Vite via `sku`. Now that we're using `tsdown` for bundling we can use native functionality like dev conditions to make consuming source code at dev time a lot simpler than it currently is. One step towards that is undoing some of the old patterns in the repo such as the `braid-src` import alias.

These days we have package.json [`exports`](https://nodejs.org/api/packages.html#exports) and [`imports`](https://nodejs.org/api/packages.html#imports) handling public entrypoints and private aliases natively. Modern tooling supports this config very well, so where possible we should use it over aliases. `braid-src` probably won't go away entirely, but we definitely don't need it in the braid package itself anymore.

## History

I _think_ some of the reasons for it being added were:
- Enforcing one consistent way to import things from site code, docs, etc.
- Reducing relative path imports (e.g. less `../../../../../foo/bar` - particularly relevant for the docs site)
- To _prevent_ vscode from suggesting potentially incorrect import paths. I think these paths were to directories that were necessary during the `crackle` times, so we shouldn't need anymore.

## Solution

Most `braid-src` usage can be replaced with public `braid-design-system` entrypoints, and the rest can use relative paths. I've added an eslint rule to enforce this as well.


packages/braid-design-system/src/lib/css/vars.docs.tsx
```diff
@@ -15,9 +9,17 @@ import {
...
+import { vars } from 'braid-design-system/css';
+import { Fragment, type ReactNode, useEffect, useRef, useState } from 'react';
+import Code from 'site/App/Code/Code';
+import { ThemedExample, useThemeSettings } from 'site/App/ThemeSetting';
+import type { CssDoc } from 'site/types';
+
+// TODO: COLORMODE RELEASE
+// Use public import
+import { Box } from '../components/Box/Box';
+import type { ReactNodeNoStrings } from '../components/private/ReactNodeNoStrings';
```
michaeltaranto: We can remove this usage in favour of looking at how its used and match the required type. In this case, `varDocs` gets [mapped to the `description`] of `ComponentExample`. So [`varDocs`] could be more correctly typed as:
```diff
-const varDocs: Record<keyof typeof vars, ReactNodeNoStrings> = {
+const varDocs: Record<keyof typeof vars, ComponentExample['description']> = {
```

[mapped to the `description`]: https://github.com/seek-oss/braid-design-system/blob/41f6f6d5e0d235de639b26e605feb23b01d8037e/packages/braid-design-system/src/lib/css/vars.docs.tsx#L362-L365
[`varDocs`]: https://github.com/seek-oss/braid-design-system/pull/2047/changes#diff-2a187431485cbe1266cdbafac2db1ff2fe89ffccabb2ea50d04e7518bec90330R86

michaeltaranto (commented):
One minor comment, but love to see this underway. Whats interesting is while the files are co-located "in the braid package", they "belong to the site". Not sure if we can rejig configs to capture that relationship, otherwise we keep fighting tooling like depcheck.

michaeltaranto (approved):

PR #2052, opened 2026-05-19
Dialog: Add `coverImage` support

michaeltaranto:
Add support for providing a URL for a cover image to display at the top of the Dialog via the `coverImage` prop.
The provided image must be compatible with the [CSS background-image “url” function].

**EXAMPLE USAGE:**
```jsx
<Dialog
  title="Cover Image"
  coverImage="https://placehold.co/1600x900/051A49/FFFFFF/png?text=​++++++16:9++++++​"
  open={true}
  onClose={() => {}}
>
  <Placeholder height={100} width="100%" label="Dialog Content" />
</Dialog>
```

[CSS background-image “url” function]: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/background-image


packages/braid-design-system/src/lib/components/private/Modal/ModalContent.tsx
```diff
@@ -110,10 +239,35 @@ export const ModalContent = ({
...
+
   const justifyContent = (
     { left: 'flexStart', center: 'center', right: 'flexEnd' } as const
   )[position];
-  const modalRadius = position === 'center' ? 'xlarge' : undefined;
+  const modalRadius = !isDrawer ? 'xlarge' : undefined;
+
+  const modalLayout = (
+    <ModalContentScrollLayout
+      scrollColumnNotDialog={allowColumnLayout}
```
SEEKilian: I don't know that this is the most parseable prop name. Maybe something like restrictContentHeight / enforceContentScrolling (or limitColumnHeight to match the className?)

SEEKilian (approved):

packages/braid-design-system/src/lib/components/private/Modal/ModalContent.tsx
michaeltaranto: I agree with you! I'll update to `applyCoverImageHeightLimit`

SEEKilian (approved):

PR #2053, opened 2026-05-20
Replace `depcheck` with `knip`

askoufis:
`depcheck` hasn't had an update for a few years. `knip` is actively maintained and much more powerful than `depcheck`, and also has OOTB support for many common config files/package patterns (e.g. it picks up vitest config, package entrypoints, etc.). At the moment it's configured to only analyze dependencies, but more lints can be configured in the future such as identifying dead code and unused exports.

A few unused deps were identified, as well as some deps that were installed at the monorepo root but have now been moved to workspace package(s) where they're actually used.


packages/braid-design-system/package.json
```diff
@@ -168,13 +168,11 @@
     "@testing-library/user-event": "^14.4.3",
     "@types/autosuggest-highlight": "^3.1.1",
     "@types/babel__core": "^7.20.1",
-    "@types/gradient-parser": "^0.1.5",
     "@types/lodash.curry": "^4.1.9",
     "@types/node": "^22.16.0",
     "@types/react": "^19.1.8",
     "@types/react-dom": "^19.1.6",
     "@types/sanitize-html": "^2.13.0",
-    "@types/testing-library__jest-dom": "^5.9.1",
```
askoufis, line 177: `gradient-parser` and `@testing-libray/jest-dom` bring their own types.

knip.json
michaeltaranto, line 1: Should we remove this in favour of the jsonc? or vice versa?

askoufis: Woops, didn't realise I committed both!

michaeltaranto (approved):

PR #2054, opened 2026-05-20
Replace more `braid-src` imports with public entrypoints

askoufis:
I think this is the rest of the easy migrations done. Everything else is actual private imports which will be harder to address, if we even want to do that.

Replacing the namespace theme import with 4 separate imports is a bit tedious. Maybe it would be worth exposing a grouped `braid-design-system/themes` entrypoint for this purpose, but for now it's a necessary evil.

I also modified some of the vitest configs to resolve the `braid-dev` export condition so entrypoints can point to source code. This is effectively the native way of pointing to source code that doesn't involve aliases, and is necessary now in order to run tests without having to build between every change.


packages/braid-design-system/vitest.config.ts
```diff
@@ -9,5 +9,8 @@ export default defineProject({
     setupFiles: './vitest-setup.ts',
     css: true,
   },
+  resolve: {
+    conditions: ['braid-dev'],
+  },
```
askoufis, line 14: Braid tests use jsdom, which relies on vite's _client_ pipeline, as opposed to the storybook tests which _don't_ use jsdom and instead rely on vite's SSR pipeline.

packages/codemod/src/fixtures/testFixture.ts
askoufis, line 1: I couldn't find any references to this file, but I'm not 100% certain it can be deleted.

site/src/App/Code/Code.tsx
```diff
@@ -11,10 +11,10 @@ import {
 } from 'braid-design-system';
 // TODO: COLORMODE RELEASE
 // Use public import
+import usePlayroomScope from 'braid-design-system/playroom/scope';
```
michaeltaranto:
```suggestion
import usePlayroomScope from 'braid-design-system/playroom/scope';
// TODO: COLORMODE RELEASE
// Use public import
```

storybook/chromatic.ts
```diff
@@ -1,5 +1,15 @@
...
+import seekBusiness from 'braid-design-system/themes/seekBusiness';
+import seekJobs from 'braid-design-system/themes/seekJobs';
+import wireframe from 'braid-design-system/themes/wireframe';
+
+const themes = {
+  docs,
+  seekBusiness,
+  seekJobs,
+  wireframe,
+};
```
michaeltaranto: What are your thoughts on having an `allThemes.ts` file locally in `storybook` and `site`. Then we can more easily find and update the list should we need to

packages/codemod/src/fixtures/testFixture.ts
michaeltaranto: Not sure actually. If there are no references we can delete it.

storybook/chromatic.ts
askoufis: Yeah I think that'll do for now.

askoufis: https://github.com/seek-oss/braid-design-system/pull/2054/commits/f6b281ea5a44ea6195d54625cd09d792fc8ffe94

michaeltaranto (approved):

PR #2069, opened 2026-06-02
Inline: Add `noWrap` to support single-row compositions

SEEKilian:
Added noWrap prop to `Inline` to allow for single row composition

### DIFF
--- a/packages/braid-design-system/src/lib/components/Inline/Inline.docs.tsx
+++ b/packages/braid-design-system/src/lib/components/Inline/Inline.docs.tsx
@@ -2,6 +2,7 @@ import source from '@braid-design-system/source.macro';
 import type { ComponentDocs } from 'site/types';
 
 import {
+  Box,
   Divider,
   Inline,
   Notice,
@@ -231,7 +232,7 @@ const docs: ComponentDocs = {
       ),
       Example: () => {
         const { value: visual } = source(
-          <ScrollContainer direction="horizontal">
+          <Box overflow="auto">
             <Inline space="small" noWrap>
               <Placeholder width={20} height={48} />
               <Placeholder width={80} height={48} />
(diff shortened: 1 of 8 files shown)


packages/braid-design-system/src/lib/components/Inline/Inline.docs.tsx
```diff
@@ -212,6 +213,65 @@ const docs: ComponentDocs = {
...
+              <Placeholder width={120} height={48} />
+              <Placeholder width={60} height={48} />
+              <Placeholder width={40} height={48} />
+              <Placeholder width={180} height={48} />
+              <Placeholder width={100} height={48} />
+              <Placeholder width={60} height={48} />
+              <Placeholder width={120} height={48} />
+              <Placeholder width={40} height={48} />
+            </Inline>
+          </ScrollContainer>,
```
michaeltaranto: Feel like we should demo this without the Scroll Container as it is may be inferred as that affordance is provided. Think a clearer demo would be a Box wrapper as described above (then it would work in Playroom too)

packages/braid-design-system/src/lib/components/Inline/Inline.docs.tsx
```diff
@@ -212,6 +214,67 @@ const docs: ComponentDocs = {
...
+            </Inline>
+          </Box>,
+        );
+
+        return {
+          code: codeDemo,
+          value: visual,
+        };
+      },
+    },
```
michaeltaranto: Now that they are aligned we can just maintain the one template:

```suggestion
      Example: () => source(
        <Box overflow="auto">
          <Inline space="small" noWrap>
            <Placeholder width={20} height={48} />
            <Placeholder width={80} height={48} />
            <Placeholder width={40} height={48} />
            <Placeholder width={150} height={48} />
            <Placeholder width={120} height={48} />
            <Placeholder width={60} height={48} />
            <Placeholder width={40} height={48} />
            <Placeholder width={180} height={48} />
            <Placeholder width={100} height={48} />
            <Placeholder width={60} height={48} />
            <Placeholder width={120} height={48} />
            <Placeholder width={40} height={48} />
          </Inline>
        </Box>,
      ),
    },
```

packages/braid-design-system/src/lib/components/Inline/Inline.docs.tsx
```diff
@@ -12,6 +13,7 @@ import {
   Tiles,
 } from '../';
 import { Placeholder } from '../../playroom/components';
+import { ScrollContainer } from '../private/ScrollContainer/ScrollContainer';
```
michaeltaranto:
```suggestion
```

michaeltaranto (approved):

PR #2075, opened 2026-06-16
Docs uplift style

cherylpaulsen:
Playing with some style tweaks for the new nav and IA:
- Update nav item active state to `soft` variant
- Remove sidebar background colour and add right border to match header nav
- Remove gap below header nav so it lines up with sidenav
- Expand sidebar nav width to line up better with Braid logo and wordmark
- Reduce logo and wordmark to size small
- Update TOC style to match border styles of Playroom preview frames
- Hide repetitive sidebar section titles


knip.jsonc
```diff
@@ -11,7 +11,6 @@
     },
     "packages/braid-design-system": {
       "entry": ["src/**/*.{docs,gallery,screenshots,snippets}.tsx"],
-      "ignoreUnresolved": ["./Icon.*Svg"],
```
askoufis, line 14: TL;DR: Locally this gets flagged as unnecessary config but it's necessary in CI.

I should've left a comment here when I added this. Locally knip might complain that this config is unnecessary, but it seems to be necessary on CI because `lint:knip` runs _before_ Icon*Svg components are generated. See [these logs](https://github.com/seek-oss/braid-design-system/actions/runs/27597489121/job/81590731015?pr=2075#step:6:14).

cherylpaulsen: Okie I've added it back! @askoufis

michaeltaranto:
Closing while we continue iterate on the IA and structure.

PR #2077, opened 2026-06-25
`Accordion`: add new `subText` prop

zhua633:
see [thread](https://seekchat.slack.com/archives/CDL5VKRHN/p1782175258679989)
**AccordionItem:** Add `subText` prop to display a caption beneath the label.

**EXAMPLE USAGE:**

```jsx
<Accordion>
  <AccordionItem label="Label" subText="Supporting caption for this item">
    <Placeholder height={80} />
  </AccordionItem>
</Accordion>
```

running local playground: 
| | collapsed | expanded|
| --- | --- | --- |
| with `subText` field | Screenshot 2026-06-25 at 12 54 03 pm | Screenshot 2026-06-25 at 12 54 07 pm |
| without `subText` field | Screenshot 2026-06-25 at 12 54 17 pm | Screenshot 2026-06-25 at 12 54 21 pm |

### DIFF
--- a/.changeset/tangy-hornets-design.md
+++ b/.changeset/tangy-hornets-design.md
@@ -1,0 +1,15 @@
+---
+'braid-design-system': minor
+---
+
+**AccordionItem:** Add `subText` prop to display a caption beneath the label.
+
+**EXAMPLE USAGE:**
+
+```jsx
+<Accordion>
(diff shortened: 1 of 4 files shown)


zhua633:
Braid probably won't support this use case, but there may be a description prop that helps describe the content of the Accordion and is tied into the semantic/aria layer for accessibility later down the line

PR #2079, opened 2026-06-30
ButtonIcon: Add support for `solid` variant

michaeltaranto:
**ButtonIcon:** Add support for `solid` variant

**EXAMPLE USAGE:**
```jsx
<ButtonIcon
  variant="solid"
  label="Solid variant"
  icon={<IconSend />}
/>
```

### DIFF
--- a/packages/braid-design-system/src/lib/components/ButtonIcon/ButtonIcon.tsx
+++ b/packages/braid-design-system/src/lib/components/ButtonIcon/ButtonIcon.tsx
@@ -98,6 +98,7 @@ const ButtonIconContent = forwardRef<HTMLButtonElement, ButtonIconProps>(
       size: size === 'small' ? 'small' : 'standard',
       radius: 'full',
     });
+    const fallbackIconTone = variant === 'solid' ? 'neutral' : tone;
 
     assert(
       icon && icon.props.size === undefined,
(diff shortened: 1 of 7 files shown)


packages/braid-design-system/src/lib/components/ButtonIcon/ButtonIcon.tsx
```diff
@@ -143,7 +143,7 @@ const ButtonIconContent = forwardRef<HTMLButtonElement, ButtonIconProps>(
             }
           >
             {cloneElement(icon, {
-              tone: icon.props.tone || tone,
+              tone: icon.props.tone || variant === 'solid' ? 'neutral' : tone,
```
jahredhope: This might be a silly question, but wouldn't this evaluate to:

```
(icon.props.tone || variant === 'solid') ? 'neutral' : tone
```
As `||` is high [precedence](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence) than ternaries?

So tone is always 'neutral' if it's truthy?

jahredhope (commented):
I like the docs changing but I think the operator precedence might cause an issue.

packages/braid-design-system/src/lib/components/ButtonIcon/ButtonIcon.tsx
michaeltaranto: Correct! I'll tidy that up

jahredhope (approved):
Code looks good, will need to resolve merge conflicts from the docs change PR

PR #2086, opened 2026-07-07
deps: Bundle font metrics

askoufis:
Only the `wireframe` and `docs` themes use font metrics from `@capsizecss/metrics`, yet consumers (most of which only use `seekJobs`) have to install the entire dependency (~27mb unpacked). By bundling only the used font metrics into `dist`, `@capsizecss/metrics` can be a dev dep instead.

The tradeoff is that consumers can no longer get new font metrics by updating dependencies in their app. For the fonts being bundled (`courierNew` and `appleSystem`) I think this tradeoff is acceptable as their metrics are unlikely to change very often.

The practical change to the built package is the following files being added to `dist` (only 4 files are added, but `du` shows all sub-directories):

image

I had to update `tsdown` in order to configure [`deps.onlyBundle`](https://tsdown.dev/options/dependencies#deps-onlybundle) as this was only available in v0.21.0. Note the `inlinedDependencies` field added to `package.json` - this is a convention used by `tsdown` to denote dependencies that have been inlined into the built package.

I figured I'd float the idea in a PR since it was quick to implement, but feel free to close this if there's a good reason not to bundle these font metrics.


.changeset/honest-waves-look.md
```diff
@@ -0,0 +1,5 @@
+---
+'braid-design-system': patch
+---
+
+Remove `@capsizecss/metrics` dependency and bundle font metrics instead
```
askoufis: This wording can probably be improved, wasn't sure on the best way to phrase it.

SEEKilian (approved):

.changeset/honest-waves-look.md
michaeltaranto:
```suggestion
Reduce consumer bundle size by including only the required fonts from `@capsizecss/metrics` 
```

michaeltaranto (approved):

PR #2108, opened 2026-08-03
Fix Dialog/Drawer Jest crash from react-focus-lock CJS interop

pete-seeker:
## Summary
- Dialog/Drawer unit tests (Jest/CJS) were crashing with `Element type is invalid… Check the render method of Modal` after the `tsdown` 0.22 bump
- Modal’s CJS build can resolve `react-focus-lock`’s default export to the module namespace instead of the FocusLock component
- Unwrap the default export in Modal so FocusLock always resolves to the real component, without pinning `tsdown`

## Test plan
- [x] Rebuild `braid-design-system` and confirm Modal CJS emits the unwrap
- [x] Against Chalice linked to this local build: previously failing Dialog/Drawer suites pass (7 suites / 93 tests)
- [ ] Confirm Braid’s own Dialog/Drawer modal test suite still passes in CI

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/calm-cups-focus.md
+++ b/.changeset/calm-cups-focus.md
@@ -1,0 +1,5 @@
+---
+'braid-design-system': patch
+---
+
+Fix Dialog/Drawer crashing under Jest (CJS) when resolving `react-focus-lock` as Modal's focus trap
--- a/.changeset/calm-cups-focus.md
+++ b/.changeset/calm-cups-focus.md
@@ -2,4 +2,4 @@
 'braid-design-system': patch
 ---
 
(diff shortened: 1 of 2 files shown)


askoufis (approved):

.changeset/calm-cups-focus.md
```diff
@@ -0,0 +1,5 @@
+---
+'braid-design-system': patch
+---
+
+Fix Dialog/Drawer crashing under Jest (CJS) when resolving `react-focus-lock` as Modal's focus trap
```
askoufis:
```suggestion
Fix Dialog/Drawer crashing under Jest (CJS) when resolving `react-focus-lock`
```

PR #2112, opened 2026-08-05
🐛 Fix docs search shortcut hydration mismatch

pete-seeker:
## 📝 What changed

- Default the docs header search shortcut to `⌘` + `K` for SSR/first paint (Mac-first)
- After mount, swap to `Ctrl` + `K` only when not on an Apple platform (`useLayoutEffect`)
- Detect Apple platforms with `navigator.userAgent` (avoids deprecated `navigator.platform`)

## 💡 Why

- Production docs (`seek-oss.github.io/braid-design-system`) were reading platform during render, so Linux SSR HTML (`Ctrl`) disagreed with Mac client hydration (`⌘`) and triggered React minified error #418
- Most docs users are on Mac, so defaulting to `⌘` keeps that path stable with no flash

## ⚠️ Impact

- Docs UI only
- Non-Apple platforms may briefly show `⌘` before swapping to `Ctrl`
- No change to Braid component APIs

### DIFF
--- a/packages/docs-ui/src/components/HeaderNavigation/HeaderNavigation.tsx
+++ b/packages/docs-ui/src/components/HeaderNavigation/HeaderNavigation.tsx
@@ -24,8 +24,7 @@ interface HeaderNavigationProps {
   themeToggle?: ReactNode;
 }
 
-const isApplePlatform = () =>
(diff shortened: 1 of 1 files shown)


jahredhope (commented):
I like the code. Slight formatting issue breaking CI.

leoacosta (approved):

PR #2119, opened 2026-08-12
Make remaining doc TitleLinks copyable

SEEKilian:
Enable `copyable` on Details headings that were still jump-only (Accessibility, Alternatives), so TOC-linked sections share the same copy-link behaviour as category/section headings

Enable `copyable` on Props TitleLinks (component names, Further References) for consistency with Details/Releases/Templates

No change to `TitleLink`’s default, and no change to guides/examples (`LinkableHeading` stays jump-to for now)


leoacosta (approved):
🚀

PR #2126, opened 2026-08-20
Extract icons package

pete-seeker:
## 📝 What changed

- Add `@braid-design-system/icons` as the canonical SVG package (public, scoped).
- Move icon SVGs out of `braid-design-system` into `packages/icons/svg/`, with platform suffixes and `Fill` drawings where needed.
- Point web `generate:icons` at that package only; Braid still ships generated React icons.
- Major changeset so the first publish is `1.0.0`.
- Consumer README: install/import, file naming, catalog list without implementation detail.

## 💡 Why

- iOS and Android can consume one SVG catalog instead of shipping their own.
- Web generation stays a Braid build step; app consumers of `braid-design-system` are unchanged.

## ⚠️ Impact

- No Braid runtime API change. SVGs are a generate-time devDependency.
- Publishing happens after this PR merges, via the Version Packages PR.

## 🧪 Test plan

- [x] `pnpm generate` (no unexpected SVG component churn)
- [x] `pnpm lint`
- [x] `pnpm test` (809 passed, 2 skipped)
- [ ] After merge: confirm Version Packages PR sets `@braid-design-system/icons` to `1.0.0`

- [x] 🧪 Code changes tested
- [x] 📚 Documentation or metadata updated


packages/braid-design-system/scripts/generateIcons.cts
```diff
@@ -51,83 +48,75 @@ const svgrConfig = {
...
+    const { stem } = parseIconFileName(svgFilePath);
+    const existing = svgPathsByStem.get(stem);
+
+    if (!existing || isWebOverride(svgFilePath) || !isWebOverride(existing)) {
+      svgPathsByStem.set(stem, svgFilePath);
+    }
+  }
+
+  return [...svgPathsByStem.values()];
+};
```
michaeltaranto: This is a great capability we are adding here! Nice work.

I think we can tighten this up though, there are a lot of nested or repeated function calls that could probably be simplified based on where the code has landed.

Happy to pair if you like.

michaeltaranto (approved):

PR #2130, opened 2026-08-27
Drawer: Ensure full height content extends to available height

michaeltaranto:
Fixes a regression in a consumers ability to insert full height content into a Drawer.

Also refining the position of the close button on mobile to be more balanced.


leoacosta (approved):

michaeltaranto:
Closing while we work through an edge case that we want to handle.

michaeltaranto:
Edge case is not related to the Drawer itself so good to proceed with this fix.

leoacosta (approved):
🚀

PR #2137, opened 2026-09-03
Drawer, Dialog: Restore default stack spacing to content

michaeltaranto:
Fixes a regression where the default gap spacing was no longer being applied to child content in `Dialog` and `Drawer`.


leoacosta (approved):
🚢

pete-seeker (approved):

PR #2144, opened 2026-09-14
Accordion: Animate height and add autoCollapse

pete-seeker:
## 📝 What changed

- Accordion items animate height when opening and closing (`prefers-reduced-motion` disables it).
- New `autoCollapse` prop: opening an item closes any other open item. Starts collapsed; item-level `expanded` cannot be used with it.

## ⚠️ Impact

- Collapsed panels stay in the document at zero height (`visibility: hidden`, `aria-hidden`, `inert`) instead of `display: none`.
- Height animation applies to all Accordion items, not only `autoCollapse`.


michaeltaranto (commented):
I'd like to iterate on this API a bit more on this one. Also might be some expanded state management improvements we can make. Lets catch up and work through it

